PR #8, opened 2019-01-04
Add a shortcode to template for single-source content

RichieEscarez:
Add shortcode to help reduce unnecessary content duplication and enable the Docsy template to handle single-source content.

Copied from https://github.com/gohugoio/hugo/tree/master/docs/layouts/shortcodes

Usage: 
To copy all content from 'shared' file_1, into another file_2, add the following line to file_ 2:

HTML
{{% readfile file="/path/from/site-root/file_1.md" %}}
{{% readfile file="/path/from/site-root/file_1.html" %}}

Markdown
{{% readfile file="/path/from/site-root/file_1.md" markdown="true" %}}

### DIFF
--- a/layouts/shortcodes/readfile.md
+++ b/layouts/shortcodes/readfile.md
@@ -1,0 +1,8 @@
+{{$file := .Get "file"}}
+{{- if eq (.Get "markdown") "true" -}}
+{{- $file  | readFile | markdownify -}}
+{{- else if  (.Get "highlight") -}}
+{{-  highlight ($file  | readFile) (.Get "highlight") "" -}}
+{{- else -}}
+{{ $file  | readFile | safeHTML }}
+{{- end -}}


RichieEscarez:
This is something that I've found works with HTML (only tested with pre-built HTML). Since I've not come across other alternatives for using single-source content (searching the web), I thought this would be a good addition to the Docsy template.

RichieEscarez:
Tested this more today and markdown works fine too. Although since Docsy by default uses "Use the new Chroma Go highlighter", the 'highlight' conditional doesnt work unless you update your config.toml to use "pygmentsUseClasses = true".

markmandel:
Curious - how does this differentiate from writing a shortcode to store single-source content?

RichieEscarez:
> Curious - how does this differentiate from writing a shortcode to store single-source content?

Interesting, i didn't consider that model but i think that would work too. I'm totally new to Hugo so just learning all this site stuff as i go. 

One advantage to using the readfile shortcode is that your single-source file can reside in any directory (ie. single source content can live in the related "content type" folder) and then consume from any other directory (versus all files living under the /shortcode dir).

PR #14, opened 2019-01-14
Adjust timing of colour change in top nav bar on front page

mk-nickyang:
Fixes https://github.com/kubeflow/website/issues/393

This issue was found in Kubeflow's website (https://github.com/kubeflow/website), which uses the Docsy theme


sarahmaddox:
Hallo @mk-nickyang  It looks as if the CLA check is still failing. Please would you try signing again - one thing to check is whether you used the same email address when signing the CLA and when committing the changes.

mk-nickyang:
I signed it!

PR #15, opened 2019-01-14
Fix color contrast for inlined code

mk-nickyang:
Fixes https://github.com/kubeflow/website/issues/402

Inline code color before fix:
[Screenshot from Gyazo](https://gyazo.com/c2a7f8da50e91c171ee10a22d10def10)

Inline code color after fix:
[Screenshot from Gyazo](https://gyazo.com/4638c5fc0c439b599f368e277e501007)

### DIFF
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -21,6 +21,7 @@
 
     // Inline code
     p code {
+        color: inherit;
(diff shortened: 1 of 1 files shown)


mk-nickyang:
I signed it!

PR #33, opened 2019-02-15
Remove href if no previous/next item

markmandel:
"" could well be a broken link, so when using a linkchecker, this comes up as a false positive.

(I recently switched to https://github.com/wjdp/htmltest which is brutally stringent, but finds WAY more things than [linkchecker](https://wummel.github.io/linkchecker/) ever did)

Another option could be to point this to `#` if this is not an acceptable solution.


LisaFC:
I think this approach works for me, @chenopis? Anyone else?

PR #64, opened 2019-04-01
Remove default navbar icon

### LINKED ISSUE #54: Title off center on mobile
The title is off center because of assets/icons/logo.svg being being included in the html. If the file is deleted then an error is generated during rendering. I couldn't figure out a way to conditionally include the logo from the assets/ directory. Page resources can't read it and readFile doesn't seem to work either.

Screenshot 2019-03-14 at 19 35 07

LisaFC:
Fixes #54 

Does this seem like a sensible approach? It basically removes the default "blank" navbar icon and updates the navbar layout to check if there's an icon before rendering. Will need to update the docs as well slightly (user won't need to do anything different but they'll only be replacing an icon if they've copied the example site, if we put one there...).


layouts/partials/navbar.html
```diff
@@ -1,7 +1,7 @@
 {{ $cover := .HasShortcode "blocks/cover" }}
 <nav class="js-navbar-scroll navbar navbar-expand navbar-dark {{ if $cover}} td-navbar-cover {{ end }}flex-column flex-md-row td-navbar">
 	<a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
-		<span class="navbar-logo">{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }} {{ end }}</span><span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>
+		<span class="navbar-logo">{{ if (fileExists "icons/logo.svg") }}{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}{{ end }}</span><span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>
```
ianlewis: This doesn't work for me and returns false even if the file exists. I had tried this before and I couldn't get fileExists to work with files in assets/

LisaFC: Huh, let me play around with it some more, it worked for me with one example....

LisaFC: Aha, figured it out. You have to actually specify the assets/ directory as well for fileExists - you don’t pass fileExists the path to the file as it appears with hugo server but use the path on the disk.

LisaFC:
OK, updating this slightly to leave the icon in the template (as we've now added a nice new icon...) but with the option to remove it.

LisaFC:
Though, hmm, now it's really fiddly to remove it if you don't want it as you can't (as far as I can tell) override a file with "no file". Let me have another think. Might put the logos in the example site rather than the theme.

LisaFC:
OR (sorry for thinking aloud in a PR) could parameterize it so you can just choose to have no logo. Would that be simpler?

RichieEscarez:
This PR is missing the actual config.toml with `navbar_logo` element. Thinking `true` by default? For example:
```
# User interface configuration
[params.ui]
.....

# Enable main nav bar logo (/assets/icons/logo.svg)
navbar_logo = true
```

assets/icons/logo.svg
```diff
@@ -1,3 +1,237 @@
-<?xml version="1.0"?>
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 240 240">
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 23.0.3, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 width="500px" height="500px" viewBox="0 0 500 500" style="enable-background:new 0 0 500 500;" xml:space="preserve">
```
RichieEscarez: The width = 500 shows as:

image

Do we want to cut the horizontal whitespace down?

RichieEscarez: ah, i see you are already fixing this with: https://github.com/google/docsy/pull/79

LisaFC:
It is indeed true by default but yes, it's probably clearer if we add it in.

RichieEscarez:
It actually doesn't work for me if the attribute and value is not in the config.toml.

LisaFC:
Huh, you may be right. I have so many versions of these projects on the go that I forget which ones have which files. I'll add it in.

LisaFC:
Already done (clearly the abovementioned "so many versions of these projects on the go" problem meant I'd forgotten I'd already added it to the example site config file).

https://github.com/LisaFC/docsysite/blob/master/config.toml

LisaFC:
(basically anything that needs config changes, the config changes won't be in the theme, they'll be in the example site)

RichieEscarez:
oh ok, but since I cloned google/docsy, i didnt get that fix (and appeared missing to me). LGTM!

LisaFC:
Yeah, we don't have the config settings in the theme, just the example site (which is why we tell people to copy that config.toml even if they're not using our site structure). It is the Hugo Way.

PR #67, opened 2019-04-03
Add subpage list to landing pages for docs

### LINKED ISSUE #40: Add new index-landing layout
Create a layout for generating "index-type" landing pages in docs sections, eg with a list of all the topics in a particular section.

LisaFC:
Gives you the option of:
- no list (current behaviour)
- bulleted list of links to subpages
- nice formatted list with page descriptions from frontmatter (default)

Default behaviour shown here: https://serene-roentgen-10fbaa.netlify.com/docs/adding-content/

This default behaviour may give unexpected results on existing sites, though I don't see anything potentially troubling on the sites I know about.

Fixes #40


LisaFC:
@sarahmaddox @RichieEscarez PTAL

LisaFC:
> Thanks @LisaFC this looks great. I have two requests:
> 
> * Some code comments may be nice, to help people understand which bits of code result in which bits of page.

Good idea! Done.

> * Could you include an example that shows what happens when a section includes one or more subsections as well as single pages?

It actually looks exactly the same: you just get the title of the section landing page and the description from that page - so in my current example if Navigation and Search was a section rather than a page, you'd get the section/index title (Navigation and Search) and any description on the index page. If you click Navigation and Search in that case you'd get brought to the Navigation and Search index page which would by default have a list of *its* contents.  Does that help or is there anything else I can demo? 

I suppose it depends on whether you want some indication on the landing page that something is a section, but then we head into duplication of menus territory especially if people are on laptop/desktop where the same full menu is on the left.

LisaFC:
Have also realized that for larger intros it might be nice to have a line or something to divide the detailed list from the text. Hang on...

sarahmaddox:
Thanks, love the code comments. :)

Regarding the request to show what happens when a section includes one or more subsections as well as single pages: When I did my quick and dirty solution, I ended up with weird ordering - all the sections appeared first, then the single pages, with no indication of why that was so. Thus I'm wondering if you've managed to solve that issue.

LisaFC:
No, mine worked fine - it's sorting $pages by weight, so it either uses the weight of the page or the weight of the section index page (which is the same approach Docsy uses to sort the entries in the left nav).

LisaFC:
Just going to add some CSS that I realized I added to the demo to make it look nicer.

layouts/docs/list.html
```diff
@@ -1,3 +1,45 @@
 {{ define "main" }}
-{{ .Render "content" }}
-{{ end }}
\ No newline at end of file
+<div class="td-content">
+	<h1>{{ .Title }}</h1>
+	{{ with .Params.description }}<div class="lead">{{ . | markdownify }}</div>{{ end }}
+	{{ .Content }}
+<div class="section-index">
```
RichieEscarez: What if this div (or the contents of the div) was a partials so that it could be easily overridden/personalized?

LisaFC: As in the section index div? Sure, though we've already given them three potential options for customizing it and it's already most of the page.

I'll have a go and see how it looks.

LisaFC: I think that's actually nicer! Thanks for suggestion.

PR #74, opened 2019-05-27
Added code styling within tables

sarahmaddox:
Fixes https://github.com/kubeflow/website/issues/730

Code formatting embedded in text within a table is hard to read - the colour is too light. This change makes the inline code style within a table the same as the style outside a table.

**Before:**
without-style-fix

**After:**
with-style-fix

### DIFF
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -20,7 +20,7 @@
     }
 
     // Inline code
-    p code, li > code {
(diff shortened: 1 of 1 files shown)


rionam:
This is soooo much better! Thank you!

LisaFC:
Ooh, lovely. Thanks for doing this, and LGTM.

PR #78, opened 2019-05-31
Update _code.scss

### LINKED ISSUE #75: Prevent <code> from wrapping
In a laptop-sized browser window, I encountered some unpleasantly wrapped text on the Kubeflow website (see the left column of [this screenshot](https://user-images.githubusercontent.com/1772874/58515489-e07ef580-8159-11e9-84e1-21e8b3b39d7d.png)).

Docsy already [sets `white-space: pre` for `pre > code`](https://github.com/google/docsy/blob/master/assets/scss/_code.scss#L49). Could we apply the same style for [inline code](https://github.com/google/docsy/blob/master/assets/scss/_code.scss#L22) as well?

I think it's very unlikely that you would ever want a `<code>...</code>` element to wrap: in general they are short, and I think it's more pleasant to view the type of identifiers or code snippets in these blocks uninterrupted/pushed to the next line, rather than wrapped.

The one issue you could run into is that the code element overflows off the right edge of its parent on a very small screen, but I think this is almost impossible (would require a very long string to overrun even a small mobile screen). (You could also add a media query to tell `<code>` to only wrap on small screens, but this seems like overkill.)

LisaFC:
Fixes #75

### DIFF
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -25,6 +25,7 @@
         padding: 0.2em 0.4em;
         margin: 0;
         font-size: 85%;
+	word-break: normal;
         background-color: rgba($black, 0.05);
         border-radius: $border-radius;
 
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -20,7 +20,7 @@
     }
 
     // Inline code
-    p code, li > code, td > code, th > code {
+    p code, li > code, table code {
         color: inherit;
         padding: 0.2em 0.4em;
         margin: 0;
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -25,7 +25,7 @@
         padding: 0.2em 0.4em;
         margin: 0;
         font-size: 85%;
-	word-break: normal;
+        word-break: normal;
         background-color: rgba($black, 0.05);
         border-radius: $border-radius;
 


RichieEscarez:
cc/ alecglassford

assets/scss/_code.scss
```diff
@@ -25,6 +25,7 @@
         padding: 0.2em 0.4em;
         margin: 0;
         font-size: 85%;
+	word-break: normal;
```
RichieEscarez: for this to resolve the actual issue in #75, the `word-break: normal;` must also applied to `tr` (and not specific to a child of `p`) for example:

Include `tr code` (without the `>` child selector):
```
// Inline code
    p code, li > code, td > code, th > code, tr code {
...
```

Or alternatively, just apply it to all elements of a table  (use `table code`):
```
// Inline code
    p code, li > code, table code {
...

```

RichieEscarez:
```suggestion
        word-break: normal;
```
fix indentation

RichieEscarez:
@alecglassford @LisaFC Any objection to applying this inline code behavior to ALL tables (whether or not they are nested elements)?

Where all tables that have inline code: `table code` use/apply: `word-break: normal;`

```
// Inline code
    p code, li > code, table code {
    ....
    word-break: normal;
    ....
```

where we remove/replace the `p td > code, th > code,` selector (since they are all nested in a `table` and inherit that behavior)

LisaFC:
Sounds fine to me.

RichieEscarez:
PTAL

PR #79, opened 2019-05-31
Update logo.svg

### LINKED ISSUE #77: Docsy logo isn't scaling properly
As demonstrated here: https://serene-roentgen-10fbaa.netlify.com/

(I hadn't updated the theme on that site in a while so I only noticed it today when I was fixing some bits)

For some reason the relevant element is the right height but 500 px wide, and it's throwing the entire top navbar off. Dunno do we need to fix it in CSS or in the actual SVG file. Will experiment.

LisaFC:
Remove height and width as they're causing wonkiness when the svg is styled with our CSS.

Fixes #77

### DIFF
--- a/assets/icons/logo.svg
+++ b/assets/icons/logo.svg
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Generator: Adobe Illustrator 23.0.3, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 width="500px" height="500px" viewBox="0 0 500 500" style="enable-background:new 0 0 500 500;" xml:space="preserve">
+	 viewBox="0 0 500 500" style="enable-background:new 0 0 500 500;" xml:space="preserve">
 <g>
 	<path style="fill:#FFFFFF;" d="M116.8525,421.9722c-5.7041,0-10.3442-4.3127-10.3442-9.6129V88.183
 		c0-5.3002,4.6401-9.6117,10.3442-9.6117H320.858c3.0347,0,9.3959,0.5498,11.7506,2.6302l0.3545,0.3442l58.905,63.2912


assets/icons/logo.svg
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Generator: Adobe Illustrator 23.0.3, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 width="500px" height="500px" viewBox="0 0 500 500" style="enable-background:new 0 0 500 500;" xml:space="preserve">
+	 viewBox="0 0 500 500" style="enable-background:new 0 0 500 500;" xml:space="preserve">
```
RichieEscarez: I could not just delete these properties but was able to scale everything down

```suggestion
	width="72px" height="85.65px" viewBox="0 0 72 85.65" style="enable-background:new 0 0 72 85.65;" xml:space="preserve">
```

LisaFC: I think the main thing is that vector graphics are insane. :) If you make that change do you get the logo looking normal in the top nav?

AFAIK because it's an inline <svg> and not in an  then the height and width are overridden by the CSS, which is:

       svg {
            display: inline-block;
            margin: 0 10px;
            height: 30px;
        }

...and if I take the height and width out the logo just scales to fit the CSS height.

This page is v helpful: https://css-tricks.com/scale-svg/

RichieEscarez: yes it renders correctly. i basically scaled down the svg keeping the scaling:
image

But i agree, my troubles might be my lack in experience with svg's and/or tooling. So without opening the svg graphic in another text editor and manually deleting the width and height attributes, I was unable to use 'gimp' to directly save my svg (excluding width and height). 
-> therefore we might need to document the need to manually remove those attributes?

Other options:

Perhaps, instead of trying to utilize an SVG here (since the benefit it adds is scaling large or small), we instead use a set size image like a JPG  (which has the benefit of small storage and fast loads) for the main menu icon since remains small and doesnt ever scale?

Another thought is that we reuse and demonstrate the use of a single SVG for both the tiny icon and again, elsewhere in a large format (a product banner?).

LisaFC: It won't save properly in Gimp anyway as Gimp isn't a vector graphics editor, though it'll render SVGs if you open them - you need something like Inkscape to actually edit them. TBH I think we may be digging in too deep here about image formats and their benefits - if someone wants to create a logo they will almost certainly either a)know how to work with SVGs or b)farm it out to someone who does (we got an external vendor to create our logo), and scalability is important for logos etc. looking good at different resolutions/screen sizes. All our existing Docsy sites (and the other Hugo ones I know of) have SVG logos as it's pretty standard.

(I did read something about how while in general you don't want SVGs to have width/height properties if you want them to be completely responsive there are some cases where you do for small images, I'll follow up. Kubeflow and Istio both don't have width/height in their SVGs)

We should possibly mention that you may want to override the CSS for SVGs (as Kubeflow did) if you want to tweak your SVG display.

RichieEscarez: Sorry, i mixed my editors up. It was Inkscape that i used. 

All sounds good to me. I think we are making the same point, that being that this seems too complicate for just an ity bity graphic. Ultimately, i think we need to mention that manual intervention of your SVG is needed, whether thats defining a specific size (overriding the CSS) or how to manually remove the width/height.

LisaFC: Yeah, I wouldn't mention much about it in this context, but anyone who's actually designing a logo will/should know what they're doing and how SVGs work with CSS. Manual intervention may not be needed depending on the tools you use, some of them save with a height/width by default and some don't (or it's configurable). I can stick a note in the relevant section saying that here is the CSS used by the navbar and that the logo is an inline SVG, if you want your logo to look good you may need to either size the image, ensure it doesn't have width/height attributes, or override our CSS.

LisaFC: Et voila: https://serene-roentgen-10fbaa.netlify.com/docs/adding-content/iconsimages/

RichieEscarez: Perfect! I was just concerned that our default SVG shows some values of 500 x 500, which a new user might confuse with height and width (=> if they use a free editor like i did, they would get frustrated by those default settings and the inability so save the file with the right settings). At least now its mentioned. Thanks!

PR #82, opened 2019-06-03
Accessibility: Add aria-label to footer icons that have a tooltip

MattyQ:
Due to the way bootstrap.js removes the content of the title attribute on elements that have tooltips, and because there is no other content that is readable by assistive technology, the icons in the footer are not currently accessible.

Adding an aria-label that ingests the name of the icon ensures that screen readers and other assistive technology will have a static reference to read, and also saves having to define a new config property just for the aria-label attribute.

We could add a new property to config.toml if that was preferred, which would let the content of the aria-label attribute be defined separately.

This could also be considered a temporary fix until the bootstrap tooltips are updated to be more accessible. If we want to treat it as temporary, probably better to stick with aria-label storing the .name, so no rework is needed in the future if someone wanted to update their template and we remove the aria-label.

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -29,7 +29,7 @@
 {{ define "footer-links-block" }}
 <ul class="list-inline mb-0">
   {{ range . }}
-  <li class="list-inline-item mx-2 h3" data-toggle="tooltip" data-placement="top" title="{{ .name }}">
+  <li class="list-inline-item mx-2 h3" data-toggle="tooltip" data-placement="top" title="{{ .name }}" aria-label="{{ .name }}">
     <a class="text-white" target="_blank" href="{{ .url }}">
       <i class="{{ .icon }}"></i>
     </a>


RichieEscarez:
If this gets merged into Docsy and then worst case we forget that is there, what would the side effect be when/if bootstrap merges in your fix?

MattyQ:
Good question! The bootstrap fix as currently proposed adds an aria-label attribute to an element only if that element doesn't already have one. If we implement aria-labels this way, they will not be overwritten by bootstrap, nor will it break how the tooltips are being loaded.

RichieEscarez:
@LisaFC, do you have a preference here? 

If we merge this, the only additional item that I think we might need, is to mention this behavior in the docs somewhere (ie. "by default, icons in the footer automatically apply the name of your icon as the `aria-label` accessibility attribute" + add a link to maybe the related bootstrap topic?).

MattyQ:
I'm also happy to tweak this PR to add a value to config.toml if we'd prefer to handle it that way. I can add a doc update to this PR, also, if that's preferred.

LisaFC:
I think the way it handles it is fine, I think making it a config setting is possibly overkill. Where would we put it in the docs? We don't have a specific accessibility section....

RichieEscarez:
Maybe somewhere in the customization topics? 

Since this issue occurs when a tooltip is added to some element, it could be categorized under "Adding Tooltips" or something along those lines:

"Due to the way bootstrap.js removes the content of the title attribute on elements that have tooltips, you must add an aria-label that ingests the name of the element to ensure that screen readers and other assistive technology have a static reference to read."

MattyQ:
FWIW, it's specifically elements that don't have any other readable content, which is why this PR only addresses the footer icons. For example, if you added a tooltip to a line of text, the way bootstrap handles it appears to work as expected (you can experience the expected behavior when using ChromeVox, for example, in Bootstrap's tooltip doc: https://getbootstrap.com/docs/4.3/components/tooltips/).

I opted to pick up the name as it more closely matches the label for the icon, rather than the title, which tooltips use for the content (it relies on aria-describedby). Basically this just restores the expected Bootstrap functionality by giving AT a static label to read for the footer icons.

LisaFC:
That sounds like it doesn't really need documenting then, I'd only call it out if it actually did something unexpected.

RichieEscarez:
Yeah, would only be unexpected for any footer icons, if or when the Bootstrap fix goes in. Sounds like we are all okay either way (even if we forget these overrides exist).

PR #87, opened 2019-06-11
ensure shortcodes work with old-Hugo and nu-Hugo

LisaFC:


LisaFC:
This should work....

RichieEscarez:
Sorry but I'm not sure why I'm having an impossible time replicating the issue (so im not able to validate this change)? I want back to Jan versions of the main _index.html and about/_index.html files (as well as the corresponding ....shortcodes/blocks/* files) but I could never get those elements to render vertically (tried on both Hugo 51 and 55.6)? Any suggestions?

LisaFC:
So about/_index.html was working fine on my local machine until I upgraded to 0.556, whereupon it stopped rendering the Markdown in those sections so they were just chunks of raw MD. I then "fixed" it with updated shortcodes, only to discover they didn't work with 0.53. The current arrangement as far as I can see works with both.

You would need to actually rewind the example site AND the template AND your Hugo version to see the original weirdly rendered horror, may not be worth it, and I'm pretty confident it's fine.

(the vertical rendering in blocks thing is a completely different issue that's got nothing to do with Hugo, it's just a default I need to switch in a forthcoming PR)

PR #89, opened 2019-06-13
Fix section styling so it's backwards-compatible

### LINKED ISSUE #83: Possible landing page bug
I've encountered unusual behavior on the landing page in the latest version of Docsy. It looks like `blocks/feature` under `blocks/section` are rendered as a vertical set rather than the expected horizontal layout:

Screen Shot 2019-06-04 at 3 32 05 PM

Here's a summary of the steps I followed to generate the screenshot:

```bash
git clone https://github.com/google/docsy-example.git **
cd docsy-example/
git submodule update --init --recursive
git submodule update --remote
npm install
npm audit fix --force
hugo server
```

** I observe essentially the same behavior with [another repo](https://github.com/LisaFC/docsy-example-test.git).

LisaFC:
Fixes #83

### DIFF
--- a/assets/scss/_boxes.scss
+++ b/assets/scss/_boxes.scss
@@ -79,7 +79,7 @@
 
 // Styling for section boxes
 .td-box {
-  .row {
+  .row.section {
        padding-left: 5rem;
        padding-right: 5rem;
        flex-direction: column;
--- a/assets/scss/_boxes.scss
+++ b/assets/scss/_boxes.scss
@@ -93,11 +93,12 @@
        }
    }
 
-  .row.features {
-   flex-direction: row;
+  .row {
+       padding-left: 5rem;
+       padding-right: 5rem;
+       flex-direction: row;
   }
 }
-
 // Styling for community page link boxes
 
 .td-box.linkbox {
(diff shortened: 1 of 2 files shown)


assets/scss/_boxes.scss
```diff
@@ -79,7 +79,7 @@
 
 // Styling for section boxes
 .td-box {
-  .row {
+  .row.section {
```
RichieEscarez: This alone resolved the rendering issues in my environment.

layouts/shortcodes/blocks/section.html
```diff
@@ -1,6 +1,6 @@
 {{ $col_id := .Get "color" | default .Ordinal }}
 {{ $height := .Get "height" | default "auto"  }}
-{{ $type   := .Get "type" | default "" }} 
+{{ $type   := .Get "type" | default "features" }}
```
RichieEscarez: While these all cause the following error:
```
hugo server
Building sites … ERROR 2019/06/18 10:15:08 error: failed to transform resource: SCSS processing failed: file 
"/usr/local/google/home/rescarez/docsytests/docsyexampletemplatetest/themes/docsy/assets/scss/blocks/_blocks.scss", 
line 1, col 1: Invalid CSS after "": expected 1 selector or at-rule, was '{{ $type := .Get "t' 
Total in 335 ms
Error: Error building site: logged 1 error(s)
```

LisaFC: Huh, odd. Let me check with my local version, I think the Docsy example site is still running with an older Hugo/Docsy. If it works without it in both then I'll remove it.

LisaFC: You're right, I didn't actually make it properly backwards compatible! Done.

Issue #88, opened 2019-06-12
Documentation issues vs project issues


sharifsalah:
Currently Docsy includes two links in the right-hand navigation (table of contents) to encourage open source contributions:

- Edit this page
- Create issue

Both of these links are configured using the `github_repo` site param in config.toml.

Arguably, 'Create issue' is somewhat ambiguous from a UX perspective, as without clicking on the link, it's not clear if the relates to creating a documentation / site issue, or a project issue. Many / most open source projects using Docsy will manage documentation in a dedicated git repo.

I propose to add a link to the open source contribution actions to encourage and make it clearer for contributors to add both project and documentation issues:

- Edit this page
- Create documentation issue
- Create project issue

I propose to keep the the `github_repo` site param for the documentation repo (so as not to impact existing sites) and add a site param `github_project_repo` (or similar) to configure an optional link to add issues to the project repo.

Here's an illustration of what this might look like:

<img width="279" alt="Screen Shot 2019-06-12 at 1 16 54 PM" src="https://user-images.githubusercontent.com/6339099/59351310-02f63e80-8d16-11e9-8f9b-6ba738d2d692.png">

I already have this working for my project, so it's fairly straightforward to contribute to Docsy. If there's enthusiasm for this proposal then I'm happy to create a PR to propose the changes myself.

rionam:
I think this is a  good idea. We made a similar clarification for g3doc,
because of similar confusion.

LisaFC:
Ooh I like it.

LisaFC:
...and by that I mean I'd love you to implement it! :)

sharifsalah:
Thanks both! I will send a PR in the next few days.

sharifsalah:
Fixed in https://github.com/google/docsy/pull/90

PR #90, opened 2019-06-14
Add link to create project issue

sharifsalah:
Updates `layouts/partials/page-meta-links.html` to add a link to create project issues.

For background context, see https://github.com/google/docsy/issues/88

I did consider using `params.links.developer` from `config.toml`, but on balance it seemed to me like some projects might prefer the flexibility to independently disable either the new **Create project issue** button or the repo link in the footer. Because **Create project issue** is just a link, it could point at any service to collect and manage customer feedback, not necessarily GitHub or a similar Git hosting service, although this would require overriding `page-meta-link.html`. (Not a great example, but think something like Google Forms)

By default, Docsy will just ignore these changes. To test them, you need to update `config.toml` to look something like the following configuration:

```toml
[params]
copyright = "Copyright Contributors to the XYZ Project"
privacy_policy = "<someurl>"
github_repo = "https://github.com/<path/to/docs/repo>"
github_project_repo = "https://github.com/<path/to/project/repo>"
```

Also, I suggested some changes to the icons for the buttons to offer some visual distinction, but I can revert to GitHub icons if that's preferred as the default.

Finally, I used Google Translate to update the Norwegian  strings in `no.toml`.

### DIFF
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -9,7 +9,7 @@
 {{ end }}
 {{ $issuesURL := printf "%s/issues/new?title=%s" $gh_repo (htmlEscape $.Title )}}
 <a href="{{ $editURL }}" target="_blank"><i class="fa fa-edit fa-fw"></i> {{ T "post_edit_this" }}</a>
-<a href="{{ $issuesURL }}" target="_blank"><i class="fas fa-book fa-fw"></i> {{ T "post_create_issue" }}</a>
+<a href="{{ $issuesURL }}" target="_blank"><i class="fab fa-github fa-fw"></i> {{ T "post_create_issue" }}</a>
 {{ if $gh_project_repo }}
 {{ $project_issueURL := printf "%s/issues/new" $gh_project_repo }}
 <a href="{{ $project_issueURL }}" target="_blank"><i class="fas fa-tasks fa-fw"></i> {{ T "post_create_project_issue" }}</a>
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -12,9 +12,7 @@
 <a href="{{ $issuesURL }}" target="_blank"><i class="fas fa-book fa-fw"></i> {{ T "post_create_issue" }}</a>
 {{ if $gh_project_repo }}
 {{ $project_issueURL := printf "%s/issues/new" $gh_project_repo }}
-{{ if $project_issueURL }}
 <a href="{{ $project_issueURL }}" target="_blank"><i class="fas fa-tasks fa-fw"></i> {{ T "post_create_project_issue" }}</a>
-{{ end }}
 {{ end }}
 </div>
 {{ end }}
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -1,5 +1,6 @@
 {{ if .Path }}
 {{ $gh_repo := ($.Param "github_repo") }}
+{{ $gh_project_repo := ($.Param "github_project_repo") }}
 {{ if $gh_repo }}
 <div class="td-page-meta ml-2 pb-1 pt-2 mb-0">
 {{ $editURL := printf "%s/edit/master/content/%s" $gh_repo .Path }}
@@ -8,7 +9,13 @@
 {{ end }}
 {{ $issuesURL := printf "%s/issues/new?title=%s" $gh_repo (htmlEscape $.Title )}}
 <a href="{{ $editURL }}" target="_blank"><i class="fa fa-edit fa-fw"></i> {{ T "post_edit_this" }}</a>
-<a href="{{ $issuesURL }}" target="_blank"><i class="fab fa-github fa-fw"></i> {{ T "post_create_issue" }}</a>
+<a href="{{ $issuesURL }}" target="_blank"><i class="fas fa-book fa-fw"></i> {{ T "post_create_issue" }}</a>
+{{ if $gh_project_repo }}
+{{ $project_issueURL := printf "%s/issues/new" $gh_project_repo }}
+{{ if $project_issueURL }}
+<a href="{{ $project_issueURL }}" target="_blank"><i class="fas fa-tasks fa-fw"></i> {{ T "post_create_project_issue" }}</a>
+{{ end }}
+{{ end }}
 </div>
 {{ end }}
 {{ end }}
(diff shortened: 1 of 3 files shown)


layouts/partials/page-meta-links.html
```diff
@@ -8,7 +9,11 @@
 {{ end }}
 {{ $issuesURL := printf "%s/issues/new?title=%s" $gh_repo (htmlEscape $.Title )}}
 <a href="{{ $editURL }}" target="_blank"><i class="fa fa-edit fa-fw"></i> {{ T "post_edit_this" }}</a>
-<a href="{{ $issuesURL }}" target="_blank"><i class="fab fa-github fa-fw"></i> {{ T "post_create_issue" }}</a>
+<a href="{{ $issuesURL }}" target="_blank"><i class="fas fa-book fa-fw"></i> {{ T "post_create_issue" }}</a>
```
RichieEscarez: I do like the options for project vs docs issues but I wonder if we are losing context without the GitHub icon (ie. that the link opens to GitHub)? 

I think we could get away with just the GitHub icon given that the first link refers to this page, so one could assume that "Create Issue" also refers to this page?

Im dont feel to strongly about this either way but the thought crossed my mind.

layouts/partials/page-meta-links.html
```diff
@@ -1,5 +1,6 @@
 {{ if .Path }}
 {{ $gh_repo := ($.Param "github_repo") }}
+{{ $gh_project_repo := ($.Param "github_project_repo") }}
```
RichieEscarez: Your PR comment suggested that this new "link" is intended to be a default generic link option for the theme (vs. all these links pointing to only GitHub), but these variables are qualified with "gh" and "github". Do you want the variables to be generic as well?

RichieEscarez:
I've tested this all and it works great. I have a couple open comments/questions. Otherwise, @LisaFC do you have any feedback too?

layouts/partials/page-meta-links.html
sharifsalah: I've reverted to the GitHub icon for the "Create documentation link". I also don't feel strongly about it either, as long as there is some visual distinction between the icons and it's easy to override if you prefer something else.

sharifsalah:
Thanks for the feedback, I think I've addressed both your comments about the files included in the PR. However for config.toml, I'm not sure which file you're referring to? I did notice that Docsy, has a [config.toml](https://github.com/google/docsy/blob/master/config.toml), but that doesn't look like it's relevant here.

If you mean config.toml in one of the sample implementations of Docsy, there are a few of those in different repos (e.g. [docsy-example](https://github.com/google/docsy-example)) and I'm not sure which is the canonical one. I'm happy to send a separate PR to the correct repo!

RichieEscarez:
> Thanks for the feedback, I think I've addressed both your comments about the files included in the PR. However for config.toml, I'm not sure which file you're referring to? I did notice that Docsy, has a [config.toml](https://github.com/google/docsy/blob/master/config.toml), but that doesn't look like it's relevant here.
> 
> If you mean config.toml in one of the sample implementations of Docsy, there are a few of those in different repos (e.g. [docsy-example](https://github.com/google/docsy-example)) and I'm not sure which is the canonical one. I'm happy to send a separate PR to the correct repo!

Oh interesting, i didn't notice that the [google/docsy config.toml](https://github.com/google/docsy/blob/master/config.toml) was a blank page. 

@LisaFC do we have a list of Docsy specific config.toml definitions (reference guide/page)? Is the file in the google/docsy-example repo, the only config.toml that includes values and comments? Should the [google/docsy config.toml](https://github.com/google/docsy/blob/master/config.toml) file include all docsy specific variables and/or default values (even if they are commented out)?

LisaFC:
Yeah, the google/docsy theme doesn't have a config.toml because themes can't have real config files. Per the Hugo docs: "A theme component can also provide its own Configuration File, e.g. config.toml. There are some restrictions to what can be configured in a theme component, and it is not possible to overwrite settings in the project" (ie if we did a theme config we would hardcode in loads of things for our sites that they probably don't want, eg analytics settings, etc.)

I'm currently trying to consolidate all our various versions of config.toml - the canonical one should be in docsy-example eventually. For now push any changes to my copy of the example site (which is where we're serving Goldydocs), as that's going to end up in docsy-example: https://github.com/LisaFC/docsy-example-test

And no, we don't have a reference, I'll have a look for next release if there's some way of generating a reference from a .toml. There are waaaaaaay too many params to try maintaining one manually.

sharifsalah:
That's great, thanks both.

@LisaFC I'll send you a PR to update confg.toml in https://github.com/LisaFC/docsy-example-test

@RichieEscarez is there anything else to do for this PR?

thanks again,

LisaFC:
Thanks!

In the absence of a mechanism for creating a config.toml reference, I wonder could we use one of our project friends' (Agones, Istio) code widgets to just pull the entire thing in from the repo on a reference page?

RichieEscarez:
> In the absence of a mechanism for creating a config.toml reference, I wonder could we use one of our project friends' (Agones, Istio) code widgets to just pull the entire thing in from the repo on a reference page?

the [readFile function](https://gohugo.io/functions/readfile/) could work for this

> Yeah, the google/docsy theme doesn't have a config.toml because themes can't have real config files. Per the Hugo docs: "A theme component can also provide its own Configuration File, e.g. config.toml. There are some restrictions to what can be configured in a theme component, and it is not possible to overwrite settings in the project" (ie if we did a theme config we would hardcode in loads of things for our sites that they probably don't want, eg analytics settings, etc.)
> 
> I'm currently trying to consolidate all our various versions of config.toml - the canonical one should be in docsy-example eventually. For now push any changes to my copy of the example site (which is where we're serving Goldydocs), as that's going to end up in docsy-example: https://github.com/LisaFC/docsy-example-test
>
> And no, we don't have a reference, I'll have a look for next release if there's some way of generating a reference from a .toml. There are waaaaaaay too many params to try maintaining one manually.

Yeah there are a lot in total, but I was thinking 'not every parameter' just the Docsy specific `params` that are not already documented by Hugo (ie. add a comment that links to the [configuration](https://gohugo.io/getting-started/configuration/)  ref page).

sharifsalah:
Thanks! And PR is already done (although to docsy-example-test)

https://github.com/LisaFC/docsy-example-test/pull/1

sharifsalah:
Oopps, I clicked **Close and comment** instead of **Comment**...

RichieEscarez:
> Thanks! And PR is already done (although to docsy-example-test)
> 
> [LisaFC/docsy-example-test#1](https://github.com/LisaFC/docsy-example-test/pull/1)

Awesome! I've add some comments there. Thanks @sharifsalah!

PR #94, opened 2019-06-24
Update page-meta-links.html

### LINKED ISSUE #32: page-meta-links.html assumed /content is in root
Content:
https://github.com/google/docsy/blob/master/layouts/partials/page-meta-links.html

Specifically if you look at:
```gotemplate
{{ $editURL := printf "%s/edit/master/content/%s/%s" $gh_repo ($.Site.Language.Lang) .Path }}
```
You can see that the editURL assums that the $gh_repo is the root -- when it may not be.

Solution - have a subdir toml config that can be set, and default to "" for the root path of the directory.

LisaFC:
Fixes #32 

(the logic is very unattractive but there you Go, as it were. Suggestions for making it less horrible welcome.)

### DIFF
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -1,11 +1,17 @@
 {{ if .Path }}
 {{ $gh_repo := ($.Param "github_repo") }}
-{{ $gh_project_repo := ($.Param "github_project_repo") }}
+{{ $gh_subdir := ($.Param "github_subdir") }}
 {{ if $gh_repo }}
 <div class="td-page-meta ml-2 pb-1 pt-2 mb-0">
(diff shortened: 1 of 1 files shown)


RichieEscarez:
@LisaFC This change causes the following error:

```
$ hugo server
Error: "/usr/local/google/home/rescarez/docsytests/docsyexampletemplatetest/themes/docsy/layouts/partials/page-meta-links.html:19:1": parse failed: template: partials/page-meta-links.html:19: undefined variable "$gh_project_repo"
```

Lines 19 and 20 both still use that variable:
```
{{ if $gh_project_repo }}
{{ $project_issueURL := printf "%s/issues/new" $gh_project_repo }}
```

LisaFC:
That is because I am a fool and did the commit that fixes this in the wrong branch. :(

The version that's in now doesn't have this error.

Issue #30, opened 2019-02-12
Allow an additional dropdown in the top menu bar via config


sarahmaddox:
For the Kubeflow site I needed to add a dropdown menu offering different versions of the docs: https://www.kubeflow.org

The Agones site needed to add a similar option:
https://agones.dev/site/docs/

To do this, I had to add a custom navbar partial. It'd be nice it we could add a dropdown via config rather than needing to override the Docsy layouts.

LisaFC:
Fixed in #95

PR #95, opened 2019-06-28
Update navbar.html

LisaFC:
Add optional drop down list for versions

### DIFF
--- a/layouts/partials/navbar-version-selector.html
+++ b/layouts/partials/navbar-version-selector.html
@@ -1,0 +1,8 @@
+<a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+	{{ .Site.Params.version_menu }}
+</a>
+<div class="dropdown-menu" aria-labelledby="navbarDropdownMenuLink">
+	{{ range .Site.Params.versions }}
+	<a class="dropdown-item" href="{{ .url }}">{{ .version }}</a>
+	{{ end }}
+</div>
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -1,7 +1,7 @@
 {{ $cover := .HasShortcode "blocks/cover" }}
 <nav class="js-navbar-scroll navbar navbar-expand navbar-dark {{ if $cover}} td-navbar-cover {{ end }}flex-column flex-md-row td-navbar">
-	<a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
-		<span class="navbar-logo">{{ if (fileExists "icons/logo.svg") }}{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}{{ end }}</span><span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>
+        <a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
+		<span class="navbar-logo">{{ if .Site.Params.ui.navbar_logo }}{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}{{ end }}</span><span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>
 	</a>
 	<div class="td-navbar-nav-scroll ml-md-auto" id="main_navbar">
 		<ul class="navbar-nav mt-2 mt-lg-0">
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -1,7 +1,7 @@
 {{ $cover := .HasShortcode "blocks/cover" }}
 <nav class="js-navbar-scroll navbar navbar-expand navbar-dark {{ if $cover}} td-navbar-cover {{ end }}flex-column flex-md-row td-navbar">
 	<a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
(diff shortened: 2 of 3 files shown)


LisaFC:
Fixes #30 once the change is documented.

Requires a version_menu param and some [params.versions], as in the kubeflow config:
https://github.com/kubeflow/website/blob/master/config.toml

LisaFC:
@sarahmaddox @markmandel

sarahmaddox:
Thanks @LisaFC! Please would you point me at a preview? I've looked at the preview generated by Netlify, and at the "Example Site" site that it points to, but neither of those has a version dropdown.

https://deploy-preview-95--docsydocs.netlify.com/

layouts/partials/navbar-version-selector.html
```diff
@@ -0,0 +1,8 @@
+<a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+	{{ .Site.Params.version_menu }}
```
RichieEscarez: This `version_menu` parameter is not used by Kubeflow (and therefore renders blank). Is this a separate "menu only" value that you want set? 

Kubeflow has a `version` value that is used in each repo branch to indicate that site's version. In my testing, if i rename this to `.Site.Params.version`, then it works as i initially expected (ie. i used below, and therefore the main nav bar showed `master` next to the dropdown menu).

My example config.toml

```
version = "master"
githubbranch = "master"

# Add new release versions here
[[params.versions]]
  version = "master"
  githubbranch = "master"
  url = "https://master.kubeflow.org"

[[params.versions]]
  version = "v0.2"
  githubbranch = "v0.2-branch"
  url = "https://v0-2.kubeflow.org"

[[params.versions]]
  version = "v0.3"
  githubbranch = "v0.3-branch"
  url = "https://v0-3.kubeflow.org"

[[params.versions]]
  version = "v0.4"
  githubbranch = "v0.4-branch"
  url = "https://v0-4.kubeflow.org"

[[params.versions]]
  version = "v0.5"
  githubbranch = "v0.5-branch"
  url = "https://v0-5.kubeflow.org"
```

layouts/partials/navbar.html
```diff
@@ -1,7 +1,7 @@
 {{ $cover := .HasShortcode "blocks/cover" }}
 <nav class="js-navbar-scroll navbar navbar-expand navbar-dark {{ if $cover}} td-navbar-cover {{ end }}flex-column flex-md-row td-navbar">
 	<a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
-		<span class="navbar-logo">{{ if .Site.Params.ui.navbar_logo }}{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}{{ end }}</span><span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>
+		<span class="navbar-logo">{{ if (fileExists "icons/logo.svg") }}{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}{{ end }}</span><span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>
```
RichieEscarez: This would not work for me unless I added the `/assets/` path:

```suggestion
		<span class="navbar-logo">{{ if (fileExists "assets/icons/logo.svg") }}{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}{{ end }}</span><span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>
```

LisaFC: That's weird because I didn't change those in this commit, that's the existing navbar code in the repo and it's working fine in both goldydocs and the Docsy guide. Let me check it again....

layouts/partials/navbar-version-selector.html
LisaFC: True, but I didn't make this change to be backwards compatible with Kubeflow's customizations (even though I borrowed from them), I made it to be backwards compatible with existing Docsy sites - I don't actually think having the current version name as the button/menu name is idea for all cases, especially if you look at some of the older "live" Kubeflow sites (https://v0-4.kubeflow.org/), they all just have a drop down called "master", which may not make it obvious what you can do with the menu. (unless this isn't the intended behaviour?) The equivalent menu in Agones is called Releases.

There's a separate menu name variable called version_menu, which I'll add to the example site config and the docs.

layouts/partials/navbar.html
LisaFC: Oh hang on, I didn't mean to change that in this commit, this is what happens when you have old changes knocking around a branch, that's actually old code. I'll revert.

(it must have worked for me locally, hence I didn't notice....)

LisaFC: Fixed!

layouts/partials/navbar-version-selector.html
LisaFC: We can demonstrate it in the example site but the menu wouldn't point to anything as we do not have any old versions....

LisaFC:
> Thanks @LisaFC! Please would you point me at a preview? I've looked at the preview generated by Netlify, and at the "Example Site" site that it points to, but neither of those has a version dropdown.
> 
> https://deploy-preview-95--docsydocs.netlify.com/

Yeah, the preview won't work because the site in this theme repo doesn't have any previous versions specified so Hugo doesn't render the menu. Working as intended, but probably not a very good demo. :)

I'll see if I can put something up on Netlify that has the menu by adding some fake old version URLs. It works fine locally.

LisaFC:
Here's it being previewed on another site: https://serene-roentgen-10fbaa.netlify.com/docs/

layouts/partials/navbar-version-selector.html
LisaFC: I've also added a commit for the instructions.

RichieEscarez: Sounds good. Thanks Lisa

PR #107, opened 2019-07-17
Add German translation of UI strings

gesellix:


LisaFC:
Thanks so much for this!

Issue #19, opened 2019-01-30
Add theme to Hugo theme gallery


LisaFC:

LisaFC:
Ta da! https://themes.gohugo.io/docsy/

PR #108, opened 2019-07-18
Create theme.toml

LisaFC:
Needed to add the theme to the Hugo theme gallery, per #19


LisaFC:
We probably do want a screencap too! I'll merge in this first anyway...

PR #121, opened 2019-08-14
Upgrade font awesome to the latest version

Fox32:
There where some breaking changes in the usage of FontAwesome, so I had to modify the scss. Sadly we have to copy the font files into the theme.

The documentation wants me to override `td-enable-google-fonts` if I want to use a custom font, but is still loading it as overriding isn't possible. This is also fixed with this PR.

I also added crosssorigin anonymous to the script tag, otherwise it's not possible to let Netlify put it on an external CDN.


LisaFC:
Thanks for this!

chalin:
@Fox32 - FYI, the following PR proposed that we drop the webfonts. WDYT?

- #1164

Also related is:

- #1102

Issue #118, opened 2019-08-04
Add link icons to headers for easy link-grabbing


sarahmaddox:
A contributor to the [Kubeflow docs](https://www.kubeflow.org/docs/about/kubeflow/) - @Bobgy - has recently added auto-generated link icons on every heading. This makes it easy for people to grab the link to a particular section of a doc. (People can grab the same link from the TOC on the right of the page, but many people don't notice the TOC.)

Here's the PR that adds the link icons:
https://github.com/kubeflow/website/pull/1009

**Inviting @Bobgy to contribute this feature to Docsy.**

Bobgy:
Thanks for creating the issue here. I'm happy to contribute here too.

For user requirement, do you think any website that uses docsy will want to turn off this feature?

For implementation,
I briefly viewed folder structure here. I think I should put my script in `assets/js/anchor.js`, then reference it in `layouts/partials/scripts.html`. Do you think that's a good idea? Or maybe right inside `assets/js/base.js`?

LisaFC:
1)I can't imagine anyone *not* wanting link anchors, they're unobtrusive and very useful, so as long as this change doesn't break existing sites I think I'm happy to not have an opt-out for now (if someone requests it we can always put it in)

2)anchor.js works for me!

Thanks for doing this!

Bobgy:
Sounds good! I'll work on this soon.

PR #123, opened 2019-08-30
Add anchor link which shows on hover to markdown headers

Bobgy:
Fixes https://github.com/google/docsy/issues/118
/assign @LisaFC

### DIFF
--- a/assets/js/anchor.js
+++ b/assets/js/anchor.js
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+(function ($) {
+    'use strict';
+
+    // Headers' anchor link that shows on hover
+    $(function () {
+        // append anchor links to headings in markdown.
+        var article = document.getElementsByTagName('main')[0];
+        if (!article) {
+            return;
+        }
+        var headings = article.querySelectorAll('h1, h2, h3, h4, h5, h6');
+        headings.forEach(function (heading) {
+            if (heading.id) {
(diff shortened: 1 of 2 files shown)


layouts/partials/scripts.html
```diff
@@ -2,8 +2,9 @@
 <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.3/umd/popper.min.js" integrity="sha384-ZMP7rVo3mIykV+2+9J3UJ46jBk0WLaUAdn689aCwoqbBJiSnjAK/l8WvCWPIPm49" crossorigin="anonymous"></script>
 <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js" integrity="sha384-ChfqqxuZUCnJSK3+MXmPNIyE6ZbWh2IMqE241rYiqJxyMiZ6OW/JmZQ5stwEULTy" crossorigin="anonymous"></script>
 {{ $jsBase := resources.Get "js/base.js" }}
+{{ $jsAnchor := resources.Get "js/anchor.js" }}
```
Bobgy: I'm not sure if this is correct. What does the second part of $jsSearch do?

assets/js/anchor.js
```diff
@@ -0,0 +1,49 @@
...
+    'use strict';
+
+    // Headers' anchor link that shows on hover
+    $(function () {
+        // append anchor links to headings in markdown.
+        var article = document.getElementsByTagName('main')[0];
+        if (!article) {
+            return;
+        }
+        var headings = article.querySelectorAll('h1, h2, h3, h4, h5, h6');
```
Bobgy: not using jQuery pattern because I'm not familiar with it. I've seen some other code here using native api too. So I assume this is fine. Let me know if you have concerns.

layouts/partials/scripts.html
LisaFC: Hrmm, I didn't write that bit but as far as I can see it's because search.js has Go template stuff (ie stuff in {{ }}) in it so we need to render it with ExecuteAsTemplate before using it. Your script doesn't so you can just use it "as is" (ie you are right!)

See: https://regisphilibert.com/blog/2018/07/hugo-pipes-and-asset-processing-pipeline/#customizing-our-sass-variables-with-executeastemplate

assets/js/anchor.js
LisaFC: Fine with me.

LisaFC:
This is fantastic! Thanks so much.

PR #125, opened 2019-09-09
Fix supported browsers config

### LINKED ISSUE #120: Warning `Replace Autoprefixer browsers option to Browserslist config.`
# Steps to reproduce

Run `hugo --config config.toml` from the root of the project.

Results in the following output:
```
Building sites …
  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.

  Using browsers option cause some error. Browserslist config
  can be used for Babel, Autoprefixer, postcss-normalize and other tools.

  If you really need to use option, rename it to overrideBrowserslist.

  Learn more at:
  https://github.com/browserslist/browserslist#readme
  https://twitter.com/browserslist
                   | EN
-------------------+-----
  Pages            |  7
  Paginator pages  |  0
  Non-page files   |  0
  Static files     | 36
  Processed images |  0
  Aliases          |  0
  Sitemaps         |  1
  Cleaned          |  0
```
# What you expected to happen
No warning messages.

# Versions
Autoprefixer is at the latest version 9.6.1

LisaFC:
Fixes #120 (I think)

This definitely works locally, though would appreciate a check from someone who knows more about Autoprefixer than I do...


dholbach:
Checking out your branch, I run into the following problem:

```
Building sites … You did not set any plugins, parser, or stringifier. Right now, PostCSS does nothing. Pick plugins for your case on https://www.postcss.parts/ and use them in postcss.config.js.
```

You might need a minimal `postcss.config.js` like:

```js
module.exports = {
  plugins: [
    require('autoprefixer')
  ]
}
```

I'm not an expert though.

coliff:
I think this PR can be closed now. This was fixed in #295.

It now uses Browserslist default config (which you can see here: https://browserl.ist/?q=defaults)

Issue #126, opened 2019-09-16
Support for Algolia Docsearch as alternative to GCSE?


alexearnshaw:
Wondering if this project would be interested in offering support for an alternative search - [Algolia DocSearch](https://community.algolia.com/docsearch/) is free for public documentation projects and is an easy to implement alternative to Google GCSE which is not ad-supported.

You basically just sent a request to Algolia to be added to the program, and (if accepted) they email you back some Javascript code which you add to the head end and body end.

To get this working in my project I ended up adding a new config parameter to config.toml and then overriding the search-input partial in my own project. However, I think the same end result could be achieved by just adding an additional `with` statement to Docsy's search-input partial?

Also I wasn't able to get Algolia's search working with the Search box on the left nav/sidebar so I disabled this in config.toml but it works fine on the top nav search box.

You can see the changes I made to get this working (after getting the Javascript code for my site from Algolia) in this PR https://github.com/Axway/axway-open-docs/pull/113

You can see the end result and try out the Algolia search here https://axway-open-docs.netlify.com/

If there's interest I'd be happy to submit a PR to this project with the needed changes/doc.

LisaFC:
Oh that's really nice! Yes, it would be great if we could provide this as an option.

alexearnshaw:
@LisaFC I've had a go at this in PR #127, I'd appreciate your input.

LisaFC:
Oh brilliant, I'll take a look!

LisaFC:
Fixed in #127

PR #127, opened 2019-09-17
Add support for Algolia DocSearch

alexearnshaw:
As discussed in #126 this is a PR for using Algolia DocSearch as an alternative to GCSE.

To test that the search displays correctly for the Docsy user guide I've used the Algolia Javascript from my site in the Docsy head end and body end. I've verified this works locally, but of course it display search results from my site. **These changes will need to be removed before this PR is merged.**

Similarly the config.toml changes are added here only for testing purposes and **need to be removed before merging**. I guess the new param needs to be added in the Docsy example project instead?

The change I've made to search-input.html works but is probably not the best way to do it? I'd appreciate any guidance here :)

### DIFF
--- a/layouts/partials/hooks/body-end.html
+++ b/layouts/partials/hooks/body-end.html
@@ -1,0 +1,9 @@
+<!-- scripts for algolia docsearch -->
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
+<script type="text/javascript"> docsearch({
+apiKey: '0fe272b76fcdd3184f3bc193bd96ca15',
+indexName: 'axway-open-docs',
+inputSelector: '.td-search-input',
+debug: false // Set debug to true if you want to inspect the dropdown
+});
+</script> 
--- a/layouts/partials/hooks/body-end.html
+++ b/layouts/partials/hooks/body-end.html
@@ -1,10 +1,3 @@
 {{ with .Site.Params.algolia_docsearch }}
-<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
-<script type="text/javascript"> docsearch({
-apiKey: '0fe272b76fcdd3184f3bc193bd96ca15',
-indexName: 'axway-open-docs',
-inputSelector: '.td-search-input',
-debug: false // Set debug to true if you want to inspect the dropdown
-});
-</script> 
+<!-- scripts for algolia docsearch -->
 {{ end }}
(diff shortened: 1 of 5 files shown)


LisaFC:
Aha, I was wondering why I was getting your site's Algolia search results in the preview search box until I actually read your message properly. Do you mind me asking how long it took to get your Algolia search setup, as I'm going to apply for Docsy as an OSS project? It would be nice for users who choose this feature to be able to see it working on our site or a local copy once they toggle the search engine.

I'll take a proper look at your implementation now.

LisaFC:
Also thanks so much for doing this and sorry for the delay in reviewing!

alexearnshaw:
No problem @LisaFC :) It only took a couple of days for Algolia to get back to me with the Javascript. First they'll send you an email to verify that you are a site owner/maintainer and once I replied to that I got the Javascript the next day.

layouts/partials/hooks/body-end.html
```diff
@@ -0,0 +1,9 @@
+<!-- scripts for algolia docsearch -->
```
LisaFC: I think we also need to stick this in a conditional because otherwise even if you set algolia_docsearch to false, if the scripts are here and you enable GCSE, you get the Algolia results from the search box rather than the GCSE ones.

I know the current file is empty but you could instead have the committed version look like

{{ with .Site.Params.algolia_docsearch  }}
&gt;!-- scripts for algolia docsearch --&lt;
{{ end }}

Does that make sense in terms of user experience? I don't want to go too bananas with conditionals but at the same time we don't want people getting unintuitive results. Also if you accidentally enable both in config.toml, the Algolia one currently wins if it has scripts there. We may need to do some rewording in the docs.

userguide/config.toml
```diff
@@ -97,14 +97,19 @@ time_format_default = "January 2, 2006"
 rss_sections = ["blog"]
 
 # Google Custom Search Engine ID. Remove or comment out to disable search.
-gcs_engine_id = "011217106833237091527:la2vtv2emlw"
+#gcs_engine_id = "011217106833237091527:la2vtv2emlw"
+
+# Enable Algolia DocSearch
+algolia_docsearch = true
```
LisaFC: Yes, add this to the example project as well (with both set to "false", though I actually like the Algolia results so much I'm toying with the idea of making them the default...)

alexearnshaw:
@LisaFC thanks for the feedback I'll get back to this as soon as I can, but it will likely be a few days.

LisaFC:
No rush, thanks again for doing this! :)

alexearnshaw:
@LisaFC I've made the suggested changes now. I tested this locally and on netlify preview with my own Javascript code, and now I've removed the Javascript and just left the comments. This should be ready to merge now unless I've missed something ;)

Also did a PR for the config file on the example site https://github.com/google/docsy-example/pull/30

LisaFC:
Ah, you're a star, this is great!

LisaFC:
BTW, going to big this nice new feature up to our not very many followers on social media (https://twitter.com/docsydocs), do you mind me using your name or do you have a public Twitter?

alexearnshaw:
Cool,  I'm not  on twitter so you can use my name np. Thanks :)

PR #135, opened 2019-09-25
Markdownify section description

markmandel:
The description for a section can contain markdown content, so this adds `markdownify`so it will be rendered correctly.


LisaFC:
Thanks Mark!

Issue #140, opened 2019-10-09
Security issues Reported


bashfulrobot:
Hi there, I use the security scans provided by GitHub, and the following issues were reported on my repo... I thought you may want to know.

Thank you.

### Remediation

Upgrade set-value to version 2.0.1 or later. For example:

```
"dependencies": {
  "set-value": ">=2.0.1"
}
```

or…

```
"devDependencies": {
  "set-value": ">=2.0.1"
}
```

Always verify the validity and compatibility of suggestions with your codebase.

### Details

CVE-2019-10747
More information
high severity
Vulnerable versions: < 2.0.1
Patched version: 2.0.1

set-value is vulnerable to Prototype Pollution in versions before 2.0.1 and version 3.0.0. The function mixin-deep could be tricked into adding or modifying properties of Object.prototype using any of the constructor, prototype and proto payloads.

### Remediation

Upgrade mem to version 4.0.0 or later. For example:

```
"dependencies": {
  "mem": ">=4.0.0"
}
```

or…

```
"devDependencies": {
  "mem": ">=4.0.0"
}
```

Always verify the validity and compatibility of suggestions with your codebase.

### Details
WS-2018-0236
More information
moderate severity
Vulnerable versions: < 4.0.0
Patched version: 4.0.0

In nodejs-mem before version 4.0.0 there is a memory leak due to old results not being removed from the cache despite reaching maxAge. Exploitation of this can lead to exhaustion of memory and subsequent denial of service.

bashfulrobot:
Thank you for your time.

PR #145, opened 2019-10-31
npm audit fix to bump for CVEs

asmacdo:
GitHub's security feature informed me that some of the deps needed a bump.

This patch was created by npm audit fix.

Partially fixes google/docsy#140 npm audit did not take care of the mem package.


LisaFC:
Thanks so much!

PR #150, opened 2019-11-17
Support sidebar search on offline search

tekezo:
- Support sidebar search on offline search.
- Improve result rendering when scroll is required by many results on offline search.
  For example, searching "tasks" on docsy-example site.


LisaFC:
Lovely! Thanks for this!

Can you update the Lunr search section in this doc too to remove the step saying you need to disable sidebar search?
https://github.com/google/docsy/blob/master/userguide/content/en/docs/Adding%20content/navigation.md

tekezo:
Thank you for your reviewing and other merges!
I updated the userguide. Please confirm the change.

PR #152, opened 2019-11-24
Offline search improvements

tekezo:
This change improves offline search:

- Update lunr.js to v2.3.8.
- Use lunr.min.js.
- Multi-language Support on offline search.
  (eg., searching "TechOS" on docsy-example matches Norsk pages.)
- Add fingerprint to offline search index json file name to prevent use old index file after site contents are updated.
- There is no need to provide search-index.md file to generate offline-search-index.json file.
- Fix an issue that search does not work when query is input before lunr indexing is finished.

And code refactoring:

- Remove global variables from offline-search.js.
- Apply code formatter (prettier) to offline-search.js.
- Load search.js only if offline search is disabled since search.js and offline-search.js are conflicting around enter key handling. (Removed a workaround code from offline-search.js for this conflicting.)

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,5 @@
+{
+    "semi": true,
+    "singleQuote": true,
+    "tabWidth": 4
+}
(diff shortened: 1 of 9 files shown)


tekezo:
Should I separate this PR into small PRs?

LisaFC:
No, this is great, sorry it's taken me a while to review it!

tekezo:
Thank you for reviewing PR and merging it!

PR #155, opened 2019-12-02
Correct copyright attribution

ocdtrekkie:
"Global Inc" is presumably a placeholder for a demo project. However, the privacy policy in this project is for Google, and of course, https://www.docsy.dev/ itself appears to show "Global Inc" as the copyright, though it should, in actuality, presumably say Google Inc.

I noticed https://example.docsy.dev/ utilizes "The Docsy Authors" as it's copyright attribution though, so presumably that would be the best copyright here as well.

### DIFF
--- a/userguide/config.toml
+++ b/userguide/config.toml
@@ -80,7 +80,7 @@ weight = 1
 # Everything below this are Site Params
 
 [params]
-copyright = "Global Inc"
+copyright = "The Docsy Authors"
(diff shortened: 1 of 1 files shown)


ocdtrekkie:
This is a trivial non-code update and shouldn't require I figure out how to sign a CLA for it with the commit address which isn't a Gmail account.

ocdtrekkie:
@googlebot I signed it!

PR #161, opened 2019-12-14
Boosting title field at offline search

tekezo:
This change increase title field priority at offline search.
(Pages which contains search keywords in title are given priority than keywords in body.)


tekezo:
Thank you for merging!

PR #163, opened 2019-12-17
Add fix for Windows formatted path values

ThomasMPorter:
Unfortunately, if you render docsy under Hugo for Windows it renders the .Path value with backslashes, which translates to `%5c` values in the resulting github URL. 

This change aims to replace those with forward slashes to make the URLs work correctly.


LisaFC:
This is great, thanks!

PR #166, opened 2019-12-28
Allow td-box-colors to be overridden

shartte:
While the comment says the color list for "paint by numbers" can be overriden, this didn't actually work for me and I suspect it's because of the missing !default annotation.
So after adding it, I was able to override it as expected in my variables file, here's the corresponding change.


LisaFC:
Thanks for spotting this!

PR #171, opened 2020-01-12
docs: fix links

dholbach:
I found a couple of 404s in the docs. Might be worth adding something like `linkchecker` to CI?


LisaFC:
Thanks so much for this! And yes, will look into a linkchecker, or at least recommend running it on the Netlify preview.

dholbach:
Thanks for merging and docsy itself!

For Flux I did the following in terms of link-checking, not sure how applicable this is for docsy:
- https://github.com/fluxcd/flux/blob/master/.circleci/config.yml#L118
- https://github.com/fluxcd/flux/blob/master/Makefile#L176-L180
- https://github.com/fluxcd/flux/blob/master/docs/Dockerfile#L5-L16

Issue #181, opened 2020-01-24
Offline search does not work with Hugo "relativeURLs" config parameter


chrvo:
With the Hugo relativeURLs param, hugo replaces all relative urls by urls relative to the current content. (https://gohugo.io/content-management/urls/#relative-urls)
However, this does not work with the search index file which is always "/offline-search-index.HASH.json"

LisaFC:
@tekezo do you have any ideas here?

tekezo:
Thank you for feedback and notifying me!
I confirmed this issue. I'll make a PR.

PR #182, opened 2020-01-24
Fix an issue that offline search does not work when relativeURLs is enabled

tekezo:
Fix https://github.com/google/docsy/issues/181

### DIFF
--- a/layouts/partials/search-input.html
+++ b/layouts/partials/search-input.html
@@ -10,9 +10,9 @@
   aria-label="{{ T "ui_search" }}"
   autocomplete="off"
   {{/* 
-    The data attribute name of json file URL have to be ended with `src` since
+    The data attribute name of the json file URL must end with `src` since
     Hugo's absurlreplacer requires `src`, `href`, `action` or `srcset` suffix for the attribute name.
-    If the absurlreplacer is not applied, URL will be started with `/`.
+    If the absurlreplacer is not applied, the URL will start with `/`.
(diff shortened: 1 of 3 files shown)


layouts/partials/search-input.html
```diff
@@ -1,14 +1,22 @@
...
 {{ $offlineSearchIndexFingerprint := $offlineSearchIndex | resources.Fingerprint "sha512" }}
 <input
   type="search"
   class="form-control td-search-input"
   placeholder="&#xf002; {{ T "ui_search" }}"
   aria-label="{{ T "ui_search" }}"
   autocomplete="off"
-  data-offline-search-index-json-url="{{ $offlineSearchIndexFingerprint.Permalink }}"
+  {{/* 
+    The data attribute name of json file URL have to be ended with `src` since
```
LisaFC:
```suggestion
    The data attribute name of the json file URL must end with `src` since
```

layouts/partials/search-input.html
```diff
@@ -1,14 +1,22 @@
...
+    Hugo's absurlreplacer requires `src`, `href`, `action` or `srcset` suffix for the attribute name.
+    If the absurlreplacer is not applied, URL will be started with `/`.
```
LisaFC:
```suggestion
    If the absurlreplacer is not applied, the URL will start with `/`.
```

tekezo:
Thank you for reviewing!
I resolved them.

LisaFC:
Thanks so much!

PR #189, opened 2020-02-02
Added a layout partial for a version banner

sarahmaddox:
I've added 3 PRs suggesting a new Docsy feature: a banner that people can display on archived doc sets, letting readers know that they're viewing an unmaintained snapshot and pointing them to the latest docs. I hope this is useful for Docsy. We've added the banner to Kubeflow and it's working well. Please would you take a look and let me know what you think? The PRs are:

* https://github.com/google/docsy/pull/189 Added a layout partial for a version banner 
* https://github.com/google/docsy-example/pull/36 Added parameters for version banner
* https://github.com/google/docsy/pull/192 Added user guide for versioning docs

Fixes https://github.com/google/docsy/issues/190

Resulting banner, tested in the Docsy Example site locally:
image

----
And here's the banner in the Docsy User Guide, served in the Netlify preview - I set the `archived_version` parameter to `true` temporarily, to test it:
image


LisaFC:
This is super, thanks!

PR #202, opened 2020-02-15
Improve offline search UX on mobile browsers

tekezo:
Hide keyboard after search keyword is entered in order to obtain screen area to show search result.


tekezo:
Thank you for reviewing and merging!

Issue #204, opened 2020-02-16
Error Cloning the Docsy theme to my project’s themes subdirectory


aborruso:
Hi,
these are my steps:

- first of all read https://www.docsy.dev/docs/getting-started/#cloning-the-docsy-theme-to-your-project-s-themes-subdirectory
- then `cd themes`
- then `git clone https://github.com/google/docsy`
- then `cd ..`
- and `hugo`

I have this error:

```
Building sites … ERROR 2020/02/16 10:30:52 Transformation failed: SCSS processing failed: file "stdin", line 6, col 1: File to import not found or unreadable: ../vendor/bootstrap/scss/bootstrap.
```

Probably it's necessary to add some note in this [page](https://www.docsy.dev/docs/getting-started/#cloning-the-docsy-theme-to-your-project-s-themes-subdirectory), but I do not know which. To add fontawesome and bootstramp repo??

Thank you

LisaFC:
Yep, you need fontawesome and bootstrap locally to use Hugo on your local machine, basically you need to recurse everything down (the two repos are submodules of the Docsy theme itself).

It's documented here:
https://www.docsy.dev/docs/deployment/#serving-your-site-locally

aborruso:
I have added a PR https://github.com/google/docsy/pull/206

Thank you

LisaFC:
Thanks!! I'll close this once your PR is in.

deitch:
Is this somewhat out of date? I had been trying using docsy as a git submodule, so:

```
git submodule update --init --recursive
```

but I still get the errors, no submodules cloned, same missing `../vendor/...` error. So I looked at the docsy source, and I see no `.gitmodules` file at all.

Is submodules no longer supported?

deitch:
Eh, never mind. I switched to hugo modules. Still cannot get dependencies working, but I will find a relevant issue or open a new one.

wtchangdm:
(Sorry for digging this old issue out again)

Hi @deitch, have you found any solutions? I switched to hugo module but can't get it work.

laurentforet:
I had the same issue, in order to make it works, I had to follow the installation pre-requisite (https://www.docsy.dev/docs/get-started/docsy-as-module/installation-prerequisites/#install-postcss) in the themes/docsy folder

PR #206, opened 2020-02-18
Added a necessary step

aborruso:
see #204

### DIFF
--- a/userguide/content/en/docs/Getting started/_index.md
+++ b/userguide/content/en/docs/Getting started/_index.md
@@ -201,6 +201,7 @@ To clone Docsy into your project's `theme` folder, run the following commands fr
 ```
 cd themes
 git clone https://github.com/google/docsy
+git submodule update --init --recursive
 ```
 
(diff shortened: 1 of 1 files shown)


userguide/content/en/docs/Getting started/_index.md
```diff
@@ -201,6 +201,7 @@ To clone Docsy into your project's `theme` folder, run the following commands fr
 ```
 cd themes
 git clone https://github.com/google/docsy
+git submodule update --init --recursive
```
LisaFC: Good call, I know we document this elsewhere but users might try running Hugo (as you did!) before they see it.

Might just expand on it a bit so they know why they're doing it, so:

```
cd themes
git clone https://github.com/google/docsy
```
If you want to build and/or [serve your site locally](/docs/deployment/#serving-your-site-locally), you also need to get local copies of the theme’s own submodules:

```
git submodule update --init --recursive
```

aborruso: I have added it

PR #209, opened 2020-02-19
Add reading time:

### LINKED ISSUE #122: Add Reading Time support
This is a feature request.

I'd like to be able to optionally display the estimated reading time for a documentation page or blog post.

It seems like  Hugo has a page variable for Reading Time so I think this could be implemented fairly easily as discussed here https://discourse.gohugo.io/t/readingtime-define/13036/3

- Modify the relevant layout/template/partial (?) for docs and blog to display an estimated reading time at the top of the page/post
- Have a config param that allows users to turn displaying the reading time on or off

Would be happy to have a go at a PR for this if someone could get me started in the right direction...I'm fairly new to hugo and docsy!

alexearnshaw:
This PR fixes #122 

This is what I've changed:
* Add new reading time partial
* Add conditional block for reading time in content layout
* Add new config param in userguide config.toml
* Remove reading time from specific page using front matter field (as a test) (see https://deploy-preview-209--docsydocs.netlify.com/docs/adding-content/lookandfeel/)

I've gone for a more minimal look than the one proposed in the discussion on #122, and am showing only minutes (not seconds). I'm also using the build in Hugo variable for reading time instead of calculating it manually.

There is one issue with this PR in that the reading time is not appearing on the index pages but only on all other pages.

E.g. it appears fine on https://deploy-preview-209--docsydocs.netlify.com/docs/adding-content/content/ but not on https://deploy-preview-209--docsydocs.netlify.com/docs/getting-started/ 

I've implemented this in the same way as the feedback widget (which appears on all doc pages) so can't understand why this is not showing on all pages?

### DIFF
--- a/layouts/_default/content.html
+++ b/layouts/_default/content.html
@@ -1,6 +1,9 @@
 <div class="td-content">
 	<h1>{{ .Title }}</h1>
-	{{ with .Params.description }}<div class="lead">{{ . | markdownify }}</div>{{ end }}
+    {{ with .Params.description }}<div class="lead">{{ . | markdownify }}</div>{{ end }}
+	{{ if (and (not .Params.hide_readingtime) (.Site.Params.ui.readingtime.enable)) }}
+        {{ partial "reading-time.html" . }}
+    {{ end }}       
 	{{ .Content }}
 	{{ if (and (not .Params.hide_feedback) (.Site.Params.ui.feedback.enable) (.Site.GoogleAnalytics)) }}
 		{{ partial "feedback.html" .Site.Params.ui.feedback }}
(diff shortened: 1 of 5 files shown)


LisaFC:
Thanks so much for doing this! I'll take a look when I have a few spare moments and see if I can figure out why it's not showing up on indexes.

LisaFC:
Sorry for the enormous delay, I've finally had a chance to look at this properly - the reason why the index pages don't have the update is because they use the docs/list layout rather than the one you updated. I'll see if I can fix it for you!

LisaFC:
This is great! I'm going to merge it in now so we have it, though I'll need to update the config file on the example project as well so it has the same parameter. Do we think it should be switched on or off by default I wonder?

We'll also need to update the docs...

alexearnshaw:
Thanks for getting this to work properly and merging it :) I think the property should be off by default.

frodriguezsmartclip:
@LisaFC I think, in my opinion, enabled by default because it's a nice feature to show it up...

Issue #223, opened 2020-03-04
Mobile navbar


tshort:
The existing navbar is a little limited on small screens. Depending on the site, some menu items get cut off, and the search bar is left off. Dropdown menus are also broken. 

Here is an example of a Hugo-themed site that uses a hamburger button that expands into a menu:

https://themes.gohugo.io/theme/educenter-hugo/

LisaFC:
Oh nice, I'll have a look at that, thanks for finding it.

Can I mark this as a duplicate of #211 as it's addressing the same problem?

LisaFC:
Actually just to clarify (now that I've looked at the Educenter theme) is the issue here with the top level nav menu? It *is* scrollable if menu items don't all display (I've just checked), I'm just concerned we'd end up with too many hamburger buttons given that the nav in the docs/blog section is also expandable in mobile, and it's nice for users to see *some* options without having to click anything even on a small screen. Open to suggestions though (inc if someone wants to make it configurable).

The search box vanishing *is* an issue, I think the way we have it set up is that the sidebar search comes across and replaces it on mobile, but that doesn't work if you've disabled sidebar search.

#211 is more about the nav in the docs/blog section vanishing on mobile if we remove the sidebar search, which is different (but related) so I'll keep the two separate for now.

tshort:
I understand the issue with having two hamburger buttons. That could be confusing. I see the [Apache Airflow site](https://airflow.apache.org/) does have two hamburger buttons, and it's not too bad.

The navbar menu is scrollable, but it's not obvious that it's there or that it could be scrollable (unless a menu item is obviously cut off, but then that looks clunky). Maybe it needs an arrow or other indicator that something's there. Or, maybe the navbar hamburger button could go there after the first couple of items to give a more complete view.

The other problem I'm having is that the dropdowns don't work on mobile. I can't find an online example of that because the versions online that have a version dropdown, that menu item drops from mobile (another problem--how do you switch versions on mobile?). For an example I have locally, I see the navbar menu, but when I click it, the submenu does not appear.

It's all rather tricky!

razonyang:
I have the similar issue, language dropdown menu doesn't show up in mobile screen.
When enable Algolia search, there is no way to switch to another language in small mobile screen, since sidebar search is disabled.

js-xc:
[...]

PR #232, opened 2020-03-19
Added drop down menu items in main menu

frank-segui:
Add drop down menu entries in the main menu.

I used a similar format to the one been used in the versions partial.

### DIFF
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -7,15 +7,29 @@
 		<ul class="navbar-nav mt-2 mt-lg-0">
 			{{ $p := . }}
 			{{ range .Site.Menus.main }}
+			{{ $active := or ($p.IsMenuCurrent "main" .) ($p.HasMenuCurrent "main" .) }}
+			{{ with .Page }}
+			{{ $active = or $active ( $.IsDescendant .)  }}
+			{{ end }}
+			{{ $url := urls.Parse .URL }}
+			{{ $baseurl := urls.Parse $.Site.Params.Baseurl }}
+			{{ if .HasChildren }}
+			<li class="nav-item dropdown d-none d-lg-block">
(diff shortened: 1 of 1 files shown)


tshort:
I tried something like this. A downside of this is that the dropdown menus are hidden in mobile view.

LisaFC:
Looks very nice, but that's a concern of mine with this as well, works fine on the web but currently drop-downs don't display in the mobile versions of Docsy sites so you'd have inaccessible menu items for those users (if you expect to have a lot of mobile users) unless you also have the item in eg the docs/blog left nav or a landing page. It's already an issue for people using the existing version and language drop-downs.

Here's the existing issue for mobile menu issues but I don't think anyone's got around in these virus-y times to looking at it yet, alas. #223

In general can you give me a use case for this? Is it so users can get to menu subitems from eg the landing page without having to go into the docs? Sometimes too many nav options can be confusing....

frank-segui:
We have a section of our docs that is divided in 2 parts, a visualization tool that has its own layouts and a documentation side that drills down into specifics of the visualization. The visualization does not have a side menu so it can take the whole screen.

Rather than having:
... | X-tool | X-docs | ...

We would rather have
... |    X   | ...
         |- tool 
         |- docs 

We do not have mobile users, but that likely is not the case for most people.

frank-segui:
I did some digging around and removing these lines show the dropdown menu in mobile.
https://github.com/google/docsy/blob/master/assets/scss/_nav.scss#L81
https://github.com/google/docsy/blob/master/assets/scss/_nav.scss#L91

and the d-none here:
https://github.com/frank-segui/docsy/blob/feature/add-main-dropdown-menu/layouts/partials/navbar.html#L17

but then the menu items that dont fit in the screen still appear halfway.

PR #241, opened 2020-04-06
Improve offline search result style (enlarge the result)

tekezo:
Enlarge the offline search result by append .offline-search-result class to offline search result popover and disable bootstrap default style (max-width) for the popover.

Before:
Screen Shot 2020-04-06 at 22 25 10

After:
Screen Shot 2020-04-06 at 22 24 13

### DIFF
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -129,7 +129,9 @@
 
             const $searchResultBody = $('<div>').css({
                 maxHeight: `calc(100vh - ${
-                    $targetSearchInput.offset().top + 180
+                    $targetSearchInput.offset().top -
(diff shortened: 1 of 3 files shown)


LisaFC:
Thanks, that's great! We should probably also put something in the docs to explain how to change the width:
https://www.docsy.dev/docs/adding-content/navigation/#configure-local-search-with-lunr

tekezo:
Thank you for feedback.
I added docs for configuration: https://github.com/google/docsy/pull/241/commits/92037d9ca5ee6e10c0f0589f677c9fb586033dc4

LisaFC:
Fantastic! Merging that in now.

PR #244, opened 2020-04-09
Add bootstrap, jquery and popper options

razonyang:
Because code.jquery.com is too slow for loading(1.6 - 2s), this PR provides options to change the CDN provider of the static resources, such as jsdelivr(100ms):

```toml
[params.jquery]
  src = "https://cdn.jsdelivr.net/npm/jquery@3.3.1/dist/jquery.min.js"
  hash = "sha256-FgpCb/KJQlLNfOu91ta32o/NMZxltwRo8QtmkMRdAu8="
```

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -10,6 +10,17 @@ time_format_default = "January 2, 2006"
 # Sections to publish in the main RSS feed.
 rss_sections = ["blog"]
 
+[params.bootstrap]
+src = "https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js"
+hash = "sha384-ChfqqxuZUCnJSK3+MXmPNIyE6ZbWh2IMqE241rYiqJxyMiZ6OW/JmZQ5stwEULTy"
+
+[params.popper]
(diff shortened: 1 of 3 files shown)


razonyang:
@googlebot I signed it!

LisaFC:
Nice! I'm just going to double check that you've specified the default values in the best place, because adding params to the theme config versus site configs is a bit murky, and I want to make sure that users that update to the latest version of the theme don't end up breaking their sites if they don't have the new params.

https://discourse.gohugo.io/t/config-toml-lookup-order-with-themes/24581

If you have any insights yourself, that would be great!

razonyang:
I have used this patch in my site, works fine, it seems that site `config.toml` will overwrites theme `config.toml`, but I didn't find any docs that mention about this, so I am not sure that this patch is compatible. Maybe it would be clearer to use `{{ if }}` `{{ else }}`.

coliff:
Good PR - though it'd be nice to see the versions here updated to the latest:
REF:
https://getbootstrap.com/docs/4.5/getting-started/introduction/#starter-template

razonyang:
I am going close this PR, since it probably is not the correct way to do so.

abeluck:
This would still be a great feature. In particular internal docs sites without external access would be possible!

PR #247, opened 2020-04-11
Update _index.md

razonyang:
Netlify deployment lacks a step for specifying `HUGO_ENV`, it causes some functionalities to not work properly, such as Google Analytics.


LisaFC:
Thanks for spotting that! Just tweaked the text a bit (and reminded users that they can also set the same settings in netlify.toml).

PR #249, opened 2020-04-13
Apply link hover color in link-variant mixin

fri-sch:
Currently the desired link hover color is passed to the mixin, but
not set. Fix it.


LisaFC:
Oh nice, thanks for spotting that!

PR #269, opened 2020-05-02
Utilize shallow submodules for bootstrap and Font-Awesome.

johnpaulett:
Avoids each end user needing to pass arguments to git clone / git submodule.

Reduces checkout size from 337M to 239M, and clone --recurse-submodules time
from 56s to 33s on personal laptop.

To do a similar operation without `shallow = true`, the user would have to do:
  git clone --shallow-submodules --recurse-submodules

or:
  git submodule update --init --recursive --depth=1

`shallow = true` added to git 2.10 (Q3 2016)

I believe this will assist end users of docsy, but I am not sure if it will make the maintainers' jobs more complicated when pulling updates to bootstrap / Font-Awesome.


LisaFC:
Thanks for that, I'll have a look into it - I appreciate the making cloning the theme easier, just need to check it won't cause issues with us maintaining the theme.  (we would also need to update all our documentation)

johnpaulett:
@LisaFC Thanks!  

I looked at the docsy & docsy-example documentation and I kept debating on if it warranted a change.  I *think* it will be transparent to the end users as the existing `--recurse-submodules` will still get the submodules, just shallow copies.  

I could maybe see 2 related changes:
* In docsy-example/.gitmodules make themes/docsy be shallow, so when people fork docsy-example, they get this shallow behavior for the docsy submodule itself.
* Perhaps add a note in [Getting Started](https://www.docsy.dev/docs/getting-started/#using-the-docsy-theme-as-a-submodule) that after `git submodule add` you can optionally run `git config -f .gitmodules submodule."themes/docsy".shallow true`

Happy to do these (or other changes), based upon what you think would be most appropriate.

deining:
This [issue](https://github.com/google/docsy-example/issues/83) in the repo of the example site addresses this issue, too. Would be great if this issued could be resolved eventually, in whatever way. Once a decision was made, I'm willing to author a PR in order to update the documentation.

johnpaulett:
Happy to rebase (or adjust if there is feedback) this PR.

johnpaulett:
I needed some of the recent changes, so I've rebased onto `google:master`.

chalin:
I've created an issue to track the subtasks associated with this change. See #755.

PR #275, opened 2020-05-23
Fix a offline search issue that escaped HTML entities is shown

tekezo:
This PR fixes a offline search issue that escaped HTML entities is shown as follows, and adds two configuration options.

## Fix an issue

Before:
Screen Shot 2020-05-23 at 14 14 37

After:
Screen Shot 2020-05-23 at 14 17 34

## Add configuration options

- `offlineSearchSummaryLength` which changes the summary length.
- `offlineSearchMaxResults` which changes the maximum result count.

### DIFF
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -98,7 +98,7 @@
                         });
                     });
                 })
-                .slice(0, 10);
+                .slice(0, $targetSearchInput.data('offlnie-search-max-results'));
 
             //
             // Make result html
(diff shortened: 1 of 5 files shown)


tekezo:
@medyagh I'm not planned to add pagination because lunr.js always returns entire results.
So, the pagination will reduce only the rendering cost.

medyagh:
> @medyagh I'm not planned to add pagination because lunr.js always returns entire results.
> So, the pagination will reduce only the rendering cost.

sounds good, I look forward for this PR to be merged ! I would like to use this feature

PR #277, opened 2020-05-25
adding Hungarian toml

priestoferis:
I needed to make this for Regolith's Hungarian landing page, so I thought I'd drop the pull here as well.

There is one problem though, which I left unsolved for now: the "X in Y" constructions in Hungarian are "X Y-ban" (or "X Y-ben" depending on vowel harmony), so I saw no real way to squeeze this in, so I left these two string untranslated for now.

### DIFF
--- a/i18n/hu.toml
+++ b/i18n/hu.toml
@@ -1,0 +1,47 @@
+
+
+# UI strings. Buttons and similar.
+
+[ui_pager_prev]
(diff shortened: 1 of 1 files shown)


priestoferis:
@googlebot I signed it!

LisaFC:
Thanks for this!

PR #295, opened 2020-06-22
Remove browsers from PostCSS Config

coliff:
Specifying browsers via this config is not recommended. The best way to provide browsers is a .browserslistrc file in your project root, or by adding a browserslist key to your package.json. (see: https://github.com/postcss/autoprefixer). However, if you don't define the supported browsers the default browsers option will be used (which is a great setup).
REF: https://github.com/browserslist/browserslist#full-list

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -16,17 +16,6 @@ limitations under the License.
 
 module.exports = {
     plugins: {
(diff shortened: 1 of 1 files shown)


LisaFC:
Thanks for that!

Issue #211, opened 2020-02-22
Mobile view with the sidebar search disabled


tshort:
With `sidebar_search_disable = true`, the mobile view loses the search bar and the navigation icon. The navigation icon is most important.

In Lunr and Algolia, the sidebar isn't supported, so this is probably a common setup. 

Any suggestions on how to improve the mobile view?

LisaFC:
Just FYI sidebar search is now supported with Lunr, but yes, agree we need to make sure we provide a good experience on mobile without it.

If you have one handy, do you have an example of a site without a sidebar search so I can see what it looks like in Mobile?  Otherwise I'll spin one up myself when I have a moment.  I don't think we intended for the nav icon to disappear as well but that is not to say that's not what happens. :)

tshort:
Sorry, I don’t have a public site with the sidebar disabled.

LisaFC:
No problem, I can put together a little demo and see what I can fix!

LisaFC:
Yep, you're right, the navigation isn't appearing, I'll try to fix that.

frodriguezsmartclip:
I suscribe to this thread, I need this too. Thanks @LisaFC  😉

LisaFC:
Fixed in #297

PR #297, opened 2020-06-23
Forces sidebar search to hide on desktop but show on mobile, when enabled

mgibson323:
I am using the Docsy theme on a project for work. Our goal was to have search enabled but only have 1 search box displayed at a time. This can be accomplished by hiding the sidebar search in `config.toml`, but this also hides it for mobile users. This creates a problem since the search box on the main menu at the top of the page is hidden on mobile. 

To fix this, I modified `assets/scss/_sidebar-tree.scss` and `layouts/partials/sidebar-tree.html`. The changes are trivial, but now the sidebar search box is hidden on desktop and shows on mobile - when it's enabled in `config.toml`.

There may be a better or more elegant way to do this so please feel free to implement it in whatever way makes sense for the project. 

Thanks!

### DIFF
--- a/layouts/partials/sidebar-tree.html
+++ b/layouts/partials/sidebar-tree.html
@@ -2,6 +2,12 @@
 {{ $shouldDelayActive := ge (len .Site.Pages) 2000 }}
 <div id="td-sidebar-menu" class="td-sidebar__inner{{ if $shouldDelayActive }} d-none{{ end }}">
   {{ if not .Site.Params.ui.sidebar_search_disable }}
+  <form class="td-sidebar__search d-flex align-items-center">
+    {{ partial "search-input.html" . }}
+    <button class="btn btn-link td-sidebar__toggle d-md-none p-0 ml-3 fas fa-bars" type="button" data-toggle="collapse" data-target="#td-section-nav" aria-controls="td-docs-nav" aria-expanded="false" aria-label="Toggle section navigation">
+    </button>
+  </form>
+  {{ else }}
   <div id="content-mobile">
   <form class="td-sidebar__search d-flex align-items-center">
     {{ partial "search-input.html" . }}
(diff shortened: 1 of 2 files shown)


mgibson323:
@googlebot I signed it!

LisaFC:
This looks really helpful, I'll take a proper look at how you've implemented it before I comment/merge but it seems like it should really help our mobile users.

LisaFC:
Also should fix #211

layouts/partials/sidebar-tree.html
```diff
@@ -2,11 +2,14 @@
 {{ $shouldDelayActive := ge (len .Site.Pages) 2000 }}
 <div id="td-sidebar-menu" class="td-sidebar__inner{{ if $shouldDelayActive }} d-none{{ end }}">
   {{ if not .Site.Params.ui.sidebar_search_disable }}
```
LisaFC, line 4: This *nearly* works but means that if someone has sidebar search enabled they won't get the extra sidebar search on desktop (ie the opposite of what they asked for). I appreciate this may be partly our fault for having a negative in the param name and hence a double negative ("if not disabled") in the partial...

What worked for me was leaving the "if enabled" behaviour "as is" and adding an extra {{else}} clause with your code

```
  {{ if not .Site.Params.ui.sidebar_search_disable }}
  <form class="td-sidebar__search d-flex align-items-center">
    {{ partial "search-input.html" . }}
    <button class="btn btn-link td-sidebar__toggle d-md-none p-0 ml-3 fas fa-bars" type="button" data-toggle="collapse" data-target="#td-section-nav" aria-controls="td-docs-nav" aria-expanded="false" aria-label="Toggle section navigation">
    </button>
  </form>
  {{ else }}
  <div id="content-mobile">
  <form class="td-sidebar__search d-flex align-items-center">
    {{ partial "search-input.html" . }}
    <button class="btn btn-link td-sidebar__toggle d-md-none p-0 ml-3 fas fa-bars" type="button" data-toggle="collapse" data-target="#td-section-nav" aria-controls="td-docs-nav" aria-expanded="false" aria-label="Toggle section navigation">
    </button>
  </form>
  </div>
  <div id="content-desktop"></div>
  {{ end }}
```

LisaFC:
Thanks!!

Issue #299, opened 2020-06-27
button menu switch small jump when closing


Arhell:
![3](https://user-images.githubusercontent.com/26163841/85929244-f2cae600-b8bb-11ea-89fa-85a49eea8216.jpg)
I use chrome on Windows 10 + on an android in chrome a small jump is also visible

PR #300, opened 2020-06-28
fix mobile toggle menu "jump" on close

Arhell:
issue #299


Arhell:
fixed "jumps" when closing the menu.
Tested on all modern browsers

LisaFC:
This is great, thanks so much for the fix

Issue #214, opened 2020-02-23
Docsy SCSS overriding Hugo 0.60+ highlight style


kmuncie:
The background colors set below seem to be overriding the native Hugo highlight styles after upgrading to Hugo 0.60+

- https://github.com/google/docsy/blob/master/assets/scss/_code.scss#L10
- https://github.com/google/docsy/blob/master/assets/scss/_code.scss#L41

This is especially noticeable with the darker code style themes such as the [default theme of monokai](https://gohugo.io/getting-started/configuration-markup#highlight) as seen below:

![docsy-highlight-bug](https://user-images.githubusercontent.com/582755/75117483-96626c00-5626-11ea-9405-388f55d95956.jpg)

Thank you for your time looking at this.

LisaFC:
Thanks for spotting that, I'll take a look.

frodriguezsmartclip:
@kmuncie yeah, therefore I've configured the Prism.JS to highlight code in my custom theme docsy. So, I dont have to depend if Hugo change the settings for highlighting or something became overriding in theme...

[Here I explain better](https://github.com/google/docsy-example/issues/37#issuecomment-586157096).

Regards!! ;)

gittur:
@LisaFC, any updated on this? the issue still persists.

LisaFC:
Thanks for the reminder! Will try to look at it this week.

gittur:
hey @LisaFC, Sorry to bother again, but any chance you could take a look at this soon :)

LisaFC:
Hey, sorry for the late reply (I was on vacation), I will try to get to it this week along with catching up on the day job. :)

LisaFC:
OK, I've fixed it but need to make sure it doesn't do anything really bizarre for people using pre-0.60 Hugo.

Might do a quick survey on the mailing list....

LisaFC:
One other thought is that if you don't specify a language for the block you don't get your chosen highlight formatting (it's just black text on a grey background) but I *think* that's a Hugo thing rather than anything in our CSS. You can get around it by setting "guessSyntax" to "true" and then all fenced code blocks end up in a highlight div (it hasn't given me anything particularly odd yet, but...).

Also my fix definitely gives ok results deploying with 0.59 so I think I'll get a PR together.

LisaFC:
Have a look at #312. It does mean that users who update to the latest template will also need to update their config.toml files if they don't want to use Monokai, but maybe that's ok?

gittur:
[...]

PR #303, opened 2020-07-03
Fixes ugly background text in pre/code blocks

benc-uk:
Currently there's strange looking grey background behind the text in code blocks on the kubernetes.io site, e.g.

image

This crops up in any code block (of which there are hundreds in the Kubernetes docs) e.g.
https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment
And countless other places

### DIFF
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -43,6 +43,7 @@
      
 
         > code {
+            background-color: $gray-100;
             padding: 0;
             margin: 0;
             font-size: 100%;


benc-uk:
@googlebot I signed it!

LisaFC:
This looks great but I'm just going to make sure that fixing the more general problem in #214 doesn't fix your problem as well - I think it's only happening in sites using Hugo 0.60 and above.

LisaFC:
Have a look at #312, I think that should fix it because it means you just get the Chroma highlighting CSS for your code blocks and we don't hard code any colors at all except for inline and non-language-specified code.

It does mean you need to update your config.toml as well as the template but that should be ok?

LisaFC:
@benc-uk Can you try updating to the latest Docsy and see if it fixes your problem without the need for this PR? You will probably need to update your config.toml to use Tacoma as the highlight style too. Docs for new syntax highlighting stuff is here: https://www.docsy.dev/docs/adding-content/lookandfeel/#code-highlighting

benc-uk:
I've pulled the Kubernetes site down as a clone, updated the docsy submodule to the latest and I still see the grey around the text. I changed the highlight style to **tacoma** in config.toml and the problem goes away (as the highlight blocks are black) but it's not in my remit to change this for the whole of the Kubernetes site.
I was hoping this would have been a quick fix.

LisaFC:
Hmm, that's odd, as it's fixed the same problem for another user in the original #214. I'm not sure where the grey is coming from as I've removed the hard-coded colour for pre blocks in the CSS. Let me have another look. Looking at the K8s site it seems to be only affecting certain codeblocks so I'll have a look at the source.

Do you know what version of Hugo the K8s site is using?

LisaFC:
AHA. I *think* I see where the problem is coming from. It's only happening in codeblocks created by the "codenew" shortcode and it looks like because those blocks are generated inside a table you're actually getting the style for "code inside a table" overriding the regular "code inside a pre". Let me do some further experimentation.

benc-uk:
Thanks for sticking with this!
It's a silly, minor thing really

LisaFC:
Heh, no worries, it's the kind of odd niggly problem that bothers me until I figure it out!

LisaFC:
Ok, I think I've fixed it! See if #314 fixes it for you and I'll get it in if it does.

benc-uk:
Just tried building the Kubernetes site with your `LisaFC-code-colour` docsy branch and it seems to have fixed it 🥇 
I think we can close this clumsy workaround of mine

Thanks for the fix ✨

PR #330, opened 2020-08-25
add rel attribute to prevent clickjacking

lpmi-13:
`noopener` and `noreferrer` both address clickjacking vulnerabilities on `target="_blank"`, though `noopener` isn't supported in some legacy browsers, while `noreferrer` is supported, and is ignored by more recent browsers. So adding both is currently the easiest way to support everything.


lpmi-13:
for reference, currently causing this security vulnerability in the kubernetes contributor website

clickjacking on kubernetes dev

LisaFC:
Thanks for this!

PR #334, opened 2020-09-09
Fix broken link in documentation

austintraver:
Updated link to point to `example.docsy.dev` which I presume is the new location


LisaFC:
Actually now that I look at this file I've realized it's an artefact of copying what was originally a separate repo into this project to nest the docs inside the theme - it's *extremely* out of date and should be deleted. Thanks for reminding me that it's there! I'll close this and delete the file.

PR #345, opened 2020-09-30
Add "Create child page" link

gwatts:
The "Edit this page" link is great, but a member of my team suggested yesterday that a "Create child page" button would be useful too as it could find the right spot in the repo for them to quickly add a new page, and they wouldn't then have to copy and paste markdown from another existing page to remember what the front matter was supposed to look like.

Seemed like a good idea, and was easy to add, hence a PR :-) 

There's only an `en` translation in this PR - not sure how other translations for other languages are generated.

### DIFF
--- a/assets/stubs/new-page-template.md
+++ b/assets/stubs/new-page-template.md
@@ -1,0 +1,16 @@
+---
+title: "Long Page Title"
+linkTitle: "Short Nav Title"
+weight: 100
+description: >-
+     Page description for heading and indexes.
+---
+
+## Heading
+
+Edit this template to create your new page.
+
+* Give it a good name, ending in `.md` - e.g. `getting-started.md`
+* Edit the "front matter" section at the top of the page (weight controls how its ordered amongst other pages in the same directory; lowest number first).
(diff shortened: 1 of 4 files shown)


gwatts:
And to see this in action, here's a link to the "adding content" section of the preview: https://deploy-preview-345--docsydocs.netlify.app/docs/adding-content/

Clicking the new "Create child page" link  in the right navigation section should take you to the right spot in Github to add a new file in that "adding-content" section, with a template filled out ready to edit

LisaFC:
This is a great idea, I really like it! Do we think this is something users will want to toggle on and off via config.toml, or just see by default?

gwatts:
@LisaFC I'm glad you like it :-)  My thought was that if you wanted an edit link already, you'd probably want the create option too, and in that case, creating redundant configuration seems less than ideal - Additionally easier to add a control later should someone request it, than take it away and risk breaking someone's existing config

LisaFC:
Gotcha. OK, let's go for it! If people don't like it we can always turn it into a toggle.

PR #353, opened 2020-10-04
Make bootstrap variables overwritable

hiddeco:
The documentation claims these can be overwritten[1][]. This is however
not possible at present due to the fact that `_variables.scss` is loaded
_after_ `_variables_project.scss`, while the declared variables in
the former file lack a `!default` flag. The result is that the values in
`_variables.scss` always take precedence over any project configured
value.

This commits simply adds the `!default` flag to the values, so that they
can be overwritten again with ease.

[1]: https://www.docsy.dev/docs/adding-content/lookandfeel/#color-palette-and-other-styles


LisaFC:
Thanks for spotting that!

PR #354, opened 2020-10-08
Fixed build error when pagination is used with an empty collection

tomoyanonymous:
This PR fixes build error when some directory contains no articles to be published(no articles or only drafts).
Just checking if `.Pages` exist before evaluate `.PageGroups`.
Also, lines 9 to 10 on the original code seem to be unused so removed it.

Maybe related to https://github.com/google/docsy-example/issues/56

A similar issue and solution for the reference
https://github.com/dillonzq/LoveIt/issues/347
https://github.com/dillonzq/LoveIt/pull/350

As mentioned in the issue above, the problem should be fixed by Hugo but just for a temporary solution...

### DIFF
--- a/layouts/blog/list.html
+++ b/layouts/blog/list.html
@@ -4,33 +4,35 @@
 {{ else }}
 {{$.Scratch.Set "blog-pages" .Pages }}
 {{ end }}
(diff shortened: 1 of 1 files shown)


tomoyanonymous:
@googlebot I signed it!

LisaFC:
Thanks for this!

PR #356, opened 2020-10-13
Add "print entire section" feature

gwatts:
*(this isn't "finished"; could use some feedback!)*

It's sometimes useful to be able to create a printout of all the pages
within a documentation or blog section, without having to select each
one at a time.

This commit adds a "print this section" button to the right navigation
by creating a new "print" output format and applying it to sections.

Each sub-section starts a new page.

Partials can be created to customize the layout on a per-type basis for:

* print/page-heading-<type>.html
* print/toc-li-<type>.html
* print/content-<type>.html

Pages can be omitted from a printed section entirely by setting
no_print=true in the page front matter.

Types can be excluded from being printable by setting outputs explicitly
in their front matter and listing only HTML, RSS, etc (in conjunction
with cascade)

To enable the feature, add the following to the site's config.toml

[outputs]
section = [ "HTML", "RSS", "print"]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ resources/
 node_modules/
 tech-doc-hugo
 
+# vim temporary files
+*~
+*.sw?
--- a/config.toml
+++ b/config.toml
@@ -15,5 +15,10 @@ rss_sections = ["blog"]
 # https://github.com/google/docsy-example/blob/master/config.toml
 
 
-
-
+[outputFormats]
+[outputFormats.PRINT]
+baseName = "index"
+isHTML = true
+mediaType = "text/html"
+path = "_print"
+permalinkable = false
--- a/i18n/de.toml
+++ b/i18n/de.toml
@@ -43,3 +43,13 @@ other = "Problem zu dieser Seite melden"
 other = "Problem melden"
 [post_posts_in]
 other = "Einträge in"
+
+# Print support
+[print_printable_section]
+other = "This the multi-page printable view of this section."
+[print_click_to_print]
+other = "Click here to print"
+[print_show_regular]
+other = "Return to the regular view of this page"
+[print_entire_section]
+other = "Print entire section"
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -43,3 +43,13 @@ other = "Create documentation issue"
 other = "Create project issue"
 [post_posts_in]
 other = "Posts in"
+
+# Print support
+[print_printable_section]
+other = "This the multi-page printable view of this section."
+[print_click_to_print]
+other = "Click here to print"
+[print_show_regular]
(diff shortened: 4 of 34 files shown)


gwatts:
@LisaFC not entirely clear to me why that deploy failed; i tested it with v0.56.6 of Hugo here and it runs ok.. it's complaining `failed to resolve output format "print" from site config`, however that format is referenced in the userguide `config.toml` and defined in the theme's `config.toml` so it ought to be working... :-/

LisaFC:
Huh, that's weird. I'll have a poke around on it today. I just checked and we're still using an old-ish version of Hugo (0.55) for the user guide so might try upgrading and see if that helps....

gwatts:
to update.. i tested this on my branch and hit the same issue with hugo with the 0.55 version of hugo, but ok with the latest version - at the same time i also realized it doesn't work quite right with the userguide.  I have a fix for the latter and I'll also figure out where the version threshold is that will cause it to work ok with netlify again (still unclear to me why it's not working so far)

Point being; don't spend any time investigating atm and I'll update this PR with my findings soon :-)

gwatts:
@LisaFC ok, so from testing I found that everything works fine from Hugo 0.56.0 (which was a major update) and later.  I bumped the netlify config to the latest Hugo release as I figure the userguide might as well use/test the latest release.

The "print entire section" link on doc sections [such as this](https://deploy-preview-356--docsydocs.netlify.app/docs/adding-content/) do at least seem to work as intended now.

The strings in this commit would still need to go through i18n and perhaps a bit more cleanup if you like the overall feature

gwatts:
@LisaFC any thoughts on this as a feature?  I'm finding it super useful for my docs site to quickly generate a PDF of the whole site, or a section and then attach it somewhere else, send it to my tablet, print it, etc.

Happy to spend the time cleaning it up and getting it to documented and mergeable state if it looks like it'll be useful

LisaFC:
Sorry for the delay in taking a look at this (lots to do in my actual main job :) ), I think this looks really useful. Let me take a closer look and see if there's anything we'd like to tweak before I merge.

gwatts:
No problem; on that basis, I'll do a bit more cleanup, add some docs, etc to the PR

gwatts:
@LisaFC Added a few more tweaks and updates, and added a bit of documentation at https://5fb495fb93b28a00082b89c3--docsydocs.netlify.app/docs/adding-content/print/

LisaFC (approved):
This is brilliant! Just a couple of clarifying nits in the doc and then I think we're good to go.

gwatts:
Thanks @LisaFC , appreciate the feedback - Have pushed a commit to the PR that incorporates your suggested updates.

LisaFC:
Great thanks! Merging now.

PR #359, opened 2020-10-17
Add zh translation for post_create_child_page and add post_create_child_page for all languages

ouuan:
1. Add zh translation for post_create_child_page.
2.  Add post_create_child_page for each language. Otherwise, it will be "[i18n] post_create_child_page". I think English is better than "[i18n] ...".


LisaFC:
Thanks, that's great!

Issue #360, opened 2020-10-22
local search with Lunr: offline search index file name too long


dseynaev:
```
$ hugo    
Start building sites … 
Total in 1879 ms
Error: Error building site: open /src/public/offline-search-index.3e9380f9c7bca031236be14648fb65301234ce776432d3846d2dba873f8747f0b305663a9f8fc710531db0e27e6dcadf8ebcc7be99c0ffd35af37c808729d590.json: file name too long
$ hugo version
Hugo Static Site Generator v0.75.1-A4A7BAB7/extended linux/amd64 BuildDate: 2020-09-15T06:57:20Z
$ cd themes/docsy
$ git log --oneline | head -n 1
27faf61 Merge pull request #359 from ouuan/child-page-zh-translation
```

this is running in a container under Ubuntu 20.04.1

I was following https://www.docsy.dev/docs/adding-content/navigation/#configure-local-search-with-lunr
not sure how to debug this further

LisaFC:
@tekezo, could you take a look as our resident Lunr expert? :)

tekezo:
Thank you for feedback!

It seems the offline search index file name is too long (154 chars) on some filesystem.
I make a pull request that shorten the file name. The PR fixes this issue.
https://github.com/google/docsy/pull/361

PR #361, opened 2020-10-23
Shorten offline search index file name to avoid `file name too long` error

tekezo:
Fix https://github.com/google/docsy/issues/360


LisaFC:
Thanks for this!

Issue #357, opened 2020-10-14
Toggle sidebar burger button is hidden for screens smaller than 400px on Firefox


jopasdev:
On Firefox for screens smaller than 400px the burger button is hidden and scrolling to right side is required to see the button.

![Screenshot from 2020-10-13 21-26-07](https://user-images.githubusercontent.com/368590/95929714-1185b780-0d9b-11eb-8e44-e5e9b56dade7.png)

LisaFC:
Fixed in #362

PR #362, opened 2020-10-26
Geriom burger icon

geriom:
This PR modifies de search bar on top of the left nav in two ways:

- Hide it in large screens to avoid having two search bars. 
- Set the width to 90% to address #357 

This is how it looks in different screen sizes:

docsy-iphone

docsy-tablet

docsy-large


LisaFC:
Ah this is brilliant, thanks Geri!

PR #363, opened 2020-10-26
typo in config value for github_branch

tomweston:
`config.toml` reference should read:

`github_branch = "release"`

not

`github_brach = "release"`


LisaFC:
Thanks for spotting that!

PR #365, opened 2020-10-27
Fix typo

jhwz:


LisaFC:
Thanks for spotting - closing as it was fixed in #363 (which came in just before your fix!)

PR #366, opened 2020-10-29
Update Bootstrap to v4.5.3

hawkeye116477:


hawkeye116477:
@googlebot I signed it!

dholbach:
https://github.com/google/docsy/blob/master/layouts/partials/scripts.html#L3 might need updating too?

Issue #173, opened 2020-01-14
Support: How to disable landing page section?


guiadco:
It would be great to use the docsy docs section without the landing page.

I  try to force `layout: docs` at top level, but i think `layout/_default` apply before docs layout.

- If i keep the docs section in the folder docs i have a blank page at top-level
- If i push the docs layout at top-level i have a page without docs layout style.
  - no navbar / no right menu

Could you help me?

LisaFC:
Hi! So just to clarify, do you want a site that doesn't have a landing page like docsy.dev but has the docs section at the root so you land straight on it?

You need to force not layout: but type: and then it works (I've just tried it) - have your content index page with frontmatter like:

---
title: "Goldydocs"
linkTitle: "Goldydocs"
type: docs
---

LisaFC:
Just playing around with it a bit more, you'd also probably want to remove any other folders from the example site (if you cloned it) and just have your docs sections as subsections of content. You'd then need to add type:docs to those pages too as they're not in a docs/ folder.

Alternatively you could make the docs layout your default layout - in your own project, add a layouts/ directory that overrides the theme one and add your own _default/ directory with the list.html, baseof.html and single.html from the Docsy docs/ layout directory. Then pages by default will get the left nav and right menu.

guiadco:
Hi, many thanks for your help.

I try to move my subsection from `docs/` to `.`

I add the `type: docs` on any page ans subsection pages, but i look this:

![image](https://user-images.githubusercontent.com/34829865/72454305-e39a2500-37c0-11ea-9085-e902edcd66a8.png)

and when i go to section `getting started`, i see that:

![image](https://user-images.githubusercontent.com/34829865/72454354-f90f4f00-37c0-11ea-89da-b9c114f96aff.png)

but i want to keep the left-menu as he looks like in the `docs/` folder.

Do you know, how i could do?

guiadco:
If i create a `layout/_default`, i have the same effect.

guiadco:
I see i loose this too:

![image](https://user-images.githubusercontent.com/34829865/72455610-658b4d80-37c3-11ea-83cf-cd8390364170.png)

LisaFC:
I'll take a look!

LisaFC:
Just to let you know I haven't forgotten about this, I just need to find a bit of time to see if I can get my test site to do what you want and then let you know how I did it. :) Hopefully have something by the end of the week.

shahidhk:
[...]

PR #368, opened 2020-11-03
Make it easier to create a "docs only" site

gwatts:
The "docs only" example created for issue #173 has been a super useful reference for my own internal docs site - However, I got to thinking whether Hugo's new target cascading front matter (introduced in 0.76.0) would allow us to do the same thing, with much less copying of Docsy's layout templates.

The problem with having to copy the templates, is that it becomes work to keep those templates up to date with the improvements made in the theme itself - Much better to have as little customization of those as possible.

To that end, I created a version of your example with zero custom layouts: https://github.com/gwatts/mostlydocs (example rendered: https://mostlydocs.netlify.app/ )

Getting the section menu to work correctly for that required a couple of small tweaks, which is what this PR is for.

The first commit hides the search result page from the section list on the home page.  The second ensures that the home page is correctly included in the section menu, if the home page is of type "docs".

It also adds a "toc_root" parameter - I use that on the example site to add a "blog" section that does not appear in that documentation menu, so that the blog appears to be separate.  Setting it to false (or omitting it), causes any blog section to be included in the tree like any other section.

Of course, if your site is really "just docs" you can just delete that blog section entirely.

None of this should have any effect on sites using Docsy today (if i've done this right ;-) ).

### DIFF
--- a/layouts/partials/section-index.html
+++ b/layouts/partials/section-index.html
@@ -1,5 +1,6 @@
 <div class="section-index">
     {{ $pages := (where .Site.Pages "Section" .Section).ByWeight }}
+    {{ $pages = (where $pages "Type" "!=" "search") }}
     {{ $parent := .Page }}
     {{ if $parent.Params.no_list }}
     {{/* If no_list is true we don't show a list of subpages */}}
--- a/layouts/partials/sidebar-tree.html
+++ b/layouts/partials/sidebar-tree.html
@@ -23,7 +23,8 @@
       {{ partial "navbar-lang-selector.html" . }}
     </div>
     {{ end }}
-    {{ template "section-tree-nav-section" (dict "page" . "section" .FirstSection "delayActive" $shouldDelayActive)  }}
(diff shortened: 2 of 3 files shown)


LisaFC:
Oh this is brilliant! You definitely weren't able to do this back when I created the "just docs" site (or at least not nearly so gracefully). Do you want to add docs on how to create a site that uses this feature as well - maybe on the increasingly huge "Adding content" page (which we may need to split)?

honzik20:
This sounds perfect.

gwatts:
Great!  I [wrote up some docs](https://deploy-preview-368--docsydocs.netlify.app/docs/adding-content/content/#alternative-site-structure).  I agree that page could probably use a bit of restructuring at some point; hopefully someone will have the time and inclination to take that on :-)

LisaFC:
I love it! Thanks so much for this. Merging now.

LisaFC:
@gwatts do you want to send a mail to the docsy-users list letting people know about the new feature? I'll update the example site with the latest Docsy as well.

gwatts:
thanks @LisaFC !  Post sent!

chalin:
In case someone falls on this PR, see:

- #2504

PR #372, opened 2020-11-18
fix:  L53 add period to end of sentence

aimeeu:
Currently renders without the period. Rather than changing `layouts/partials/print/render.html`, adding a period to the parameter in `eng.toml`.

### DIFF
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -50,6 +50,6 @@ other = "This the multi-page printable view of this section."
 [print_click_to_print]
 other = "Click here to print"
 [print_show_regular]
-other = "Return to the regular view of this page."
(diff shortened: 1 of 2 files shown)


aimeeu:
@googlebot I signed it!

LisaFC:
I'd add it to the partial instead, otherwise it'll just look a bit odd with the period being part of the link for this phrase but not the one on the previous line (also you'd need to add it to every language version).

Put it at the end of https://github.com/google/docsy/blob/fd2ba717a63c80538a8e18a79e2e0c1bbce69b69/layouts/partials/print/render.html#L47

LisaFC:
Great, thanks!

PR #373, opened 2020-11-19
Fix Prism.js with non-empty BaseURL.

trombonehero:
When using a BaseURL, we need to supply that URL as a prefix to the path
used to load Prism.js.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -35,6 +35,6 @@
 {{end}}
 {{ if .Site.Params.prism_syntax_highlighting }}
 <!-- stylesheet for Prism -->
-<link rel="stylesheet" href="{{ .Site.BaseURL }}/css/prism.css" />
+<link rel="stylesheet" href="{{ "/css/prism.css" | relURL }}"/>
 {{ end }}
 {{ partial "hooks/head-end.html" . }}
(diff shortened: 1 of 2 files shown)


layouts/partials/head.html
```diff
@@ -33,8 +33,8 @@
   integrity="sha384-vRQ9bDyE0Wnu+lMfm57BlYLO0/XauFuKpVsZPs7KEDwYKktWi5+Kz3MP8++DFlRY"
   crossorigin="anonymous"></script>
 {{end}}
-{{ with .Site.Params.prism_syntax_highlighting }}
+{{ if .Site.Params.prism_syntax_highlighting }}
 <!-- stylesheet for Prism -->
-<link rel="stylesheet" href="/css/prism.css" />
+<link rel="stylesheet" href="{{ .Site.BaseURL }}/css/prism.css" />
```
gwatts, line 38: A somewhat "cleaner" way of creating this url (and the js one) is to use 
```html
<link rel="stylesheet" href="{{ "/css/prism.css" | relURL }}" />
```

That will correctly prepend `BaseURL` as needed

gwatts (commented):
Good catch - I meant to open up the same PR today but forgot about it :-)

layouts/partials/head.html
trombonehero: Great point, thanks!

trombonehero:
So, I think this is ready to merge?

LisaFC:
Thanks for this!

PR #378, opened 2020-11-25
Remove markdownify for non-Markdown files

trombonehero:
When using Asciidoc (or RST, I suppose), the Inner content passed to these shortcodes should already be HTML-formatted. Making this change allows us to pass formatted content (which may include a shortcode) to a shortcode, which previously only worked with Markdown.

There might be an opportunity for a new abstraction here (maybe a partial?), but this approach seems to work OK for me. Maybe it'll be useful to others too?


LisaFC:
This looks very useful, thanks! We may tweak how we use markdownify in some of our shortcodes/layouts in future (we had to come up with a quick solution last year that worked with Hugo both before and after 0.6.0), but this is great for now.

PR #379, opened 2020-11-26
Add Persian language translations and override some styles in rtl direction

theSinner:
Hi

I added the Persian language and also override some styles in RTL direction and add [Vazir](https://github.com/rastikerdar/vazir-font) font when the language is Persian which is a good-looking open-source Persian font.

Here is the Docsy Example in the Persian language:

Screen Shot 2020-11-26 at 16 34 26
Screen Shot 2020-11-26 at 16 34 39
Screen Shot 2020-11-26 at 16 35 09

I also added Persian content in the docsy-example project.

### DIFF
--- a/assets/scss/main.scss
+++ b/assets/scss/main.scss
@@ -35,7 +35,7 @@ footer {
     min-height: 150px;
 
     @include media-breakpoint-down(md) {
-         min-height: 200px;
+        min-height: 200px;
     }
 }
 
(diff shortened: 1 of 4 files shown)


theSinner:
@googlebot I signed it!

gwatts:
wow that looks really good - Missing a few of the newer strings that were added to `en.toml` in a recent commit; might be nice to include those for completeness?

theSinner:
> wow that looks really good - Missing a few of the newer strings that were added to `en.toml` in a recent commit; might be nice to include those for completeness?

Of course. I added them in my new commit.

LisaFC:
This is great, thanks so much!

PR #383, opened 2020-12-01
Add id support for navbar menu entries

tommyblue:
I noticed [in the official docs](https://gohugo.io/content-management/menus/#add-non-content-entries-to-a-menu) the use of the `identifier` key to add specific `id` to a navbar entry. This is useful if, like in my case, specific entries need to be identified.

This PR adds this behaviour. Using `identifier = <value>` when adding a menu entry, the `a` tag will have the specified `id`


tommyblue:
@googlebot I signed it!

LisaFC:
Thanks!

PR #388, opened 2020-12-09
Added page resources to official figure shortcode

narrenfrei:
I've adopted the official shortcode ```figure.html``` to be able to use page resources and also the original src path.

The ```.Page.Resources.GetMatch (printf "**%s*" (.Get "src"))``` I've copied from the docsy shortcode ```imgproc.html```.

In my case it works like expected. Maybe this is useful for others too.


LisaFC:
Thanks for this!

chalin:
@narrenfrei - nice addition, have you considered submitting this change to the hugo repo (so that we can drop the override here)?

PR #394, opened 2020-12-13
Fix print partial for older Hugo versions

gwatts:
This was working fine for Hugo v0.75.0 and later, but not earlier - Not
sure why it was working for any version!

Tested now on 0.73.0, 0.74.0 and master

Resolves google/docsy-example#91


LisaFC:
Great, thanks!

Issue #253, opened 2020-04-17
RTL (Right-to-Left) support for special languages as in Arabic, Persian, Hebrew, etc


Hitaro:
- Is there any official support for [RTL](https://mhf-ir.github.io/i18nplease/)?

- Where should we put a rtl.css file? and how it should be pointed to in config file?

LisaFC:
The template itself doesn't but it looks like it would be fairly straightforward to add with Hugo.

If you have Hugo 0.68.0 or later you can specify in your config file whether the language direction is rtl, like so:
https://gohugo.io/content-management/multilingual/#configure-languages

...and then check on that before including your CSS. The language direction param is very new so I haven't seen any examples yet.... 

The Docsy project CSS lives as SCSS files in /assets/scss/ - depending on how different you'd need the CSS to be for RTL, you could create a separate main-rtl.scss that includes any special versions of the subfiles you need for rtl styling, and then do the check for whether the current language is RTL in /layouts/partials/head-css.html

LisaFC:
We now have RTL support for Persian thanks to @theSinner, including RTL CSS! Please reopen if there's anything else you need.

Fixed in #379

theSinner:
Hi
Yes, I set the RTL styles just for the Persian language, because I had to specify lang.
I think it’s better to add other RTL languages to support all of them.
Based on [this](https://www.andiamo.co.uk/resources/right-to-left-languages/) we have 8 RTL languages.
I will add them too in a particular PR.

LisaFC:
Fixed in #395! Great work!

PR #395, opened 2020-12-14
Add all RTL languages to the RTL styles

theSinner:
Hi
Based on this [issue](https://github.com/google/docsy/issues/253) I added other RTL languages to the RTL styles and now Docsy shows all of them in the right direction.
I also added custom fonts (from Google font) for Arabic and Hebrew languages.


LisaFC:
This is fantastic thanks!

PR #397, opened 2020-12-15
Show print section page, even if github_repo not set

### LINKED ISSUE #396: Print Support won't work if you don't have the repository links enabled
Currently, the link is only rendered if the github_repo property has a value, and therefore, only visible if also the repository links are visible as well.

It would be nice to split those so one can enable the print support without the other links (edit page, new page, new issue, etc.).

gwatts:
This diff is easier to view if you hide the whitespace changes by adding `&w=1` to the github URL :-) 

Resolves #396


LisaFC:
Thanks and thanks!

PR #407, opened 2021-01-05
Fixing typos and wrong content

deining:


userguide/content/en/docs/Adding content/content.md
```diff
@@ -104,7 +104,7 @@ title: "Adding Content"
 linkTitle: "Adding Content"
 weight: 1
 description: >
-  How to add content to your Docsy site.
```
LisaFC, line 107: I'd leave this as is, I don't think your change makes the description any clearer.

LisaFC (commented):
Thanks for the typo fixes! I think the description for the section is clear "as is" and doesn't need changing.

userguide/content/en/docs/Adding content/content.md
deining: The motivation of my change is **not** to make the description any clearer.
Have a look at the context and you will understand why I proposed the change:

```
Page frontmatter

...

For example, here’s the frontmatter for this page:

---
title: "Adding Content"
linkTitle: "Adding Content"
weight: 1
description: >
    How to add content to your Docsy site.
---
```

This is simply wrong: the frontmatter of the page `content.md` that is referred to is:

```
---
title: "Adding Content"
linkTitle: "Adding Content"
weight: 1
description: >
  Add different types of content to your Docsy site.
---
```

I guess the description was changed at some point in the past and this needs to be adjusted IMHO.
I hope that makes things clearer.

LisaFC: Aha, sorry, I only looked at the changed section and didn't clock that it was an example within the page. That's fine!

LisaFC (approved):

PR #410, opened 2021-01-11
Add workaround for Mac and mention postcss install req for v8+

RichieEscarez:
- update our outdated npm install commands (look like --save is no longer a valid flag)
- add mention of manually installing  postcss for version v8+ of postcscli
- add MacOS instructions for workaround for `Error: too many open files` and `fatal error: pipe failed`

### DIFF
--- a/userguide/content/en/docs/Getting started/_index.md
+++ b/userguide/content/en/docs/Getting started/_index.md
@@ -52,11 +52,17 @@ Install Hugo using [Brew](https://gohugo.io/getting-started/installing/#homebrew
 
 ### Install PostCSS
 
-To build or update your site's CSS resources, you also need [`PostCSS`](https://postcss.org/) to create the final assets. If you need to install it, you must have a recent version of [NodeJS](https://nodejs.org/en/) installed on your machine so you can use `npm`, the Node package manager. By default `npm` installs tools under the directory where you run `npm install`:
+To build or update your site's CSS resources, you also need [`PostCSS`](https://postcss.org/) to create the final assets. If you need to install it, you must have a recent version of [NodeJS](https://nodejs.org/en/) installed on your machine so you can use `npm`, the Node package manager. By default `npm` installs tools under the directory where you run [`npm install`](https://docs.npmjs.com/cli/v6/commands/npm-install#description):
 
 ```
-sudo npm install -D --save autoprefixer
-sudo npm install -D --save postcss-cli
+sudo npm install -D autoprefixer
+sudo npm install -D postcss-cli
+```
+
+Starting in version 8 of `postcss-cli`, you must also separately install `postcss`:
+
+```
+sudo npm install -D postcss
 ```
 
 Note that versions of `PostCSS` later than 5.0.1 will not load `autoprefixer` if installed [globally](https://flaviocopes.com/npm-packages-local-global/), you must use a local install.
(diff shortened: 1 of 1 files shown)


userguide/content/en/docs/Getting started/_index.md
```diff
@@ -255,6 +262,30 @@ gcs_engine_id = "011737558837375720776:fsdu1nryfng"
 
 To use your own Custom Search Engine, replace the value in the `gcs_engine_id` with the ID of your own search engine.
 
+## Known issues
+
+### MacOS
+
+#### `too many open files` - File Descriptors for "hugo server"
+
+By default, MacOS permits a very small number of open File Descriptors which can cause the following error:
```
LisaFC, line 271: cause the following error when building and previewing your site locally with `hugo server`, particularly with larger sites or more recent `hugo` builds:

userguide/content/en/docs/Getting started/_index.md
```diff
@@ -255,6 +262,30 @@ gcs_engine_id = "011737558837375720776:fsdu1nryfng"
...
+
+```
+ERROR 2020/04/14 12:37:16 Error: listen tcp 127.0.0.1:1313: socket: too many open files
+```
+
+You can temporarily allow more open files by running the following commands:
+
+Note that you might need to set these limits for each new shell. Also, depending on the number of files that you want to
```
LisaFC, line 279: I'd move this note to below the commands.

userguide/content/en/docs/Getting started/_index.md
```diff
@@ -255,6 +262,30 @@ gcs_engine_id = "011737558837375720776:fsdu1nryfng"
 
 To use your own Custom Search Engine, replace the value in the `gcs_engine_id` with the ID of your own search engine.
 
+## Known issues
+
+### MacOS
+
+#### `too many open files` - File Descriptors for "hugo server"
```
LisaFC, line 269: Should we maybe change this to the name of the top-level error that users see when running `hugo server` - what actually shows up is `pipe failed`?

So something like:

#### `fatal error: pipe failed` error with `hugo server`

userguide/content/en/docs/Getting started/_index.md
```diff
@@ -255,6 +262,30 @@ gcs_engine_id = "011737558837375720776:fsdu1nryfng"
...
+
+By default, MacOS permits a very small number of open File Descriptors which can cause the following error:
+
+```
+ERROR 2020/04/14 12:37:16 Error: listen tcp 127.0.0.1:1313: socket: too many open files
```
LisaFC, line 274: Again would it be more helpful to show the hugo error rather than the Mac one, as the hugo error is what users will see at the top of their stack trace? I don't know if any of the users who've had this problem have dug down far enough to get the socket error.

```
fatal error: pipe failed
```

LisaFC:
This is brilliant Richie, just a couple of notes.

userguide/content/en/docs/Getting started/_index.md
RichieEscarez: I believe this depends on your version of postcss. pre v8, you will see the `Error: too many open files` message, which is why i commented that if it worked, we would add it. [Example](https://github.com/gohugoio/hugo/issues/7773)

That said, it sounds like Knative (we are floating around 2800 files per doc version) is the only project that size who have seen the later version postcss errors.

Should we include both? Or just address the latest version?

userguide/content/en/docs/Getting started/_index.md
RichieEscarez: I wasnt sure what allow too many File Descriptors would do (im not fully familiar with what these settings really do) and wanted to make sure we said something before they ran the command with my generically large number. I slammed today but take a pass to link more Mac ref info about these settings and rewrite this section so that users can make an informed guess / estimate a reasonable value.

userguide/content/en/docs/Getting started/_index.md
RichieEscarez: This is what is surfaced when when you run `hugo server` (pre postcss v8)

userguide/content/en/docs/Getting started/_index.md
LisaFC: I think all the users I've seen have had the `fatal error: pipe failed` error but yeah, if that's the error you got and your workaround fixes both then we should do both.

So how about for the title:

`too many open files` or `fatal error: pipe failed` errors with `hugo server`

?

userguide/content/en/docs/Getting started/_index.md
LisaFC: As per above let's add both. So.

..which can cause the following errors:

```
(your error
```

or 

```
the pipe failed error
```

LisaFC (approved):

LisaFC:
Thanks heaps, Richie! Merging now.

PR #413, opened 2021-01-13
Use hugo.isProduction for production environment checks

skmcgrail:
This pull request updates the production environment checks to use the [hugo.isProduction](https://gohugo.io/functions/hugo) function. This ensures that the production environment can be specified by either `hugo --environment production` or `HUGO_ENV=production`.


LisaFC:
Oh excellent, I don't think that function existed when those checks were created!

PR #424, opened 2021-01-20
add i18n message of Chinese (zh) for print support

skyao:
Just add i18n message of Chinese (zh) for print support.


LisaFC:
Thanks!

PR #432, opened 2021-01-27
Replace jquery.js URL to a CDN location

LinuxSuRen:


LisaFC:
What's the advantage of using jsdelivr over code.jquery.com as our CDN?

LinuxSuRen:
From this doc, it's a pretty good CDN. https://github.com/jsdelivr/jsdelivr#why-jsdelivr

gwatts:
`code.jquery.com` is stackpath.. which is one of the CDNs jsdelivr uses.... i've never had an issue as far as I know so this is a bit of a solution looking for a problem imo.

It brings up a good question though around what js we actually want to host as part of the repo, vs pulling in via cdn (especially as these resources are [far less cacheable these days](https://www.stefanjudis.com/notes/say-goodbye-to-resource-caching-across-sites-and-domains/))

We have a number of libraries now that we pull in via CDN, and a number that are either in the repo, or a git submodule

Not sure i have a good answer fwiw; just interested :-)

chalin:
- `head.html` has long loaded jQuery from code.jquery.com with SRI, so there's nothing left to merge. Thanks for the proposal.

PR #463, opened 2021-02-24
Fix figure shortcode src values

eladzlot:
Prevent leading and trailing space in the image source attribute when using the figure shortcode.


eladzlot:
@googlebot I signed it!

LisaFC:
Thanks for this!

eladzlot:
My pleasure. Thanks for all the great work!

PR #459, opened 2021-02-23
Navbar: remove extra whitespace added by recent change

chalin:
PR #383 introduced extra space in the navbar entries:

>

PR #465, opened 2021-02-25
Revert #383 that used menu .Identifier to create nav item ids

### LINKED ISSUE #460: Revert #383: menu .Identifier fields meant for structuring menus not HTML ids
After looking at #383, I reread the [`menu` docs under Nesting](https://gohugo.io/content-management/menus/#nesting) where it states that (emphasis mine):

> In practice, `.Name` and `.Identifier` are only used to structure relationships and **therefore never displayed**.

In my opinion, `.Identifier` was never intended to be rendered as an HTML `id` field.

`id` fields are meant to be unique in any given page and #383 (silently) introduces `id` fields for all menus with `. Identifier` fields -- which means all menus with nested subentries. I think that #383 should be reverted.

@LisaFC - if you agree, I don't mind submitting a PR to revert #383. Thanks.

chalin:
- Reverts #383; context: https://github.com/google/docsy/issues/460#issuecomment-785225338
- Closes #460
- Closes #459, because it supersedes it

### DIFF
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -14,7 +14,7 @@
 				{{ end }}
 				{{ $url := urls.Parse .URL }}
 				{{ $baseurl := urls.Parse $.Site.Params.Baseurl }}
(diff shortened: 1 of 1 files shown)


LisaFC:
Thanks!

PR #467, opened 2021-02-26
Added support for plantuml

akojimsg:
### 1 . To configure Plantuml, update `config.toml`:

```
[params.plantuml]
enable = true
theme = "default"

#Set url to plantuml server 
#default is http://www.plantuml.com/plantuml/svg/
svg_image_url = "https://www.plantuml.com/plantuml/svg/"

```

### 2. Add diagram in markdown content (*.md) as shown in this example

````
```plantuml
participant participant as Foo
actor       actor       as Foo1
boundary    boundary    as Foo2
control     control     as Foo3
entity      entity      as Foo4
database    database    as Foo5
collections collections as Foo6
queue       queue       as Foo7
Foo -> Foo1 : To actor 
Foo -> Foo2 : To boundary
Foo -> Foo3 : To control
Foo -> Foo4 : To entity
Foo -> Foo5 : To database
Foo -> Foo6 : To collections
Foo -> Foo7: To queue
```
````

### 3. Result

image

### CODE REVIEW

**Comment by LisaFC** [comment]
This is very nice! Could you add something to the user guide too, so users will know the feature is available and where to configure it?  Maybe a section under this one:
https://www.docsy.dev/docs/adding-content/lookandfeel/#diagrams-with-mermaid

...called "UML diagrams with plantuml", or similar?

(I think we may soon need to hive off a separate page for diagrams)

**Comment by gwatts** [comment]
This is a really nice feature, but I have a few questions/reservations:

* This is using the "demo server" provided by plantuml.com.  Any diagrams source are sent to that server unencrypted, along with the resulting enerated image (doesn't seem to support `https`).   Probably fine for a public docs site, but we'd definitely want to call that out and perhaps have it disabled by default for that reason.
* `plantuml.js` includes `deflate.js` inline - It's unclear (especially with the included copyright text) whether that's actually GPL, though [this site indicates are that it is](http://www33146ue.sakura.ne.jp/staff/iz/amuse/javascript/expert/)
* In any event, I think we should separate out static third party javascript like this into `static/js` to be consistent

**Comment by akojimsg** [comment]
> This is a really nice feature, but I have a few questions/reservations:
> 
> * This is using the "demo server" provided by plantuml.com.  Any diagrams source are sent to that server unencrypted, along with the resulting enerated image (doesn't seem to support `https`).   Probably fine for a public docs site, but we'd definitely want to call that out and perhaps have it disabled by default for that reason.
> * `plantuml.js` includes `deflate.js` inline - It's unclear (especially with the included copyright text) whether that's actually GPL, though [this site indicates are that it is](http://www33146ue.sakura.ne.jp/staff/iz/amuse/javascript/expert/)
> * In any event, I think we should separate out static third party javascript like this into `static/js` to be consistent

*Not sure if I understand the first point, however url can be configured as a parameter thus it is completely customizable. It is possible to use any server running an instance of plantuml to generate an image.
*I have moved to the `deflate.js` to `static/js` and tested it. It works as expected

**Comment by LisaFC** [comment]
I'm still not sure about @gwatts first issue - does the user need to be running their own plantuml server to avoid using the demo server? Is there other guidance we should offer here?

File: userguide/content/en/docs/Adding content/lookandfeel.md [hunk 1]
```diff
@@ -191,6 +191,74 @@ See the [Mermaid documentation](https://mermaid-js.github.io/mermaid/#/Setup?id=
...
+collections collections as Foo6
+queue       queue       as Foo7
+Foo -> Foo1 : To actor 
+Foo -> Foo2 : To boundary
+Foo -> Foo3 : To control
+Foo -> Foo4 : To entity
+Foo -> Foo5 : To database
+Foo -> Foo6 : To collections
+Foo -> Foo7: To queue
+```
```
LisaFC, line 241: This isn't actually rendering in the Netlify preview, can you check what's going on?

theletterf:
Shouldn't we go for something more markdown-friendly? Project Mermaid comes to mind...

https://mermaid-js.github.io/mermaid/

LisaFC:
@theletterf we already have Mermaid support thanks to @gwatts! :)

https://www.docsy.dev/docs/adding-content/lookandfeel/#diagrams-with-mermaid

theletterf:
@LisaFC That’s fantastic! Then I’m not super sure this PR would bring added value...

LisaFC:
Are there users who would prefer to use PlantUML as an alternative? Has different features, might have existing content?

gwatts:
Mermaid is probably a good choice for most cases as it's all rendered in the browser, but PlantUML does allow for a much wider set of image types.  If you have a use case for those types and are either willing to deal with the insecure demo server, or can run your own server then I think it's pretty compelling

Having it off by default and making it clear that if you want to use it, you should stand up your own PlantUML server seems like a good option to me (assuming that license around `deflate.js` is ok)

userguide/content/en/docs/Adding content/lookandfeel.md
akojimsg: Probably related to the javascript not loading or something around that. Do you have a link to take a look?

LisaFC: It's here: https://deploy-preview-467--docsydocs.netlify.app/docs/adding-content/lookandfeel/#uml-diagrams-with-plantuml

If you look at the checks section on the PR page here in GitHub, one of the checks is "deploy preview", and if you click **Details** for that you get a preview version of the Docsy user guide with your PR applied. It's a very handy Netlify feature for checking changes.

akojimsg: Should be working now. It was not enables in the config.toml file

akojimsg:
@googlebot I fixed it.

akojimsg:
@googlebot I fixed it.

LisaFC:
@akojimsg are there other commit authors in the change that need to sign the CLA?

LisaFC:
@gwatts I changed some of the text in the user guide update in this PR, can you see if it accurately reflects what our message should be around this feature?

akojimsg:
@googlebot I fixed it.

akojimsg:
> @akojimsg are there other commit authors in the change that need to sign the CLA?

I was using the wrong email, that has been added to the google group and issue is fixed.

userguide/content/en/docs/Adding content/lookandfeel.md
```diff
@@ -191,6 +191,73 @@ See the [Mermaid documentation](https://mermaid-js.github.io/mermaid/#/Setup?id=
 
 Settings can also be overridden on a per-diagram basis by making use of the `%%init%%` header at the start of the diagram definition.  See the [Mermaid theming documentation](https://mermaid-js.github.io/mermaid/#/theming?id=themes-at-the-local-or-current-level).
 
+## UML Diagrams with Plantuml
+
+[PlantUML](https://plantuml.com/en/) is an alternative to Mermaid that lets you quickly create UML diagrams, including sequence diagrams, use case diagrams, and state diagrams. Unlike Mermaid diagrams, which are entirely rendered in the browser, PlantUML uses a PlantUML server to create diagrams. You can use the provided default demo server (not recommended for production use), or run a server yourself. Plantuml offers a wider range of image types than Mermaid, so may be a better choice for some use cases.
```
gwatts, line 196: s/Plantuml/PlantUML/ in the last sentence

gwatts (approved):
@LisaFC user guide text looks good to me!  Just the one typo

userguide/content/en/docs/Adding content/lookandfeel.md
LisaFC: Done.

LisaFC:
Thanks for this and for everyone's feedback!

Issue #457, opened 2021-02-23
Icons in the menus


narrenfrei:
I'm working on some updates improvements around our menus. #342 #449 #450 
In #342 was asked about adding icons to the menu items. So I took a look, if there already exists a official possibility in Hugo and/or Docsy.

In the examples of the Hugo docs (https://gohugo.io/content-management/menus/#add-non-content-entries-to-a-menu) they are using the parameter `pre` of the menu entries for adding icons. But in Docsy the parameter `pre` (and `post`) is not supported.

**My suggestion:**

- Add support for `pre` and `post` for menu entries threw config.toml, to support this "official" feature in the hugo docs.
- Introduce in Docsy a new meta data (front matter) term (maybe called `icon`) to define an icon at page level directly in the content files. In this parameter a icon could get defined by the common Font Awsome classes (eg. "fa fa-road").
  - This new parameter ´icon` could get used to show icons in the side bar tree menu.
  - And I would also make a possibility to set icons to the main menu also in this way (with the common Font Awsome classes). Because the HTML markup (and default formatting in CSS) is better definable, when only the classes are allowed than a injection of any code threw pre/post parameter.

This approach would also solve #444 

Any comments?

narrenfrei:
As it's not possible (or not so easy) to define a new entry for the menu properties (https://gohugo.io/variables/menus/), I think it's better to use ONLY the parameter `pre` for icons in the main menu, as already suggested in the official Hugo docs. And use my suggested front matter parameter `icon`only for the side bar tree menu (and the section-index.html).

narrenfrei:
First version is online:

https://narrenfrei.github.io/docsy-dev-page/docs/

Symbolics:
I like it. You are doing wonderful work, and clearly have some talent/skills as a web designer.

My only comment, as a user, is that both the triangle and icon may not be needed at the top level. Perhaps use the icon to expand the section at the top, and the triangle for 2nd and 3rd level expansions.

narrenfrei:
I'm also not really happy with how it looks when using icons AND the toggle buttons for folding. This is due to the fact that the theme should remain as flexible as possible: 

Maybe someone want's a foldable menu, but doesn't want to use icons. What happens, when icons are only defined for some menu items? 

[...]

Issue #449, opened 2021-02-15
Add external links to sidebar


huehnerlady:
Hi,

I would like to add some external links to the sidebar of a section on my docsy page. 
![image](https://user-images.githubusercontent.com/12533979/107918811-a4617900-6f6a-11eb-9418-67ee912a865c.png)

In this screenshot for Example the `Example Page` should be external.
I could not find any way in the documentation to do so. Is there a way? If not are there plans to introduce that?

LisaFC:
It's a tricky one because the sidebar tree partial actually iterates through your page files recursively, printing out sidebar items as it goes, so getting it to add something in in the middle would not be trivial  - unless it was at the end of the menu.  Or you could replace the generated sidebar entirely with a non-autogenerated TOC? Then you could put whatever you like in it though you lose the benefit of the menu automatically updating when you add pages. Let me have a think if there are any other options (or if anyone else has suggestions).

(it's very easy to add things to the top nav because it's a straightforward menu with all the items added explicitly, either in metadata or config: https://www.docsy.dev/docs/adding-content/navigation/#top-level-menu)

huehnerlady:
@LisaFC many thanks for your quick response 🙂 

I was thinking it to be similar to the navbar. I can add items in the menu.[lang].toml as well as the items in the tree being then ordered by `[menu.main.weight]`

So I would guess something like that would be what I am looking for and in addition it would be a similar mechanism to the top navbar, so not something completely unique :)

LisaFC:
Yeah, the sidebar's slightly different in that it iterates through all your pages (including subsections) and adds them to the sidebar even if you haven't explicitly added them to a menu - if there's a page file, it just gets pulled in and ordered by weight. Here's the code:
https://github.com/google/docsy/blob/master/layouts/partials/sidebar-tree.html#L43

You could possibly reimplement it using Hugo menus but then I think you'd have to add menu metadata (or config) for every page you wanted to include in order to "build" the menu.

huehnerlady:
That is unfortunate. I think for now we will wrap the external page we want to reference to into a page on the website then 🙂

laszewsk:
I have the same desire. If you find or have a solution let me know.

We thought even we are in the need of a second TOC on the left-hand side. E.g. one that is autogenerated, but the other one is hardcoded.

narrenfrei:
I struggled also into that problem, when I started with Docsy. My simply solution was to make a blank site holding only the link to the target. 

[...]

Issue #348, opened 2020-10-02
Left-side menu / TOC jumps around after you click a menu


prdsoftware:
Is this a bug?  When you click on a menu link on the left / TOC section, the vertical scroll bar will jump around and vertically re-position the menu structure, often pushing the link that you've clicked on off-screen.  You then need to scroll down to see the link you've click and view any sub-nodes it may have.  It's super confusing and leads to a poor user experience.

Is this a bug, or perhaps there is a configuration setting that adjusts this?

Is anyone else experiencing this?

Thanks in advance.

quickstar:
Hi
I'm experiencing the same issue. It's very annoying especially if your documentation grows and you have several nested levels. In my opinion the selected chapter should always be visible in the view port of the TOC.

LisaFC:
Hi folks, do you have an example of this behaviour I can look at? I think my own sites and the examples I've looked at have sufficiently short TOCs that it's not happening.

LisaFC:
Also what size screen/device are you looking at the site on?

prdsoftware:
You can see it on the Kubernetes site here https://kubernetes.io/docs/reference/access-authn-authz/

Click around the sub-levels of this page and you should see the topic you've clicked on be pushed off-screen.

Using a 24in monitor running at 1920x1200

LisaFC:
Aha! Finally managed to make it happen. Yes, that is weird. Will have a think about how we'd fix that - any ideas welcome.

narrenfrei:
Here is my suggestion for scroll the menu to active menu item
https://github.com/narrenfrei/docsy/commit/09fb1433f2d25d149e014a1fc582febf12bb01ed

narrenfrei:
Added some tweaks for JS scrolling active menu item to center of nav element
https://github.com/narrenfrei/docsy/blob/restructure-sidebar-tree-menu/layouts/partials/sidebar-tree.html

This version have also build in my suggestion for #450 and #449

narrenfrei:
I've made a first suggestion for #342. But I have noticed, that my JS to scroll doesnt't work within the version with a foldable sidebar menu. I have to take a further look once more.

prdsoftware:
Any update on this issue?  It looks like it's had some work done on it, but hasn't made it into a release?  Would love to see this bug fixed, as it really affects the usability.

prdsoftware:
Any news on this one?

fekete-robert:
[...]

Issue #342, opened 2020-09-28
Toggle expand/collapse for left-side menus/TOC


prdsoftware:
Firstly, love the theme - thanks for the hard work that's gone into this.

Is it possible to expand/collapse the menu structure on the left?

While I'm at it, is it possible to reduce the font-size a point or 2 for the left-menu structure.  Many titles are wrapping over 2 lines, but with a font size slightly smaller, this wouldn't happen - at least for the words I'm using - is this a CSS thing?

gwatts:
@prdsoftware to change whether the left menu is expanded/compact, in  `config.toml` you can set the `sidebar_menu_compact` parameter (see https://www.docsy.dev/docs/adding-content/navigation/#section-menu ):
```toml
 [params.ui]
# Enable to show the side bar menu in its compact state.
sidebar_menu_compact = false
```

You can also control what appears in that left-menu on a per page basis, so you can choose a shorter title that fits better there by setting `linkTitle` in the page matter at the top of each page (see also https://www.docsy.dev/docs/adding-content/navigation/ )

```yaml
---
title: "My nice long page title"
linkTitle: "Short title"
weight: 10
---
```

If you want to make the font size smaller for that nav, try this in your project's `assets/sccs/_variables_project.scss`  file (see also https://www.docsy.dev/docs/adding-content/lookandfeel/)
```css
.td-sidebar-nav {
  font-size: 0.8em !important;
}
```

(there might be a better way of overriding that css; it's not my strong suit ;-) )

prdsoftware:
Thanks @gwatts.  I was referring to the ability to manually toggle the collapse/expand state.  Have a +/- button so that the user can toggle it as they wish.  Is there a setting for this?

I'm also aware of the linktitle, but I'll fiddle around with the CSS.  I find the font size just a touch too big by default.

gwatts:
understood - afaik Docsy doesn’t have any facility for dynamically expanding that tree at the current time

hopefully the SCSS snippet is useful (tester working for me)

prdsoftware:
@gwatts Thanks for the CSS - works great!  Just what I was after.

Hopefully the dynamic TOC expansion will make it in one day.   For a documentation theme, this seems like a must-have feature.

Thanks again.

quickstar:
Hi agree
For me it's a must-have feature as well! I would highly appreciate such a feature!

prdsoftware:
[...]

PR #475, opened 2021-03-10
rebuilt sidebar menu structure

narrenfrei:
**Basic information**

- Corrected markup of nested menu lists
- best possible support of the existing CSS
- additional classes to simplify project-specific layout changes
- Cache mode of sidebar menu now caches the menu only once and not for each menu item with children separately 

**Addressed problems**

- structure of nested lists
- The same IDs ($ sid) were used several times when menu items had several sub-items with several sub-items again. The ID of HTML elements should always be unique.
- The class "active" was not consistent because the actually active `<a>` tag and also the corresponding parent `<a>` element received the class, but not grandparent elements. Now really only the active `<a>` element gets the class "active", but all `<li>` ancestors get the class "active-path".
- With the compact menu (`.Site.Params.ui.sidebar_menu_compact = true`) all ancestors and descendants were visible until now. As a result, when the lowest level was active, the whole menu was visible - not really compact ;-) All the child pages were only hidden when a sub-page was opened. At the moment I would have implemented it in such a way that all ancestors, siblings and direct descendants are always shown. In addition, the `.Site.Params.ui.ul_show` parameter can be used to set a desired menu depth to always be visible. For example, the 1st menu level can always be displayed.
  - **Is this behavior okay or should the old behavior be retained?**
- The class "collapsed" was originally used for the `<a>` elements of `<li class = "td-sidebar-nav__section-title">`. This was deleted without replacement because (1) it is not used in the CSS definitions (just "collapse") and (2) the control of the fade-in and fade-out via the `<li>` seems to work without any problems.

**Changes**

- No additional (unnecessary) `<ul>` for menu items with children.
- The classes "td-sidebar-nav__section" and "td-sidebar-nav__section-title" were merged onto an `<li>` element in order to correct the structure of the nested lists and nevertheless to support the existing CSS as best as possible.
- The sidebar menu is now cached not only for each parent element (.CurrentSection.RelPermalink), but for the entire page tree (.FirstSection.RelPermalink). So the site generation for really big sites should be a little bit faster, hopefully. To make this possible, the JS in the sidebar.html was also adapted.
  - **Is this behavior okay or should the old behavior be retained?**

**New Features**

- Introduced the class `navRoot` to make it easier to format the heading of the section, which is part of the `<ul>` structure, as "heading".
- Counting the structure depth for nested lists (ul-0, ul-1, ...)
- Introduction of `<span>` around menu links
- New (optional) parameters in the config.toml
  - `.Site.Params.ui.ul_show` (default: 1): Possibility to specify a menu depth that is always visible, even with compact menus
  - `.Site.Params.ui.sidebar_cache_limit` (default: 2000): Limit from which the sidebar menu is cached (previously hardcoded)
  - `.Site.Params.ui.sidebar_menu_truncate` (default: 50): Limit from which a submenu is shortened (previously hardcoded)
- Rewriting of the definition of "section-tree-nav-section" within sidebar-tree.html
  - One run of section-tree-nav-section per menu entry (previously Pages `.IsPage` were output directly). ==> Advantage only 1 definition for the markup per menu entry and therefore easier to understand and to maintain.
   - Since the "acitve" class is quite general, the `<span>` element of the active menu item is also marked with the "td-sidebar-nav-active-item" class.

**Further steps**
Because this PR is the basis for further steps, I would dedicate myself again on the other open issues I'm working on: #342, #348, #449, #457 and some CSS tweaks, when this PR is OK,

### DIFF
--- a/layouts/partials/sidebar-tree.html
+++ b/layouts/partials/sidebar-tree.html
@@ -1,5 +1,6 @@
 {{/* We cache this partial for bigger sites and set the active class client side. */}}
-{{ $shouldDelayActive := ge (len .Site.Pages) 2000 }}
+{{ $sidebarCacheLimit := cond (isset .Site.Params.ui "sidebar_cache_limit") .Site.Params.ui.sidebar_cache_limit 2000 }}
+{{ $shouldDelayActive := ge (len .Site.Pages) $sidebarCacheLimit }}
 <div id="td-sidebar-menu" class="td-sidebar__inner{{ if $shouldDelayActive }} d-none{{ end }}">
   {{ if not .Site.Params.ui.sidebar_search_disable }}
   <form class="td-sidebar__search d-flex align-items-center">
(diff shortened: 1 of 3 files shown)


narrenfrei:
I've now published 3 versions to test the frontend:

- [sidebar_menu_compact = false](http://docsy-dev-1.narrenfrei.at/docs/tasks/) 
- [sidebar_menu_compact = true](http://docsy-dev-2.narrenfrei.at/docs/tasks/) 
- [cached version](http://docsy-dev-3.narrenfrei.at/docs/tasks/)

LisaFC:
This is fantastic work! I'm going to go through everything in detail now to see if I think we need any tweaks/anything else, but thanks so much for doing this.

LisaFC:
Re the caching, I think we *do* want your new behaviour.  We had a couple of issues with slow generation for very big sites with deep content hierarchies (hence LOTS OF RECURSION generating the navbar)- eg #144 -  so anything that improves that is good, and it doesn't seem to create any other unwanted issues that I can see.

Any other users have any thoughts?

PR #484, opened 2021-03-21
Removed pager when printing

dinotoa:
The pager should not be presented in the printer page.
Added d-print-none style to avoid this problem.

### DIFF
--- a/layouts/partials/pager.html
+++ b/layouts/partials/pager.html
@@ -1,4 +1,4 @@
-<ul class="list-unstyled d-flex justify-content-between align-items-center mb-0 pt-5">
+<ul class="list-unstyled d-flex justify-content-between align-items-center mb-0 pt-5 d-print-none">
   <li>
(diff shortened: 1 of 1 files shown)


LisaFC:
@gwatts can you review? Thanks!

gwatts:
i don't think i've seen a case where the pager did show up.. @dinotoa am i missing a case where this is actually needed; i'm sure i must be!

dinotoa:
@gwatts the pager appears in printed blog posts, see the attached pdf.

Step to reproduce:
* open https://example.docsy.dev/blog/2018/10/06/easy-documentation-with-docsy/;
* print the page;
* look at the second page of the print preview.

[docsy-blog-post.pdf](https://github.com/google/docsy/files/6226468/docsy-blog-post.pdf)

gwatts:
ahh i see - this only occurs when doing a regular "print page", vs using "print entire section" - i was only looking at the latter - Thanks for clarifying!

This PR makes sense to me now - @LisaFC i think this is good!

PR #485, opened 2021-03-21
Icons for sidebar menu - WIP

narrenfrei:
My first (and simplest) solution for adding icons to sidebar menu items.

Here you can find a preview: http://docsy-dev-6.narrenfrei.at/docs/test-sidebar-icons/
I've only added icons to thew menu items, to show the "worst case for the layout". I also made only minimal layout threw CSS. I think, if someone wants to use icons in the sidebar menu, he would add it for all menu icons and maybe also would make his own CSS.

As an alternative I have also build a little bit more layouted version: http://docsy-dev-5.narrenfrei.at/docs/test-sidebar-icons/
In this version I have redesigned the text of the first menu item, because it's the headline for the whole section. So there was a little bit more space to align the icons and menu item text a little bit better.

Please let me know, which version you would prefer. Or if I should adapt something.


narrenfrei:
@LisaFC 
Can I ask you (or anybody else) for a short response on which proposed solution I should pursue? 

The quite minimal, which offers the possibility to add icons in the side bar navigation to the generated HTML code, but the further layout have to be done project specific by the users.
http://docsy-dev-6.narrenfrei.at/docs/test-sidebar-icons/

Or version 2, where I also integrated a little bit more CSS styling. But here also further styling will be necessary in most projects, I think.
http://docsy-dev-5.narrenfrei.at/docs/test-sidebar-icons/

LisaFC:
I think the more aligned/styled one looks a bit better? It will still look fine for users who don't add icons.

narrenfrei:
Thanks for your response. I think also that the a little bit more styled version looks better. 

But the less styled version would have the advantage, that nothing changes for users who don't want to use icons. I don't know if this is important for Docsy.

narrenfrei:
I've made a new PR #508

PR #488, opened 2021-03-24
Ajustes de Estilo no Tema

felipecotto:
estou refazendo o PR pois voltou a uma versão anterior.


chalin:
- The CLA was never signed and the change predates the monorepo restructure, so it can't be reviewed as-is. Thanks.

PR #491, opened 2021-03-25
Adding Latex support

deining:
This PR adds support for typesetting mathematical formulae using LaTeX. It uses the [KaTeX](https://katex.org/) math library for this purpose.

**Note**: Existing add-ons like Mermaid or PlantUML are enabled inside `config.toml`. Since I was on the impression, that `config.toml` is pretty extensive already, I created an extra config file `katex.toml`that does contains KaTeX options and allows enabling of the funtionality. Since hugo does not support [splitting the params](https://discourse.gohugo.io/t/config-dir-support-added-please-take-it-for-a-spin/15769/11), this file has  to reside in `data/config`. If you are happy with this solution, we should probably add files like `plantuml.toml` and `mermaid.toml`, in this directory too in order ro make enabling and configuration consistent for all add-ons. What do you think?

Have a look at the [Netlify preview](https://deploy-preview-491--docsydocs.netlify.app/docs/adding-content/lookandfeel/#latex-support-with-katex) to see LaTeX support in action.

I hope you like the feature. Awaiting your feedback.


LisaFC:
This is really great!

@gwatts, @akojimsg, anyone else, what do you think of the config model here where the config is in a separate file?

It does make a certain amount of sense for features like this that are configurable but users won't need to configure unless they're using the feature, IYKWIM, and our config.toml is already quite large even with just more "core" configurable functionality eg search, navigation, languages. It would mean that users who copy the example site wouldn't necessarily get a sample config for these features (as it doesn't use them), but we could point to the config in the Docsy user guide. Thoughts?

akojimsg:
We can create a config file for parameters (params.toml).

Generally, configuration can be placed in their own files
 - https://gohugo.io/getting-started/configuration/

LisaFC:
> We can create a config file for parameters (params.toml).
> 
> Generally, configuration can be placed in their own files
>  - https://gohugo.io/getting-started/configuration/

True, though that's a generic params file rather than params for a specific feature - one of the reasons our config.toml is so long is that we have a lot of params!

deining:
@LisaFC

> This is really great!

Glad you like it.

> It would mean that users who copy the example site wouldn't necessarily get a sample config for these features

Why not? We could (and probably should) add `data/config/katex.toml` to the example site. Here we can set `enable = false` as in the docsy theme or (in case we add formulae to the example site) we can (and have to) set it to `true`. If the user copies the example site, he has the config file inside his `data` folder. If the user uses the example site as theme, he has the config file inside the themes `data` folder. He can either edit/overwrite it directly there or copy the config file to his `data` folder and edit it then.

akojimsg:
I just took a look at #491, it is a nice feature. Having multiple plug-ins
with their respective configs is a potential problem. I think the proposed
approach is a solution.

gwatts:
While i understand the motivation to keep `config.toml` from becoming too big, i don't think i'm a fan of creating a plethora of configuration files in the `data` directory.

I find it much easier to deal with configuration in a single global `config.yaml` - especially as there's a documented process on how to override that configuration in frontmatter, a standard way to access it, etc. 

Most of the time the theme should provide reasonable defaults; i don't think `config.yaml` should have every possible option commented out as documentation; that's what the user guide is for

Besides the current configuration isn't really that large; its a couple of hundred lines and easy to skim - would be worse if it were nested XML ;-)

LisaFC:
I've thought about this for a bit and I think @gwatts is right - we should probably keep the params in the main config.toml for now as it's the most straightforward approach.

deining:
With revision 2bf057a I moved configuration data to `config.toml`, as requested. Additionally, I bumped to latest KaTeX version 0.13.1, released yesterday. I hope the PR is now ready to be merged. If there are further pending issues, please let me know.

LisaFC:
Looks good! Thanks so much for this feature.

PR #493, opened 2021-03-27
Update Korean translations in i18n ko.toml

seokho-son:
This PR updates i18n/ko.toml to complete Korean translation.

Please note that some items in i18n/ko.toml remain not translated 
since they are not suitable to be translated into Korean (Grammar). E.g., "in", "By", "Posts in", ...

By the way, I am contributing to https://kubernetes.io/ (https://github.com/kubernetes/website).
And Docsy is very useful for us ! Thank you.


LisaFC:
Thanks and thanks! :)

PR #496, opened 2021-04-01
Do not use site.BaseURL directly

yuanying:
The site.BaseURL can sometimes be "/". In that case, the URL of the
resource related to swagger-ui will be corrupted.

This patch fixes it.


LisaFC:
Thanks for this!

PR #497, opened 2021-04-01
Fix de translation for previous and next

huehnerlady:
While switching my website from German to english I realized that the buttons 'next' and 'previous' swapped all of a sudden. Looking into the toml file I figured out that they are mixed up so I thought I fix it :)


LisaFC:
Thanks for spotting this!

PR #499, opened 2021-04-02
Removed pagination widget form printed blog post list

dinotoa:
# Problem

The printed blog post list page shows the pagination widget (_internal/pagination.html) when the page is printed and the list is long enough.
The pagination widget is not shown if the link "Print entire section" is used instead.

# Way to reproduce

* Add blog posts until the pagination widget appear;
* Print the page (not the section);

# Proposed solution

Added d-print-none to the row with the pagination widget.

### DIFF
--- a/layouts/blog/list.html
+++ b/layouts/blog/list.html
@@ -28,7 +28,7 @@
 		{{ end }}
 	</div>
 </div>
-<div class="row pl-2 pt-2">
+<div class="row pl-2 pt-2 d-print-none">
 	<div class="col">
 		{{ if .Pages }}
 			{{ template "_internal/pagination.html" . }}


LisaFC:
@gwatts, want to take a look?

gwatts:
While i agree it doesn't look so good if its present, printing the pager does emphasize that you're only printing a partial list and there's more content that wasn't listed.

On that basis, i'm not sure if it would be confusing if you just removed the pager altogether for printed output.. @LisaFC have some thoughts?

dinotoa:
I agree with @gwatts that it communicates that there is more content, which can be useful or not, depending on the context. 

I prefer to have it hidden, but it's probably a matter of personal taste.

The best solution could probably be changing its appearance when printed, as at present it looks very "screen-ish". 😄

@LisaFC and @gwatts, it's up to you to decide 😉

PR #502, opened 2021-04-07
Support Google Analytics v4 (gtag.js)

tekezo:
Using `_internal/google_analytics.html` when G-MEASUREMENT_ID is specified  to support Google Analytics v4.

Hugo's documentation for Google Analytics v4 support is here.
<https://gohugo.io/templates/internal#google-analytics>

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -39,5 +39,9 @@
 {{ partial "hooks/head-end.html" . }}
 <!--To comply with GDPR, cookie consent scripts places in head-end must execute before Google Analytics is enabled -->
 {{ if eq (getenv "HUGO_ENV") "production" }}
+{{ if hasPrefix .Site.GoogleAnalytics "G-"}}
+{{ template "_internal/google_analytics.html" . }}
+{{ else }}
 {{ template "_internal/google_analytics_async.html" . }}
+{{ end }}
 {{ end }}


Symbolics:
Does this mean we won't have to manually adjust the `layouts/partials/head.html` in our local project?

tekezo:
Yes.

I was not aware of each project overriding `head.html`.
I think that such basic function should be supported by default.

RichieEscarez:
It looks like this is already supported in Hugo (and this PR can be closed): https://github.com/gohugoio/hugo/pull/8287

If this doesn't work today (specifying the UA-# in the same config - `googleAnalytics`), Docsy might need to figure out which version that was merged into and then upgrade to that version.

tekezo:
@RichieEscarez 
Unfortunately, gtag.js is not supported by `_internal/google_analytics_async.html`, which is currently used by docsy, and we need to use `_internal/google_analytics.html` instead.
(This limitation still exists with the latest version of hugo.)

Therefore, this PR is still needed. 
(Or we can change it to always use `_internal/google_analytics.html`.)

thesuperzapper:
@RichieEscarez can you please review this again, @tekezo is correct that to use GA v4, the `_internal/google_analytics.html` must be used, not `_internal/google_analytics_async.html`.

This means, without this PR, users will still have to overwrite docsy's head.html partial to use GA v4.

RichieEscarez:
Sorry about the delay and thanks for pinging me again (I let this get lost in my inbox). 

Digging more into this, it looks like using the faster "async" file is supported only for `UA-..` IDs: https://developers.google.com/analytics/devguides/collection/analyticsjs#alternative_async_tag
But is meant for newer browsers only.

The `_internal/google_analytics.html` file actually supports both `UA-..` and `G-..` IDs: https://github.com/gohugoio/hugo/blob/32ead4b1e73c3388a90dc5ef5661cdb264227d29/tpl/tplimpl/embedded/templates/google_analytics.html#L3

I cant really test out all the various scenarios myself but ive individually used each of the `UA-` and `G-` IDs and the logic looks good to me.

/lgtm

PR #506, opened 2021-04-08
WIP: New Feature: Navigation indication on the right TOC

### LINKED ISSUE #349: New Feature: Navigation indication on the right TOC
Hi

I think it would make sense to have a visual indication on which chapter on page the reader currently is.

The [Airflow](https://airflow.apache.org/docs/stable/tutorial.html#instantiate-a-dag) example has a nice implementation to showcase this feature.
Bildschirmfoto am 2020-10-02 um 14 48 02

As one can see the chapter which the reader has scrolled to, is visually highlighted on the right.

I think such a feature would greatly enhance the usability of docsy in general. What do you think?

shuuji3:
Resolves #349.

This PR implements the navigation indication on the right TOC like [Apache Airflow Documentation](https://airflow.apache.org/docs/apache-airflow/stable/index.html).

### Screen Recording

https://user-images.githubusercontent.com/1425259/114054044-75acb280-98ca-11eb-8b0a-1ba33bf455a6.mov

### Current Issue

When the depth is more than 1, the left margin is not correct.

https://user-images.githubusercontent.com/1425259/114054057-79d8d000-98ca-11eb-8b97-b93f3bae90d0.mov

Any feedback is welcome! 🙂

### DIFF
--- a/assets/js/lodash-debounce.js
+++ b/assets/js/lodash-debounce.js
@@ -1,0 +1,376 @@
+// Derived from https://unpkg.com/lodash.debounce@4.0.8/index.js
+/**
+ * lodash (Custom Build) <https://lodash.com/>
+ * Build: `lodash modularize exports="npm" -o ./`
+ * Copyright jQuery Foundation and other contributors <https://jquery.org/>
+ * Released under MIT license <https://lodash.com/license>
+ * Based on Underscore.js 1.8.3 <http://underscorejs.org/LICENSE>
+ * Copyright Jeremy Ashkenas, DocumentCloud and Investigative Reporters & Editors
+ */
+
+/** Used as the `TypeError` message for "Functions" methods. */
+var FUNC_ERROR_TEXT = 'Expected a function';
+
+/** Used as references for various `Number` constants. */
+var NAN = 0 / 0;
+
+/** `Object#toString` result references. */
+var symbolTag = '[object Symbol]';
+
+/** Used to match leading and trailing whitespace. */
+var reTrim = /^\s+|\s+$/g;
+
+/** Used to detect bad signed hexadecimal string values. */
+var reIsBadHex = /^[-+]0x[0-9a-f]+$/i;
+
+/** Used to detect binary string values. */
+var reIsBinary = /^0b[01]+$/i;
+
+/** Used to detect octal string values. */
+var reIsOctal = /^0o[0-7]+$/i;
+
+/** Built-in method references without a dependency on `root`. */
+var freeParseInt = parseInt;
+
+/** Detect free variable `global` from Node.js. */
+var freeGlobal = typeof global == 'object' && global && global.Object === Object && global;
+
+/** Detect free variable `self`. */
+var freeSelf = typeof self == 'object' && self && self.Object === Object && self;
+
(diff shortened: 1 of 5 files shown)


shuuji3:
I need to find a way to adjust the appropriate margin width for deeper TOC items.

narrenfrei:
Can you offer a live preview?

narrenfrei:
I took a look on https://airflow.apache.org/docs/apache-airflow/stable/installation.html
Maybe a tip: They define the indention of the structure level not by margin (or padding) of each `<ul>` tag of the list, but direct on the `<a>` tags in dependency of the number of `<li>` tags before:

```
.roadmap .wy-nav-side-toc .wy-menu-vertical li li > a {
    padding-left:2.427em
}

.roadmap .wy-nav-side-toc .wy-menu-vertical li li li > a {
    padding-left:4.045em
}

.roadmap .wy-nav-side-toc .wy-menu-vertical li li li li > a {
    padding-left:5.663em
}
```

shuuji3:
Thank you for the tip, @narrenfrei! Understood. Maybe we can use the same technique. I'll investigate the Docsy implementation more.

You can see the live preview here: https://deploy-preview-506--docsydocs.netlify.app/docs/getting-started/

narrenfrei:
Thanks for the preview. I think, I've already found a solution. But for this we would have to change the old CSS for specifying the old "standard alignment" of the nested lists (levels of TOC) and use the ’<ul>’ tags instead of the ’<li>’ tags to set the margin and padding. 

@LisaFC Would it be OK for you, to change the CSS a little bit?

narrenfrei:
> Thanks for the preview. I think, I've already found a solution. But for this we would have to change the old CSS for specifying the old "standard alignment" of the nested lists (levels of TOC) and use the ul tags instead of the li tags to set the margin and padding. 
> 
> @LisaFC Would it be OK for you, to change the CSS a little bit?

Sorry for my last post. But I thought a little bit more about the solution in my mind and it won't work for this issue. 

The possibilities I see now, (1) change the generated HTML code of the TOC a little bit, (2) tweak the JS or (3) a other CSS trick, I haven't in mind at the moment.

narrenfrei:
I've been thinking about our problem again. At the moment I think that it will really be best or simplest to design the CSS similar to https://airflow.apache.org/docs/apache-airflow/stable/installation.html

- No indentation for the `<li>` elements, instead the `<a>` elements directly following the `<li>` elements, depending on the number of previous `<li>` elements.
- If you mark the current position over the border of the `<li>` elements (as in your current proposal), the entire block with its sub-headings is always highlighted first and when you scroll down only the sub-heading. Alternatively, the border could also be given directly to the `<a>` element, in which case only the parent element would be highlighted.
- For each menu depth that is to be indented, the indentation distance would have to be defined separately via CSS. This would no longer indent an infinite number of levels, but only those that are defined in the CSS. This might even be an advantage, otherwise there is hardly any space left for the text after indenting too far.

e-minguez:
Bumping this as it would be really useful, thanks! :)

fekete-robert:
Hi, I have a PR that allows you to use a javascript-based ToC on the right
which has this feature: https://github.com/google/docsy/pull/1049

Issue #470, opened 2021-03-02
Upgrade to Bootstrap 5.2


LisaFC:
Bootstrap 5 has some changes that seem to break Docsy site builds #304 #464.

---

(**Edit**: @chalin)

## Info: the last BSv4 build

A link to the last edition of the User Guide that was build with BSv4: https://deploy-preview-1356--docsydocs.netlify.app. This link was taken from the last PR/commit of [v0.6.x](https://github.com/google/docsy/releases/tag/v0.6.x-last-bs4-unofficial).

## Prep

- [x] #1193
  - [x] #1481
- [x] #1454
- [x] #1458

## Tasks
 
All of the section headings below that are hyperlinks, link to a section of the [Bootstrap 5.2 migration guide][mig] of the corresponding name. A GitHub user name on a section or subsection heading declares that person as taking responsibility for making the corresponding changes to Docsy.

### ✅ [Dependencies](https://getbootstrap.com/docs/5.2/migration/#dependencies) - @chalin

- BS dropped jQuery, but it is still used if present. We'll handle this later via 
  #1436
- [x] Upgrade to Popper v2 is done automatically thanks to https://github.com/google/docsy/pull/1268
- [x] #1372
- [x] #1374

### ✅ [Sass](https://getbootstrap.com/docs/5.2/migration/#sass) - @chalin

- [x] Renamed `color-yiq()` function and related variables to `color-contrast()`, etc. For details, see corresponding [Sass](https://getbootstrap.com/docs/5.2/migration/#sass) entry:
  - [x] #517
  - [x] #1363
- [x] ⚠️ `media-breakpoint-down()` change argument to be next breakpoint of higher value
  #1367
  Projects will have to make a similar change to their styles.
- `$enable-print-styles` N/A
- `color()` etc. N/A
- `color-level` etc. N/A
- `reduced-motion` etc. N/A
- Gradient related changes: `bg-gradient-variant()` is gone, etc.
  - [x] #1369
- Removed previously deprecated mixins. N/A
- `scale-color` N/A
- No other change is applicable to Docsy.

[Customization][]:

- [x] #1462

[customization]: https://getbootstrap.com/docs/5.3/customize

### ✅ [Color system] - @chalin

While I used to think that [Color system] changes were optional, I've since changed my mind. IMHO, without #1388 we would have gotten weird behavior. For details see the issue's opening comment.

- [x] #1384
- [x] ⚠️ #1388
  This change only needs to be done by Docsy, not projects using Docsy.
[...]

PR #517, opened 2021-04-12
Update to Bootstrap v5: initial work

deining:
With official release bootstrap 5 coming soon, I just had a look at #470 to overcome the compatibility issues mentioned there.
It turned out to be quite easy, there was only one thing I had to fix as mentioned in the [migration guide](https://getbootstrap.kr/docs/5.0/migration/):

`color-yiq() function and related variables are renamed to color-contrast()`

With this PR applied, hugo runs with bootstrap 5. You may have a look at the [netlify deploy](https://deploy-preview-517--docsydocs.netlify.app/) associated with this PR.

I thought I share this already, though we should wait with the merge until bootstrap 5 is officially released. Maybe it is of help for others that want to play around with bootstrap 5 beta version already.

Preview: https://deploy-preview-517--docsydocs.netlify.app/

### DIFF
--- a/assets/scss/_boxes.scss
+++ b/assets/scss/_boxes.scss
@@ -3,7 +3,7 @@
 
 // box-variant creates the main style for a colored section used on the site.
 @mixin box-variant($parent, $color-name, $color-value) {
-    $text-color: color-yiq($color-value);
+    $text-color: color-contrast($color-value);
     $link-color: mix($blue, $text-color, lightness($color-value));
     $link-hover-color: rgba($link-color, 0.5) !default;
 
--- a/assets/scss/_colors.scss
+++ b/assets/scss/_colors.scss
@@ -1,6 +1,6 @@
 // Add some local palette classes so you can do -bg-warning -text-warning etc. Even -bg-1 if you want to paint by numbers.
 @mixin palette-variant($color-name, $color-value) {
-    $text-color: color-yiq($color-value);
+    $text-color: color-contrast($color-value);
     $link-color: mix($blue, $text-color, lightness($color-value));
    
     $link-hover-color: rgba($link-color, .5) !default;
--- a/dependencies/go.mod
+++ b/dependencies/go.mod
@@ -4,5 +4,5 @@ go 1.12
 
 require (
 	github.com/FortAwesome/Font-Awesome v0.0.0-20221115183454-96cafbd73ec4 // indirect
-	github.com/twbs/bootstrap v4.6.2+incompatible // indirect
+	github.com/twbs/bootstrap v5.2.3+incompatible // indirect
 )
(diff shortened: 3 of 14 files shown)


LisaFC:
Oh, that's good that it's a simple fix! I'll try making the change to a site running on Netlify with the latest Bootstrap and Hugo versions and see what happens.

deining:
> I'll try making the change to a site running on Netlify with the latest Bootstrap and Hugo versions and see what happens.

 Even easier: have a look at the [netlify deploy](https://deploy-preview-517--docsydocs.netlify.app/) associated with this PR. Overall it looks pretty good, but a few quirks need to be fixed:

 - The font for the text `Welcome to Docsy` on the start screen looks quite different.
 - The top heading is not perfectly aligned
 - For all documentation pages, the left margin of the main text area in the middle was reduced to zero.

LisaFC:
Ah yes, of course! And ooh yes, it seems like it's doing something very odd to some of our fonts and alignment. I'll play around with it and see what's happening in the generated HTML etc. Definitely won't move to v5 until we can come up with something that works with it *and* doesn't break formatting with v4.

deining:
> Definitely won't move to v5 until we can come up with something that works with it _and_ doesn't break formatting with v4.

I fully understand and agree with this decision.
Bootstrap has [accordion](https://getbootstrap.com/docs/5.0/components/accordion/) control implemented which look nice to me and could be useful in various places. Implementing shortcodes for accordions should be quite easy since their markup is quite similar to the tabbed panes we already have. I'm willing to contribute a PR once bootstrap 5 is out and integrated into docsy theme.

> I'll play around with it and see what's happening in the generated HTML etc.

Looking forward to your findings! Good luck!

bigpick:
Starting looking at this today after seeing the same sort of issues on one of my sites when playing around with updating everything to latest upstream;

From the breaking changes in the bootstrap migration guide, my (uneducated, after a few mins of skimming code) guess would be that the

> * Breaking: Media query mixins parameters have changed for a more logical approach.
>
> media-breakpoint-down() uses the breakpoint itself instead of the next breakpoint (e.g., media-breakpoint-down(lg) instead of media-breakpoint-down(md) targets viewports smaller than lg).
Similarly, the second parameter in media-breakpoint-between() also uses the breakpoint itself instead of the next breakpoint (e.g., media-between(sm, lg) instead of media-breakpoint-between(sm, md) targets viewports between sm and lg).

is whats at fault?

LisaFC:
@geriom can you figure out how/if you fixed this in Tekton?

chalin:
@deining et all: Shall we use this as a base to move forward with the upgrade or create a new PR entirely? A lot has changed since this was originally submitted. 

/cc @glasnt

chalin:
I've rebased and upgraded BS to v5.2.0. I'll then progressively work my way through the migration document and commit sectional changes as I go.

deining:
> I've rebased and upgraded BS to v5.2.0. I'll then progressively work my way through the migration document and commit sectional changes as I go.

Great.

I just bumped BS to latest version v5.2.3 and Font-Awesome to version 6.2.1.

chalin:
As agreed out-of-band, I'll be merging this initial work. I'll do so without squashing so that individual commits (which address separate migration tasks) remain visible.

PR #525, opened 2021-04-19
Search partial: add missing semicolon of html entity

deining:


LisaFC:
Ooh, thanks for spotting that!

PR #526, opened 2021-04-19
Shortcode 'feature': fix output of invalid nested <p> tags

deining:
This PR fixes invalid HTML output caused by by shortcode `feature`:

HTML source from docsy.dev:

```
<div class="col-lg-4 mb-5 mb-lg-0 text-center ">
  <div class="mb-4 h1">
    <i class="fas fa-lightbulb"></i>
  </div>
  <h4 class="h3">See Docsy in action!</h4>
  <p class="mb-0"><p>As well as our example site, there’s a growing number of projects using Docsy for their doc sites.</p>
</p>
  <p><a href="/docs/examples/">Read more …</a></p>
</div>
```

Note that markup of paragraph `As well as our example site, there’s ...` contains nested `<p>` tags, which is invalid.

### DIFF
--- a/layouts/shortcodes/blocks/feature.html
+++ b/layouts/shortcodes/blocks/feature.html
@@ -4,7 +4,19 @@
   <div class="mb-4 h1">
     <i class="{{ if not (or (hasPrefix $icon "fas ") (hasPrefix $icon "fab ")) }}fas {{ end }}{{ $icon }}"></i>
   </div>
-  <h4 class="h3">{{ .Get "title" | markdownify }}</h4>
-  <p class="mb-0">{{ .Inner }}</p>
+  <h4 class="h3">
+    {{ if eq .Page.File.Ext "md" }}
+      {{ .Get "title" | markdownify }}
+    {{ else }}
+      {{ .Get "title" | htmlUnescape | safeHTML }}
+    {{ end }}
+  </h4>
+  <p class="mb-0">
+    {{ if eq .Page.File.Ext "md" }}
+      {{ .Inner | markdownify }}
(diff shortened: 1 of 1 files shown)


deining:
I just realized that shortcode `lead` treats its `.Inner` content either as markdown or as HTML, depending on the extension of the page that makes use of the the shortcode:

```
{{ if eq .Page.File.Ext "md" }}
    {{ .Inner | markdownify }}
{{ else }}
    {{ .Inner | htmlUnescape | safeHTML }}
{{ end }}
```

Does it make sense to implement this for shortcode `feature` in a similar way?

LisaFC:
> I just realized that shortcode `lead` treats its `.Inner` content either as markdown or as HTML, depending on the extension of the page that makes use of the the shortcode:
> 
> ```
> {{ if eq .Page.File.Ext "md" }}
>     {{ .Inner | markdownify }}
> {{ else }}
>     {{ .Inner | htmlUnescape | safeHTML }}
> {{ end }}
> ```
> 
> Does it make sense to implement this for shortcode `feature` in a similar way?

It probably does - I think it's "fixed" in `lead` because it wasn't working on HTML pages, users might want to use `feature` in the same way.

deining:
> > Does it make sense to implement this for shortcode `feature` in a similar way?
> 
> It probably does - I think it's "fixed" in `lead` because it wasn't working on HTML pages, users might want to use `feature` in the same way.

I rebased my PR and added the code passage that differentiates between `.md` and `.html` content. Hopfully my PR is ready for merging now.

LisaFC:
Looks good!

PR #533, opened 2021-04-27
Preview of user guide: eliminate warning

deining:
This PR removes a warning that is printed out when previewing the user guide.

```
/path/to/docsy/userguide
$ hugo server --disableFastRender --themesDir=../..
Start building sites …
WARN 2021/04/27 12:15:37 found no layout file for "JSON" for kind "home": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
...
```


LisaFC:
Oh excellent, thanks for finding that.

Issue #571, opened 2021-05-26
Section menu doesn't show siblings of parent pages


AaronCrawfis:
After https://github.com/google/docsy/commit/70bf4bc580ce984c042dcfd9eeb0637cfe63cc59#diff-936974843f4ebd4ed902f880848d8640095e1506bb3d6c42bd04114a6746e14a the section menu doesn't show the sibling pages of parents anymore.

I would like to be able to show all pages at all parent levels. Is there a new parameter I'm missing?

## Before commit (correct)

All parent pages, and siblings of parents are shown.

![image](https://user-images.githubusercontent.com/54363786/119608679-31f30700-bdab-11eb-98ca-ebbfeb30bdce.png)

## After commit (incorrect)

Siblings of parents are missing. (ignore indentation, that's due to something else)

![image](https://user-images.githubusercontent.com/54363786/119608879-8ac29f80-bdab-11eb-87e9-6e659fceb67c.png)

AaronCrawfis:
@narrenfrei any ideas?

narrenfrei:
I'll take a look. Is the site public? Could you share a URL with me.

narrenfrei:
I've already found it. When I can remember me correctly, I'd gave some thought around while coding, what's the better behavior. Thought I'd decided to make it the same way, as it was before. Maybe not? I'll make a PR (hopefully) in the evening.

narrenfrei:
#575 should restore the old behavior

narrenfrei:
@AaronCrawfis
I think, this Issue can be closed, if #575 works as expected.

AaronCrawfis:
Perfect, thanks! Confirmed working

PR #518, opened 2021-04-13
Foldable sidebar (Issue #342)

narrenfrei:
As already discussed in #342 this is my suggestion for a foldable sidebar menu. It should also work with a cached sidebar menu. I've also added some general CSS tweaks for the sidebar menu.

Here you can find a preview (that also have set icons in the sidebar menu partly): https://deploy-preview-518--docsydocs.netlify.app/docs/

I've also added a short how to: https://deploy-preview-518--docsydocs.netlify.app/docs/adding-content/navigation/#section-menu

This PR is based on the PR #511 (manual extern links) and should merged after it.

PR #575, opened 2021-05-26
Restore old behavior from sidebar menu #571

narrenfrei:
With my PR #475 the behavior of the sidebar menu changed a little bit as described in the description of #475: "At the moment I would have implemented it in such a way that all ancestors, siblings and direct descendants are always shown."

In #571 the old behavior was desired: "All parent pages, and siblings of parents are shown."

I don't know what's the better behavior, but with this PR it's possible to restore the old behavior. For me both is OK, because in future I will use the foldable sidebar from #518 ;-)


LisaFC:
This is great, thanks!

PR #586, opened 2021-06-08
Allow setting the color of the root of the sidebar tree separately from the  primary color

roberthbailey:
Our site currently sets the primary color to white and also the background
of the sidebar to white. Updating our docsy theme makes it so that the root
of the sidebar tree disappears. This allows us to keep our theming and keep
the root of the sidebar a dark contrasting color from the background.

I haven't actually tested this, but was chatting with @RichieEscarez and figured that this might be the solution to my problem.

### DIFF
--- a/assets/scss/_sidebar-tree.scss
+++ b/assets/scss/_sidebar-tree.scss
@@ -99,8 +99,8 @@
 
     .td-sidebar-link.tree-root{
         font-weight: $font-weight-bold;
-        color: $primary;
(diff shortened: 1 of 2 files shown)


RichieEscarez:
This is working as expected:

/lgtm

Example: 
image

@LisaFC @geriom @emckean, I can't think of any conflict that this might cause but I'm worried I'm missing something? Before we merge can one of you please provide me with some reassurance?

PR #554, opened 2021-05-11
taxonomy support #272 #324 #539

narrenfrei:
As mentioned in #272 and #324 I continued the development of support for taxonomies / tagging.
Because I've decided to code the taxonomy functionality from scratch, I've closed #539 and made a new PR.

You can catch a current preview at https://deploy-preview-554--docsydocs.netlify.app/docs/dev-test-content/
At the moment it's completely WIP and the special pages for taxonomies are very raw.
Some commits for this PR will follow the next days.

PR #596, opened 2021-06-15
dev content removed and taxonomy manual adopted

narrenfrei:
Removed content for testing taxonomy support #554 and left a minimal showcase for the taxonomy support in the userguide.

At the moment the taxonomy clouds for tags and categories are shown in the whole docs section. If wanted, I can disable them globally. I also thinking about a way switch this clouds on and off individually per page. But for this I only have a very first concept.

I've also changed the side bar navigation tree into the new foldable variant. I hope, that's OK?

@LisaFC Is this minimal showcase of the taxonomy support (https://deploy-preview-596--docsydocs.netlify.app/docs/adding-content/taxonomy/) OK for you? Shall I remove the taxonomy clouds in the right sidebar? Shall I change the side bar navigation tree back to the old one?


LisaFC:
Oh this is great.  I think showing soe taxonomy info in the sidebar TOC is useful because it shows users that the feature is there - do we want to tag more of the content? I'll have a look at the documentation as well.

I'd change the left nav back to the fully expanded one, just because we say that it's the default and it's what we use in the example site (though I do like the foldable version!)

userguide/config.toml
```diff
@@ -80,22 +80,6 @@ description = "Docsy does docs"
 languageName ="English"
 # Weight used for sorting.
 weight = 1
-[languages.no]
```
LisaFC, line 83: Did you mean to remove our Norwegian language settings?

userguide/config.toml
```diff
@@ -186,7 +170,7 @@ no = 'Sorry to hear that. Please <a href="https://github.com/google/docsy/issues
 # If you want this feature, but occasionally need to remove the Reading time from a single page, 
 # add "hide_readingtime: true" to the page's front matter
 [params.ui.readingtime]
-enable = true
+enable = false
```
LisaFC, line 173: I can't remember whether we had this switched on by default...

userguide/config.toml
narrenfrei: Yes. Norwegian in the userguide was introduced by me for testing the multilingual taxonomy support. Only the Docsy escape site had the Norwegian content before, I think (and hope).

userguide/config.toml
narrenfrei: The reading time was switched off. I had switched it on to test the layout together with the tags.

userguide/config.toml
LisaFC: You are right - it was just the example site! (too many sites, I can't remember what features are in what site...)

PR #599, opened 2021-06-16
Corrections and additions to French i18n

boydkelly:
A few strings were untranslated.   One correction.  The expression 'Créer un problème' in French impies the user is the source of the problem.   The replacement verb 'signaler' has more the sens of 'Report an issue'.  (Maybe thats better in English too... but we won't go there!)

### DIFF
--- a/i18n/fr.toml
+++ b/i18n/fr.toml
@@ -38,18 +38,21 @@ other = "Modifier cette page"
 [post_create_child_page]
 other = "Create child page"
 [post_create_issue]
-other = "Créer un problème dans la documentation"
+other = "Signaler un problème dans la documentation"
 [post_create_project_issue]
-other = "Créer un problème dans le projet"
+other = "Signler un problème dans le projet"
 [post_posts_in]
 other = "Messages dans"
+[post_reading_time]
+other = "minutes à lire"
+
 
 # Print support
 [print_printable_section]
-other = "This the multi-page printable view of this section."
+other = "Version imprimable multipages."
 [print_click_to_print]
-other = "Click here to print"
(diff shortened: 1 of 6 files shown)


i18n/fr.toml
```diff
@@ -38,18 +38,21 @@ other = "Modifier cette page"
 [post_create_child_page]
 other = "Create child page"
 [post_create_issue]
-other = "Créer un problème dans la documentation"
+other = "Signaler un problème dans la documentation"
 [post_create_project_issue]
-other = "Créer un problème dans le projet"
+other = "Signler un problème dans le projet"
```
LisaFC, line 43: Should be "Signaler"?

userguide/config.toml
```diff
@@ -37,7 +37,7 @@ blog = "/:section/:year/:month/:day/:slug/"
 [outputs]
 home = [ "HTML" ]
 page = [ "HTML" ]
-section = [ "HTML", "RSS", "print"]
+#section = [ "HTML", "RSS", "print"]
```
LisaFC, line 40: Did you intend to comment this out?

i18n/fr.toml
boydkelly: aarrg!  typo!!!  Yes it should!

userguide/config.toml
boydkelly: No!   Sorry about that.   I thought I had squashed that from the commit.   (For some reason the print output was causing a hugo error.  Not sure what I did so just commented it for the time being.  Please ignore that.)

LisaFC:
Looks good, thanks!

PR #602, opened 2021-06-16
Add reading time to i8n and reading-time.html partial

boydkelly:
Just a small improvement to the mulilangue functionality.   Allows the 'reading time' message on blog list page to be translated.


LisaFC:
Nice, thanks!

PR #606, opened 2021-06-19
Add footer support for <card-code> shortcode

mhansen:
Copied from <card> shortcode.

I was a little surprised that card-code didn't support footers like card does, I expected them to work the same way except one contains code.


LisaFC:
@deining, do you want to take a look at this, as you implemented the card feature?

deining:
Shortcode `card-code` was deprecated meanwhile; use shortcode `card` with named parameter `code=true` instead. With `card`,  you will have footer support :smile:
This ticket is not applicable any more and can be closed.

mhansen:
Thanks! Sounds good

PR #608, opened 2021-06-20
Fix incorrect use of isset - WIP

### LINKED ISSUE #607: WARNING: calling (isset %!q(<nil>) "taxonomycloud") with unsupported collection type "invalid" (<nil>) will always return false.
After updating Graphviz's Docsy (https://gitlab.com/graphviz/graphviz.gitlab.io) to the latest Docsy version, I get this error when running:

```
```

Not very useful, so I added some more logging to track down the caller in https://github.com/gohugoio/hugo/pull/8668, which yields:

```
WARNING: calling (isset %!q(<nil>) "taxonomycloud") with unsupported collection type "invalid" (<nil>) will always return false.
WARNING: calling (isset %!q(<nil>) "taxonomypageheader") with unsupported collection type "invalid" (<nil>) will always return false.
```

The culprits:
- https://github.com/google/docsy/blob/master/layouts/partials/taxonomy_terms_clouds.html#L2 which calls `{{ if isset .Site.Params.Taxonomy "taxonomycloud" }}`
- a few other places calling `{{ if isset .Site.Params.Taxonomy "taxonomypageheader" }}`

This isn't an error in the output: if taxonomy isn't set in the `config.toml`, then no taxonomy is output, as expected. But it is a bit of warning noise for older users upgrading, or new users who deleted the taxonomy section.

How about we check if `.Site.Params.Taxonomy` is defined before looking to see if it has values set inside it?

narrenfrei:
This PR is my first try to fix #607.

I used `if isset` in a wrong way. It's a well disused topic in Hugo (eg https://github.com/gohugoio/hugo/issues/3092) and it's better to use only `if` or `with` instead of `if isset` (eg https://discourse.gohugo.io/t/solved-isset-how-to-check-variable-is-empty/1764/5)

@LisaFC I'm planning to make some more improvements to the taxonomy support in the next days. Would you like to wait with the merge for them (then a collect all updates in this PR) or should a make separate PRs?

### DIFF
--- a/layouts/_default/content.html
+++ b/layouts/_default/content.html
@@ -3,7 +3,7 @@
 	{{ with .Params.description }}<div class="lead">{{ . | markdownify }}</div>{{ end }}
 	<header class="article-meta">
 		{{ $context := . }}
-		{{ if isset .Site.Params.Taxonomy "taxonomypageheader" }}
+		{{ if .Site.Params.Taxonomy.taxonomyPageHeader }}
 			{{ range $index, $taxo := .Site.Params.Taxonomy.taxonomyPageHeader }}
 				{{ partial "taxonomy_terms_article.html" (dict "context" $context "taxo" $taxo ) }}
 			{{ end }}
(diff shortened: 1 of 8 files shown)


LisaFC:
I'm happy to merge this one now, especially as it fixes an issue.

Issue #573, opened 2021-05-26
Problems with getting Mermaid/PlantUML to work with Hugo


TueDissingWork:
Hey,

I am trying to configure either Mermaid or PlantUML for integrating digrams within the code base. However I can't seem to get it working.
Was trying to follow this guide: https://www.docsy.dev/docs/adding-content/lookandfeel/#diagrams-with-mermaid and updated the `config.toml` as per the guide.
When generating the site, no errors are presented and no diagrams aer generated either :-(

Do I need to install some thirdparty dependencies, or I am missing something?

How can I try to troubleshoot this?

/Tue

LisaFC:
Are you using a recent version of the theme, as otherwise you won't have the relevant JS?

@gwatts, any other suggestions?

TueDissingWork:
> 
> 
> Are you using a recent version of the theme, as otherwise you won't have the relevant JS?
> 
We configured the following option: `min_version = 0.53`. Which seems to be the recommended minimum version. 
Should I try with a more recent version?

joelhoisko:
The same happens to me, enabling mermaid in `config.toml` also doesn't work for me. From the browsers console I can see that `mermaid` is defined, but the markdown code blocks just display as code blocks. Adding the mermaid cdn, calling the initialize command and using `<div class="mermaid">` blocks works fine.

Hugo: `hugo v0.81.0-59D15C97+extended linux/amd64`
Docsy: commit `b2ce247` from March 19 2921

joelhoisko:
I found the problem!
TLDR: Comment out the `guessSyntax = "true"` line from `config.toml:

I noticed, that all of the code blocks with mermaid tags were being rendered as
```html
<code class="language-fallback" data-lang="fallback"></code>
```

Then looking at the [`assets/js/mermaid.js`](https://github.com/google/docsy/blob/master/assets/js/mermaid.js), it will only add the correct `class="mermaid"` to elements that have the `language-mermaid` class:
```js
var needMermaid = false;
    $('.language-mermaid').parent().replaceWith(function() {
        needMermaid = true;
        return $('<pre class="mermaid">').text($(this).text());
    });

    if (!needMermaid)  {
        mermaid.initialize({startOnLoad: false});
        return;
    }
```

What fixed this for me, was to comment out the `guessSyntax = "true"` line from `config.toml:
```toml
[markup]
  [markup.goldmark]
    [markup.goldmark.renderer]
      unsafe = true
  [markup.highlight]
[...]

PR #614, opened 2021-07-01
Fix Mermaid enabling docs

joelhoisko:
There's a bug that causes Mermaid not to work if `guessSyntax = "true" is set. https://github.com/google/docsy/issues/573#issuecomment-872068168 

A nicer fix would be to dive into Hugo's markup rendered and Chroma, but this is at least something until the underlying bug is squashed.


joelhoisko:
@googlebot I signed it!

LisaFC:
Great - that's a really useful tip!

PR #628, opened 2021-07-09
Fix duplicate hr before last modified on pages without subpages, closes #627

### LINKED ISSUE #627: Duplicate horizontal bar at bottom of leaf pages
If you add the last modified info to the bottom of your pages (_page-meta-lastmod_ partial), then the pages that do not have child pages will end up having two horizontal lines before the last modified information (at the end of the page). The cause is that the (sub)page index (_section-index_ partial) adds a horizontal line even if the pages collection is empty.

thebe14:


thebe14:
@googlebot I signed it!

LisaFC:
Thanks!

PR #635, opened 2021-07-16
Change navbar search border colour to make it stand out

at055612:
Currently the navbar search input has border `none` so you just see a hint of the drop shadow.  As the input text is shorter than the input it looks like there is a lot of empty space to the right of it.  

Before the change it looks like 

image

After the change it looks like

image

### DIFF
--- a/assets/scss/_nav.scss
+++ b/assets/scss/_nav.scss
@@ -59,7 +59,7 @@
     }
 
     .td-search-input {
-        border: none;
+        border-color: $navbar-dark-color;
 
         @include placeholder {
             color: $navbar-dark-color;


LisaFC:
I like it, though I wonder should we make it an option rather than the default? Anyone else?

narrenfrei:
I've also changed the lookout from this search input in our docs a little bit. You can have a look at this test page:
https://60f2abac014e3400078b4632--sad-volhard-b41b3b.netlify.app

This example has 3 changes to the current official version:

1. make the search input a little bit more visable
2. let the placeholder visible until typing in something (like in the sidebar)
3. give the language dropdown in the sidebar (on smaller devices) a little bit more padding

The version of @at055612 is also OK for me and in my opinion better the the current search input. I would make it as Docsy default if you merge it.

LisaFC:
I like them both but I think I like @narrenfrei's version slightly better. Anyone else?

at055612:
I'm happy with @narrenfrei's version.  A bit more subtle.  As long as the border can be seen I am ok.

Jassi10000-zz (changes requested):
@at055612 
Buddy you must resolve the merge conflicts before you make a PR

at055612:
> @at055612 Buddy you must resolve the merge conflicts before you make a PR

The PR was not in conflict when created.  It is now but I was waiting on a decision from the maintainers whether to proceed with this PR or for @narrenfrei's change to be used instead.

chalin:
My vote is to preserve the current default. Projects can easily customize this to their own tastes.

prdsoftware:
The new version of Algolia has different styling for the search box that makes it stand out by default.

chalin:
- Made moot by the later search-box restyling (noted in-thread, 2022). Thanks.

PR #540, opened 2021-05-02
Accessibility and SEO improvements

jcabak:
- Background and foreground colours have a sufficient contrast ratio. (In “Docs” for date label “Last modified”)
- Links have a discernible name
- Added `aria-label` for links “read more” and “next” and previous” article in blog
- Links to cross-origin destinations are safe (added `rel="noopener"`)
- Document have a meta description
- Text remains visible during webfont load (added `font-display: swap;`)

Issue #611, opened 2021-06-26
Check hugo.Environment == “production” rather than HUGO_ENV=production environment variable


mhansen:
Hi, this is forked off from https://github.com/google/docsy/issues/217 to track a systemic fix for how Docsy by-default sets all your pages to `noindex`. I think this is a fairly high-severity issue: sites migrating to Docsy might find themselves delisted from Google until Google reindexes.

https://discourse.gohugo.io/t/checking-env-variable-prevents-site-from-being-indexed-by-google-on-aws/23707
 suggests that instead of

```
{{ if eq (getenv “HUGO_ENV”) “production” }} 
```

we should use

```
{{ if eq hugo.Environment “production” }}
```

Then running with

`hugo` will build for production config.

`hugo.Environment` is defined at https://gohugo.io/variables/hugo/

I wonder how much backwards-compatibility concerns there are here for people updating though.

LisaFC:
Default no-index is also a Hugo default (I think), and I know a lot of users don't want their site indexed while they're actively working on it (eg testing a live deployment method) or sharing it as a preview.

We do tell users to switch to "production" in the hosting and deployment page but could possibly make it more prominent to ensure they see it?

mhansen:
I can't find much by googling if hugo has noindex as the default. I *think*
hugo doesn't ship with any default HTML templates (except maybe the
sitemap.xml?) so it's up to the users to specify whether they want it to be
indexed or not.

I suppose this might be a tricky tradeoff if there are other users who want
their site not indexed. On the other side, it looks like the current
approach led to Graphviz and gRPC's pages being delisted from Google for a
while (https://github.com/grpc/grpc.io/issues/600).

The docs about setting HUGO_ENV were under the "Deployment with Netlify"
section and I wasn't deploying with Netlify (GitLab pages instead) so I
didn't read them :-)

https://www.docsy.dev/docs/deployment/#deployment-with-netlify

LisaFC:
How about we make the information more prominent on the deployment page, as I don't want to fix one set of problems ("oh no, my site is not indexed!") by creating new problems for other people ("oh no, my site *is* indexed!"). :)

I'll do a PR.

LisaFC:
[...]

PR #653, opened 2021-08-09
Use hugo.IsProduction instead of checking HUGO_ENV

### LINKED ISSUE #652: Changes introducing hugo.IsProduction were overwritten
_Originally posted by @chalin in https://github.com/google/docsy/issues/611#issuecomment-895335837_

---

@LisaFC et al.: somehow #540 reverted #413. The latter introduced the use of `hugo.IsProduction`, in replacement of the environment-variable checking code. IMHO, use of `hugo.IsProduction` is the better (if not proper) way to determine the build environment, in particular since the build environment can be specified using the [-e flag][] command line flag.

[-e flag]: https://gohugo.io/commands/hugo/#options

Possibly related: #611
Related: #651

/cc @nate-double-u

chalin:
- Reapplies #413, which was removed by mistake in #540. For details, see https://github.com/google/docsy/pull/540/files#r685339766
- Closes #652

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -7,7 +7,7 @@
 <link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink | safeURL }}">
 {{ end -}}
 
-{{ if and (eq (getenv "HUGO_ENV") "production") (ne $outputFormat "print") -}}
+{{ if and hugo.IsProduction (ne $outputFormat "print") -}}
 <META NAME="ROBOTS" CONTENT="INDEX, FOLLOW">
 {{ else -}}
 <META NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">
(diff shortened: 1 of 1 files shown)


chalin:
@LisaFC - given that this essentially fixes a regression, can it be merged soon? /cc @nate-double-u

LisaFC:
Yep, I'm just looking at it now - the main thing is that it needs to be documented and announced because it's a breaking change with regard to whether your site is indexed.

(which I didn't think the original change was because I hadn't fully explored the mysterious ways of how the relevant environment variable and the current Hugo environment are not the same thing, as I did in #611)

chalin:
Thanks. AFAICT, `hugo.IsProduction` will be true if `HUGO_ENV=production` is set, so it might not be a breaking change.

LisaFC:
I *think* that's true but I need to check, I remember finding some unintuitive behaviour when I explored it in #611 

We also need to document for the case where `HUGO_ENV=production` is *not* set and the user expects their site not to be indexed - if we make this change it *will* be indexed unless they explicitly specify that this is a non-production build.

LisaFC:
> Thanks. AFAICT, `hugo.IsProduction` will be true if `HUGO_ENV=production` is set, so it might not be a breaking change.

Annoyingly, I just checked, and it isn't (or at least `HUGO_ENV`=production in your Netlify setup doesn't override setting an environment in your hugo command with `-e`) - though it shouldn't be a problem anyway for existing sites as users will get a production build by default.

chalin:
Right, `production` is the default for the `build` command. It makes sense that `-e` takes precedence. 

Btw, let me know if there is anything I can do to help.

PR #655, opened 2021-08-10
Adding Selenium to the list of sites using Docsy

diemol:
👋 
We are now using Docsy for our Selenium website, we thought it was a good idea to add it to the list of sites using Docsy.

Thanks for this theme!


diemol:
@googlebot I signed it!

PR #662, opened 2021-08-18
FIX: meta description tag creation in head.html (SEO)

aimeeu:
Change meta `description` tag creation so the content is the actual page description or the first 150 characters of the page content rather than a value defined in `config.toml`.

**Problem**:  
Introduced in PR #540 
`<meta name="description" content="{{ .Site.Params.Description }}">` in `head.html` adds the meta description tag with the _exact same content_ to each page. According to Google's [Search Engine Optimization (SEO) Starter Guide](https://developers.google.com/search/docs/beginner/seo-starter-guide), "A page's description meta tag gives Google and other search engines a summary of what the page is about. A page's title may be a few words or a phrase, whereas a page's description meta tag might be a sentence or two or even a short paragraph." So including the exact same meta description tag on each page causes search engines and third-party search providers (like Elastic App Search) to think every page on your site has the same description. In this case, each page's description is whatever you set in `config.toml`. 

For example, if you have configured:

```yaml
[languages.en]
title = "Armory Docs"
description = "Unlocking Innovation by Making Software Delivery Continuous, Scalable, and Safe "
```

Then every rendered page in your site will have the `<meta name="description" content="Unlocking Innovation by Making Software Delivery Continuous, Scalable, and Safe">` tag. 

I discovered this issue after upgrading Docsy and seeing irregularities in my SEO and Elastic indexing... and consequently poor search results. Last year I had created my meta `description` tag in my copy of `head-end.html`. After recently upgrading Docsy, I noticed that two meta descriptions tag were created for each page: the first tag, created in `head.html`, used the `.Site.Params.Description` and the second (from my custom `head-end.html`) used `.Page.Description`. The indexing engine picked up the first (incorrect) meta tag created in `head.html`.  

**Fix**: 
change the tag creation in `head.html` to use the page's frontmatter `description`. If there isn't a description, use the first 150 characters of the page's content. Add content to the documentation explaining how Docsy creates the meta `description` tag.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -15,7 +15,14 @@
 
 {{ partialCached "favicons.html" . }}
 <title>{{ if .IsHome }}{{ .Site.Title }}{{ else }}{{ with .Title }}{{ . }} | {{ end }}{{ .Site.Title }}{{ end }}</title>
-<meta name="description" content="{{ .Site.Params.Description }}">
+<!-- for search engine optimization and third-party search engines like Elastic App Search-->
+{{ if .Page.Description }}
+  <meta name="description" content="{{ .Page.Description }}">
+{{ else }}
+  {{ $desc := (.Page.Content | safeHTML | truncate 150) }}
+  <meta name="description" content="{{ $desc }}">
+{{ end }}
+
 {{- template "_internal/opengraph.html" . -}}
 {{- template "_internal/google_news.html" . -}}
 {{- template "_internal/schema.html" . -}}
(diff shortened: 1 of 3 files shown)


aimeeu:
Deploy preview:  https://deploy-preview-662--docsydocs.netlify.app/docs/adding-content/feedback/#search-engine-optimization-meta-tags

If you do a view source on the page, you'll see the meta description tag now contains the page's frontmatter description instead of "Docsy does docs" like in the current docsy.dev production docs site.

LisaFC:
Oh this is *great*!!! This will be really helpful for users.

PR #669, opened 2021-08-24
Fixing infinite loop when several tabs are present

diemol:
Based on how Bootstrap suggests to select tabs https://getbootstrap.com/docs/4.6/components/navs/#tabshow

We have an issue on the Selenium docs, where this js script is blocking the navigations. I made the change locally and based on the recommendation from Bootstrap, this is a much better way to select the tab, and to avoid the infinite loop.

### DIFF
--- a/static/js/tabpane-persist.js
+++ b/static/js/tabpane-persist.js
@@ -4,15 +4,16 @@ if (typeof Storage !== 'undefined') {
         document
             .querySelectorAll('.tab-' + activeLanguage)
             .forEach((element) => {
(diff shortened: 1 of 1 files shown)


diemol:
For more context:
If you navigate to https://www.selenium.dev/documentation/getting_started/installing_browser_drivers/, and then select one of the tabs, the js script is triggered and the navigation halts as the script enters into an infinite loop which makes our docs unusable. 

Hopefully we can merge this soon, thanks!

PR #674, opened 2021-08-29
Improve Reading time format

haghighi-ahmad:
Using Zero (`0`) for reading time is not beautiful, and it's better to have a specific phrase for 0 length posts.   
This PR Improves the Reading time format, uses `less than a minute` for short posts (reading time <= 1), and uses local language to represent reading time. 

image


haghighi-ahmad:
@googlebot I signed it!

LisaFC:
Nice! This is a really helpful improvement

LisaFC:
I'm just going to switch off reading time in the PR (now that I've seen the updated feature in action!) as it's not normally switched on in the user guide, but otherwise it's good to go.

PR #691, opened 2021-09-16
Offline search: fingerprint data in production only

chalin:
Extending #651 to the offline search data file.

/cc @nate-double-u


nate-double-u (approved):

nate-double-u:
This is similar to the change made over at https://github.com/etcd-io/website/pull/484
(I just realized I'd approved w/o context, sorry about that)

PR #700, opened 2021-09-21
.gitignore cleanup, and tweak for userguide/public

chalin:


userguide/.gitignore
```diff
@@ -1,6 +1,2 @@
-
-public/
+/public
```
chalin, line 1: This allows `public` to be a symlink.

PR #702, opened 2021-09-22
Support disabling of navbar translucency over cover blocks

chalin:
Docsy has special formatting and code in support of translucent navbars when over a `block/cover`. This sometimes causes problems, e.g., see https://github.com/grpc/grpc.io/issues/866.

This PR introduces the `navbar_translucent_over_cover_disable` configuration option, allowing site devs to disable this special behavior.

### DIFF
--- a/layouts/shortcodes/blocks/cover.html
+++ b/layouts/shortcodes/blocks/cover.html
@@ -24,7 +24,7 @@
 }
 </style>
 {{ end }}
-<section id="{{ $blockID }}" class="row td-cover-block td-cover-block--height-{{ $height }} js-td-cover td-overlay td-overlay--dark -bg-{{ $col_id }}">
+<section id="{{ $blockID }}" class="row td-cover-block td-cover-block--height-{{ $height }}{{ if not .Site.Params.ui.navbar_translucent_over_cover_disable }} js-td-cover{{ end }} td-overlay td-overlay--dark -bg-{{ $col_id }}">
   <div class="container td-overlay__inner">
     <div class="row">
       <div class="col-12">
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -1,4 +1,4 @@
-{{ $cover := .HasShortcode "blocks/cover" }}
+{{ $cover := and (.HasShortcode "blocks/cover") (not .Site.Params.ui.navbar_translucent_over_cover_disable) }}
 <nav class="js-navbar-scroll navbar navbar-expand navbar-dark {{ if $cover}} td-navbar-cover {{ end }}flex-column flex-md-row td-navbar">
         <a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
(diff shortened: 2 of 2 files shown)


layouts/shortcodes/blocks/cover.html
```diff
@@ -24,7 +24,7 @@
 }
 </style>
 {{ end }}
-<section id="{{ $blockID }}" class="row td-cover-block td-cover-block--height-{{ $height }} js-td-cover td-overlay td-overlay--dark -bg-{{ $col_id }}">
+<section id="{{ $blockID }}" class="row td-cover-block td-cover-block--height-{{ $height }}{{ if not .Site.Params.ui.navbar_translucent_over_cover_disable }} js-td-cover{{ end }} td-overlay td-overlay--dark -bg-{{ $col_id }}">
```
chalin, line 27: While not strictly necessary, this change prevents the navbar-style-fiddling JS code from being triggered when translucency is disabled: 
https://github.com/google/docsy/blob/db8f7a7528ee0387728c07312118c3f749aab3d8/assets/js/base.js#L35-L40

LisaFC:
Ah this is great! I can see the problem on the grpc site. Is there a good place to document this parameter so other users can find it?

(I worked on the earliest iteration of the grpc website, back before Docsy existed. :) )

chalin:
> Is there a good place to document this parameter so other users can find it?

There are a few options, but I was hoping that we could get this merged first. I promise to: (1) add the config option to the example site config file, (2) submit a followup PR with some added documentation -- possible under the https://www.docsy.dev/docs/adding-content/lookandfeel/ section.

WDYT?

> I worked on the earliest iteration of the grpc website, back before Docsy existed. :)

Oh cool! Right, come to think of it, I do remember seeing your name in some commits from way back! 😄

LisaFC:
Works for me! I'll merge this now.

chalin:
> (1) add the config option to the example site config file,

Done, but for some reason the build(s) are failing. I tried to "fix" them, but I believe that the cache might need to be reset. Here's the main PR I submitted: https://github.com/google/docsy-example/pull/130

LisaFC:
Yeah, Netlify preview builds sometimes go weird like that, I'll manually clear the cache and rerun it.

LisaFC:
So even more weirdly (and I may have to seek Netlify help on this), there are TWO sets of preview deploys coming from the docsy-example repo, one of which always fails. I have no idea where the other one came from, unless it's an artefact of a site name/Netlify settings change.

Behold (if you can see these):
Success! https://app.netlify.com/sites/goldydocs/deploys/614c7fd8cd0a68ebf63be5b1
Fail! https://app.netlify.com/sites/docsy-example/deploys/614b64ce1c53810009bd07b0

Your PRs are absolutely fine and the previews are building with no issues.

chalin:
It seems like you have your own "LisaFC's team/goldydocs" Netlify account in addition to the "Docsy/docsy-example" account. I suspect that the config settings of the latter are out-of-date. That's why I usually like to put as many of the Netlify config settings as possible in the project's `netlify.toml` file.

LisaFC:
Yeah, the Docsy/docsy-example account is from ... somewhen. I'll see if I can track it down and/or dig into with Netlify.

PR #719, opened 2021-10-13
Fix the search field background color when offline search is enabled.

### LINKED ISSUE #718: lunr search: text field background fills in after you press enter, causing poor contrast
To repro, set `offlineSearch = true` in your `config.toml` then try searching, and press enter after entering a search term.

Expected: text box has high contast (maybe staying with a white background?)
Actual: text box has a blue background with black text on it.

image

Here's a video:

https://user-images.githubusercontent.com/105529/136772259-13eb5464-c967-4de3-97c3-696256feb5e8.mov

tekezo:
Change the search field background color to always be the active color when offline search is enabled to fix #718.

## Before

before

## After

after

### DIFF
--- a/assets/scss/_nav.scss
+++ b/assets/scss/_nav.scss
@@ -64,6 +64,16 @@
         @include placeholder {
             color: $navbar-dark-color;
         }
+
+        &.offline-search-input {
+            // Set the search field background color to active color when offline search is enabled.
+            // https://github.com/google/docsy/issues/718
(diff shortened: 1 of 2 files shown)


chalin (commented):
I believe that we can solve this more simply without introducing a new class. I'll investigate this later today.

chalin:
@tekezo - FYI, see #773 as an alternative that handles search wither it is offline or not.

PR #749, opened 2021-10-30
Use shallow clone for submodules

abitrolly:
This brings down the size of cloned repo a bit - from 306M to 265M.


LisaFC:
Good idea! I don't *see* any potential downside/breaking possibilities (nobody is going to need the entire history of Bootstrap, are they?)

chalin:
FYI, our build scripts are already setup to fetch shallow clones , see https://github.com/google/docsy/blob/280b1359a64e24857174e12bdb99d8ec10fdc040/package.json#L8-L9

Interesting, I didn't know about the [shallow option](https://git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtshallow). For some projects it makes sense to sometimes fetch a non-shallow clone, but I don't think that is the case here.

That being said, I'd vote to use the build scripts as they are now as opposed to the `shallow` option. If you do choose to merge this PR, then we should remove the `--depth 1` from the scripts.

deining:
Please note PR #269 which is identical to this PR. A look at this [comment](https://github.com/google/docsy/pull/269#issuecomment-626995225) might be useful. 
Issue [#84](https://github.com/google/docsy-example/issues/83) on the docsy-example site addresses shallow cloning, too.

LisaFC:
@chalin, presumably many users aren't using the build scripts? This will affect users who have Docsy itself as a submodule (I presume) and who are using their regular Git workflow to make local copies of their sites as per https://www.docsy.dev/docs/getting-started/#using-the-theme

As per previous PR (I'm sorry that fell through the net and didn't get resolved) it looks like we might need to make some changes to the docs if we do switch to the shallow option. Is there any other possible breakage/confusion for our existing users? Does this change help them? (I think so as it means their own local copy isn't pulling down as much non-Docsy stuff)

Also argh, Git submodules make my head hurt (for all that I use them and they're a very handy way of keeping dependencies up to date)

chalin:
For the record: on other projects, I've encountered situations where shallow cloning causes some extra headaches if you try to submit changes to submodules via the submodules themselves. I don't foresee that happening here.

chalin:
I've created an issue to track the subtasks associated with this change. See #755.

johnpaulett:
> As per previous PR (I'm sorry that fell through the net and didn't get resolved) it looks like we might need to make some changes to the docs if we do switch to the shallow option. Is there any other possible breakage/confusion for our existing users? Does this change help them? (I think so as it means their own local copy isn't pulling down as much non-Docsy stuff)

As an end-user of docsy, where I have a submodule to docsy in my repo (to the fork with the shallow PR), I also have `shallow = true` in my .gitmodules for docsy, also when checking out my application repository, I believe I also use `git clone --recurse-submodules <myapp.git>` so that it gets docsy, then the shallow FontAwesome etc.

For existing users, if they don't change anything, I don't think there is any negative.  

> Also argh, Git submodules make my head hurt (for all that I use them and they're a very handy way of keeping dependencies up to date)

100% agree on the head hurt. They are a useful tool when there is no other packaging solution, but the submodule UX always feels poorly integrated into the core git operations.  

While I'm not suggesting that shallow progress stop, it may be interesting to see if there is an alternative approach to distributing docsy other than a git clone?  Perhaps on a certain cadence tag docsy and use GitHub Releases to publish a tgz with the the minimal set of content (e.g. no `.git` folders for docsy / FontAwesome / bootstrap).  Then the User Guide could suggest `wget  <>.tgz` (end applications would incorporate this into their own build pipeline instead of using the git submodule)?  Only docsy developers would need the git clones. Definitely a bigger change, perhaps best discussed elsewhere.

abitrolly:
I made it for https://github.com/jenkins-x/jx-docs where the checkout process seemed long and heavy. It fetches `docsy` as submodule without build scripts.

abitrolly:
> While I'm not suggesting that shallow progress stop, it may be interesting to see if there is an alternative approach to distributing docsy other than a git clone? 

If anybody opens an issue, I'd subscribe.

chalin:
Closed by #269

PR #754, opened 2021-11-02
Fix production build, and other script tweaks

chalin:


netlify.toml
```diff
@@ -10,7 +10,4 @@ HUGO_VERSION = "0.88.1"
 HUGO_THEME = "repo"
 
 [context.production]
-command = "npm run build:production"
-
-[context.production.environment]
-HUGO_ENV = "production"
+command = "npm install && npm run build:production"
```
chalin, line 13: @LisaFC: this is the main change that fixes the production build. My bad for not propagating the preview fix to this line too.

netlify.toml
```diff
@@ -10,7 +10,4 @@ HUGO_VERSION = "0.88.1"
 HUGO_THEME = "repo"
 
 [context.production]
-command = "npm run build:production"
-
-[context.production.environment]
-HUGO_ENV = "production"
```
chalin, line 16: FYI, we're using Hugo command defaults and the `-e` flag to set the build mode so we don't need this anymore.

LisaFC:
Thanks for the really quick fix!

chalin:
For the record, the build is green again:

> Screen Shot 2021-11-02 at 12 58 28 PM

PR #759, opened 2021-11-05
Fix wide print margin

### LINKED ISSUE #688: Print layout is leaving wide right margin
When printing docs pages, whether from Ctrl-P or "Print entire section" the page looks fine when rendered, but when printed (print preview, saving as PDF, etc) there is always a wider right margin.  It appears as if the stylesheet might be leaving the space for the right-side navigation which isn't shown.

gwatts:
Don't really need margins at all for the print view

Resolves #688


LisaFC:
Thanks for this!

PR #760, opened 2021-11-05
fix: "This the multi-page …" -> "This is the multi-page …"

theneva:
Hi! 👋 

I stumbled upon this mistake in https://kubernetes.io/docs/reference/using-api/_print/, and figured I'd send a fix. I don't know the project at all, so I have no idea if any tests need to be updated – let me know if that's the case :smile:


chalin (approved):
LGTM

PR #772, opened 2021-11-17
sidebar-tree: dont show search if disabled

Finkregh:
before the search was always shown, ignoring `sidebar_search_disable`


LisaFC:
This was added only for mobile view, as otherwise there's no search bar! See #297

Finkregh:
i put the config option in the wrong place, thanks @LisaFC

PR #779, opened 2021-11-24
Add GH labels for effort, priority and more

chalin:


chalin:
@LisaFC @emckean - For the settings in this PR to be synced the https://probot.github.io/ needs to be installed -- I thought that it was enabled by default for GH projects. Hmm, in that case, I'll retract the settings and just add them by hand as the need arises.

chalin:
Another option, I suppose, would be to install & activate the probot, let it sync the labels and then deactivate it. WDYT @emckean @LisaFC? I haven't done this before myself so I'll inquire about the steps & effort involved.

Issue #783, opened 2021-11-25
Cleanup: don't hard-code styling into layouts


chalin:
We should avoid using Bootstrap utility classes (or even raw CSS) to hard-code styling of elements inside of layouts. Instead define an appropriate `td-` prefixed class, and add the styling to the class. Otherwise, it becomes difficult for clients to customize element styles.

Issue #782, opened 2021-11-25
Clean up `layouts/partials/breadcrumb.html`


chalin:
- [ ] What is the `spb-1` class applied to the `<ol>` element in the following excerpt:
  ```html
  <nav aria-label="breadcrumb" class="d-none d-md-block d-print-none">
	  <ol class="breadcrumb spb-1">
		  {{ template "breadcrumbnav" (dict "p1" . "p2" .) }}
	  </ol>
  </nav	>
  ```
  @LisaFC @emckean @bep any idea?
  It isn't in the generated styles or in the JS (as an anchor) AFAICT, so I'd vote to drop it.
  FYI, this file hasn't changed since the original commit 4 years ago: https://github.com/google/docsy/blame/master/layouts/partials/breadcrumb.html
- [x] Drop unnecessary whitespace from the partial and untabify. See the excerpt above. - #784
   Note the extra <tab> inside the `nav` closing tag.
- [x] Ensure it doesn't generate any unnecessary whitespace: e.g., use `{{ ... -}}` - #784

Issue #781, opened 2021-11-25
Support hiding single/root-page breadcrumbs


chalin:
Breadcrumbs on the root page just use up page real estate without adding any benefits:

> <img src="https://user-images.githubusercontent.com/4140793/143495583-b5faf2bc-a06a-47ce-8569-18eb2cb645ad.png" width=300>

One might argue uniformity of page design, but not all site pages have breadcrumbs anyway.

chalin:
At a minimum, this should be configurable, if only via styling. If/once #784 is merged, I'll show you what I have in mind.

LisaFC:
Yep, that makes sense to me.

PR #784, opened 2021-11-25
Show breadcrumbs on mobile too

### LINKED ISSUE #780: Show breadcrumbs on mobile
I'm not sure if there was a reason to hide breadcrumbs on narrow displays. I'd argue that the navigational benefits of breadcrumbs can be even more useful on mobile, and they don't have (IMHO) a negative impact on the page display. Here's an example of a deeply nested page:

> 

This looks perfectly fine to me.

Thoughts @LisaFC @emckean?

chalin:
- Closes #780
- Contributes to #783 -- note that the equivalent of the `d-print-none` was moved into the style definition so that clients can override this behavior if they so choose.
- Contributes to #782

Preview, for example visit the following links on mobile or in a browser with a narrow window:

- https://deploy-preview-784--docsydocs.netlify.app/docs
- https://deploy-preview-784--docsydocs.netlify.app/docs/adding-content/shortcodes

For a screenshot, see below.

No net change in the generated site files except for the `nav`'s change in class styles, whitespace being trimmed, and the new style definition:

```console
$ npm run build
...
$ (cd public && git diff -bw --ignore-blank-lines -I 'class="(td-breadcrumbs|d-none d-md-block d-print-none)"' -- . ':(exclude)*.xml' ':(exclude)*.map')
... see below for the syntax-highlighted diff ...
```
```diff
diff --git a/scss/main.css b/scss/main.css
index 5fcf278..b8f8160 100644
--- a/scss/main.css
+++ b/scss/main.css
@@ -13797,8 +13797,12 @@ nav.foldable-nav .with-child ul {
   .btn-sm, .btn-group-sm > .btn {
     border-radius: 1rem; }
 
-.breadcrumb {
-  background: none;
+@media print {
+  .td-breadcrumbs {
+    display: none !important; } }
+
+.td-breadcrumbs .breadcrumb {
+  background: inherit;
   padding-left: 0;
   padding-top: 0; }
```

#### Screenshot

>

### DIFF
--- a/layouts/partials/breadcrumb.html
+++ b/layouts/partials/breadcrumb.html
@@ -1,4 +1,4 @@
-<nav aria-label="breadcrumb" class="d-none d-md-block d-print-none">
+<nav aria-label="breadcrumb" class="td-breadcrumbs">
 	<ol class="breadcrumb spb-1">
 		{{ template "breadcrumbnav" (dict "p1" . "p2" .) }}
 	</ol>
--- a/layouts/partials/breadcrumb.html
+++ b/layouts/partials/breadcrumb.html
@@ -1,18 +1,20 @@
 <nav aria-label="breadcrumb" class="td-breadcrumbs">
-	<ol class="breadcrumb spb-1">
-		{{ template "breadcrumbnav" (dict "p1" . "p2" .) }}
-	</ol>
-</nav	>
-{{ define "breadcrumbnav" }}
-{{ if .p1.Parent }}
-{{ if not .p1.Parent.IsHome }}
-{{ template "breadcrumbnav" (dict "p1" .p1.Parent "p2" .p2 )  }}
-{{ end }}
-{{ else if not .p1.IsHome }}
-{{ template "breadcrumbnav" (dict "p1" .p1.Site.Home "p2" .p2 )  }}
-{{ end }}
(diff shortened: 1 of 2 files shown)


layouts/partials/breadcrumb.html
```diff
@@ -1,18 +1,20 @@
...
-{{ else if not .p1.IsHome }}
-{{ template "breadcrumbnav" (dict "p1" .p1.Site.Home "p2" .p2 )  }}
-{{ end }}
-{{ $isActive :=  eq .p1 .p2  }}
-<li class="breadcrumb-item{{ if $isActive }} active{{ end }}" {{ if $isActive }}aria-current="page"{{ end }}>
-	<a href="{{ .p1.Permalink }}">{{ .p1.LinkTitle }}</a>
-</li>
-{{ end }}
\ No newline at end of file
+<nav aria-label="breadcrumb" class="td-breadcrumbs">
```
chalin, line 1: The only significant change here is the use of the `td-breadcrumbs` class. Every other change in the partial is a whitespace fix.

chalin:
I'll take https://github.com/google/docsy/issues/780#issuecomment-980112295 as a thumbs up and merge this now. I'll add support for #781 before updating the docs.

PR #785, opened 2021-11-29
Folder 'userguide': adding README.md

deining:


chalin:
I prefer to avoid unnecessary repetition/duplication. The (top-level) repo README already has all the necessary information (in fact those commands could be updated to use NPM scripts instead). 

I'd vote to close this PR. @LisaFC @emckean - thoughts?

chalin:
Btw, I'm not against the addition of a README file. If we do decide to include the README, it should be at most a title and a link to the top-level repo README documentation section, IMHO.

LisaFC:
Agree about the lack of repetition but I wonder would it be useful to have something (possibly just by updating the top level README instead) explicitly telling users that the user guide files are in the /userguide/ subdir?

chalin:
> Agree about the lack of repetition but I wonder would it be useful to have something (possibly just by updating the top level README instead) explicitly telling users that the user guide files are in the /userguide/ subdir?

Yes, I like that! @deining, would you be willing to make that change? Thanks!

chalin (commented):
See @LisaFC's suggestion -- https://github.com/google/docsy/pull/785#issuecomment-982805844

deining:
> Yes, I like that! @deining, would you be willing to make that change? Thanks!

Following your proposal, I authored the requested changes.

deining:
Another comment: Currently, `README.md` states:

```
git clone --recurse-submodules --depth 1 https://github.com/google/docsy.git
cd docsy/userguide/
hugo server --themesDir ../..
```

This instructions aren't really helpful for users who cloned the repo already.

> I prefer to avoid unnecessary repetition/duplication. The (top-level) repo README already has all the necessary information (in fact those commands could be updated to use NPM scripts instead).

I like this idea. If npm scripts were implemented, we could overcome the limit mentioned above.

chalin:
> If npm scripts were implemented

They are. To set things up one must run `npm install`, after that:

- To build, run `npm build`
- To serve locally, run `npm serve`

chalin:
@deining, you wrote:

> > I prefer to avoid unnecessary repetition/duplication. The (top-level) repo README already has all the necessary information (in fact those commands could be updated to use NPM scripts instead).
> 
> I like this idea. If npm scripts were implemented, we could overcome the limit mentioned above.

I think then that we can close this PR?

deining:
> I think then that we can close this PR?

Yes, closing this now.

PR #786, opened 2021-11-29
Support hiding single/root breadcrumbs

### LINKED ISSUE #781: Support hiding single/root-page breadcrumbs

### LINKED ISSUE #782: Clean up `layouts/partials/breadcrumb.html`

chalin:
- Closes #781
- Closes #782

To hide, use:

```scss
.td-breadcrumbs__single { display: none !important; }
```

Preview (w/o showing the hiding activated):

- https://deploy-preview-786--docsydocs.netlify.app/docs/adding-content/shortcodes

### DIFF
--- a/layouts/partials/breadcrumb.html
+++ b/layouts/partials/breadcrumb.html
@@ -1,5 +1,6 @@
-<nav aria-label="breadcrumb" class="td-breadcrumbs">
-  <ol class="breadcrumb spb-1">
+<nav aria-label="breadcrumb" class="td-breadcrumbs
+    {{- if .Parent.IsHome }} td-breadcrumbs__single {{- end }}">
+  <ol class="breadcrumb">
     {{- template "breadcrumbnav" (dict "p1" . "p2" .) }}
   </ol>
 </nav>


LisaFC:
Merging, but we should also add something to https://www.docsy.dev/docs/adding-content/navigation/#breadcrumb-navigation to explain how to override the styling

PR #520, opened 2021-04-16
Make docsy a go module (with its own submodules)

deining:
Making docsy a go module was discussed in [issue 456](https://github.com/google/docsy/issues/456#issuecomment-786122011) already. I'm citing from there:

```
I think we looked at making Docsy into a module
(with its own modules - it currently has Git submodules)
and it would have required a fair bit of refactoring, (...).
It's definitely something to revisit though.
```

I worked myself into hugo modules recently and ended up with this PR which turns docsy into a module —without the need of any refactoring of the existing codebase. The only change needed is to add a `module` section (with several `imports` and `mounts`) to `config.toml`.

With a new hugo site set up and a fresh copy of the docsy theme cloned to the `themes` folder, let's compare the proposed `modules` approach with the current `submodules` approach:

**Resolving dependencies with Submodules (current HEAD)**
```
$ # inside /path/to/mysite/themes
$ time git submodule update --init --recursive
Submodule 'assets/vendor/Font-Awesome' (https://github.com/FortAwesome/Font-Awesome.git) registered for path 'assets/vendor/Font-Awesome'
Submodule 'assets/vendor/bootstrap' (https://github.com/twbs/bootstrap.git) registered for path 'assets/vendor/bootstrap'
Cloning into '/path/to/docsy/assets/vendor/Font-Awesome'...
Cloning into '/path/to/docsy/assets/vendor/bootstrap'...
Submodule path 'assets/vendor/Font-Awesome': checked out 'fcec2d1b01ff069ac10500ac42e4478d20d21f4c'
Submodule path 'assets/vendor/bootstrap': checked out '6ffb0b48e455430f8a5359ed689ad64c1143fac2'

real    0m29.895s
```  
**Pulling in dependencies with Docsy as Hugo module (this PR applied)**
```
$ # inside /path/to/mysite/themes
time hugo mod vendor --ignoreVendorPaths **
hugo: downloading modules …
hugo: collected modules in 6274 ms

real    0m7.421s
```  

As you can see, the process is **more than 4 times faster** now.

Let's compare downloaded file sizes:

**With Submodules (current HEAD)**
```
$ # inside /path/to/mysite/_vendor
$ du . -h
12K     ./github.com/deining/hugo-mod-font-awesome-scss
12K     ./github.com/deining
254K    ./github.com/FortAwesome/Font-Awesome/scss
254K    ./github.com/FortAwesome/Font-Awesome
254K    ./github.com/FortAwesome
[...]

Issue #283, opened 2020-06-11
Can't get docsy to work with hugo modules.


wlai-axway:
Hi, I am trying to use hugo modules to manage dependencies. When using hugo modules then everything has to be configured this way and the theme = ["docsy"] configuration needs to be replaced by:

> [module]
>   [[module.imports]]
>     path = "github.com/google/docsy"
>#theme = ["docsy"]

That doesn't work because docsy uses submodules and hugo modules doesn't seem to take care of that so hugo build results in failure due to missing bootstrap or Font-Awesome files.

Alternatively the docsy theme can still be kept as a git submodule and you tell hugo to load it as a module using this configuration:
> [module]
>   [[module.imports]]
>     path = "docsy"
>#theme = ["docsy"]

That configuration seems to work fine and hugo builds with no errors but the generated pages doesn't seem right. Here's a test using https://github.com/google/docsy-example :
![image](https://user-images.githubusercontent.com/20820993/84372972-d4829c00-abd3-11ea-9772-72ed7090778d.png)

LisaFC:
Hmm, I'll have a look. We built the theme before Hugo modules existed so it may need some tweaking to work with them.

d-henn:
Is this still something being looked at? Would love to be able to convert to hugo modules!

deining:
Support for using the theme as hugo module was introduced with #871, so this ticket can be closed.

PR #796, opened 2021-12-07
Preparations for using docsy theme as hugo module

deining:
Using docsy theme as module was discussed previously (e.g. #283, #520). Feedback was encouraging, so I went ahead and I'm really close to a submit a PR that offers the possibility to use docsy theme as module. My module approach can coexist with the current submodule solution, so it's up to the user to decide which method to choose. I'm super excited and I see a lot of advantages when using modules (modules offer the possibility to check whether and extended hugo version is used, we can check for min version of hugo, we can pull in content from different repos, ...). This PR prepares the way for using docsy as module and is mandatory for my next PR which will turn docsy into a module. This PR contains one file only and is a workaround for a bug in the module handling of `go` language.

@LisaFC : Can you please merge this PR? Also, I intend to prepare the docsy-example site for using the theme as module, too. To avoid version confusion with bootstrap and Font-Awesome, it were good if [PR 155](https://github.com/google/docsy-example/pull/155) on the docsy-example repo could the processed, too. Thanks in advance.

### DIFF
--- a/assets/bootstrap/scss/vendor/_rfs.scss
+++ b/assets/bootstrap/scss/vendor/_rfs.scss
@@ -1,0 +1,228 @@
+// stylelint-disable property-blacklist, scss/dollar-variable-default
+
+// SCSS RFS mixin
+//
+// Automated responsive font sizes
+//
+// Licensed under MIT (https://github.com/twbs/rfs/blob/v8.x/LICENSE)
+
+// Configuration
+
+// Base font size
+$rfs-base-font-size: 1.25rem !default;
+$rfs-font-size-unit: rem !default;
+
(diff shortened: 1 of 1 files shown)


assets/bootstrap/scss/vendor/_rfs.scss
```diff
@@ -0,0 +1,228 @@
+// stylelint-disable property-blacklist, scss/dollar-variable-default
```
chalin, line 1: Why is this file being brought in separately as opposed to getting it via the module?

chalin:
Can this be closed now that #806 was merged into the feature branch?

deining:
> Can this be closed now that #806 was merged into the feature branch?

Yes, this issue is obsolete now.

PR #800, opened 2021-12-10
Plantuml svg loader

erwin-faceit:
This PR gives the possibility to use an SVG loader instead of an using an img tag when using PlantUML. The reason for this is that the image tag does not support links or other responsive actions in the generated UML diagram.

By default, nothing changes. If you want this PR to be enabled, you'll need to change the configuration like this
```
[params.plantuml]
enable = true
svg = true
```

Setting the params.plantuml.svg property to true will add an additional JavaScript file (svg-loader.min.js), and will write svg tags instead of img tags.


LisaFC:
Ooh, this looks very helpful. Do you want to also add some instructions to https://www.docsy.dev/docs/adding-content/diagrams-and-formulae/#uml-diagrams-with-plantuml so users will know how to enable this feature?

erwin-faceit:
@LisaFC: A small line perhaps, but is this enough, or would you prefer a more extensive writeup about this?

LisaFC:
No, that's perfect! I'm sure someone will let us know if they want more information.

deining:
I just realized that this PR sets submodule `bootstrap` back to `v4.6.0`, while we were at `v4.6.1` already. Is/was this intended?

erwin-faceit:
No, this was not intended! Did I merge the main branch incorrectly?

PR #807, opened 2021-12-15
Breadcrumb fix for sites with docs at the root

### LINKED ISSUE #804: executing "partials/breadcrumb.html" at <.Parent.IsHome>: nil pointer evaluating page.Page.IsHome
Hi there, on https://gitlab.com/graphviz/graphviz.gitlab.io/, updating to the latest Docsy, I get this error:

```
git clone https://gitlab.com/graphviz/graphviz.gitlab.io.git
cd graphviz.gitlab.io
git submodule update --remote
hugo serve
Start building sites … 
hugo v0.89.4+extended darwin/amd64 BuildDate=unknown
Error: Error building site: failed to render pages: render of "home" failed: execute of template failed: template: docs/list.html:23:62: executing "docs/list.html" at <partial "breadcrumb.html" .>: error calling partial: "/Users/mark/projects/graphviz.gitlab.io/themes/docsy/layouts/partials/breadcrumb.html:2:18": execute of template failed: template: partials/breadcrumb.html:2:18: executing "partials/breadcrumb.html" at <.Parent.IsHome>: nil pointer evaluating page.Page.IsHome
```

Looks like we're trying to read `Parent.IsHome` in `breadcrumb.html` but not every page has a `.Parent`. 

https://github.com/google/docsy/blob/e7fd4d8c266182f28123364e0e40c882047ff4da/layouts/partials/breadcrumb.html#L2

In particular, looks like the `home` page doesn't have a parent.

Maybe this is specific to sites that use the `docs` layout for their root?

Looks like @chalin (tagging as FYI) updated breadcrumbs.html 7 days ago in https://github.com/google/docsy/pull/786.

chalin:
- Closes #804
- Tested this over https://graphviz.org/ and Lisa's JustDocs (both of which are mentioned in #804)
- Note that this solution assumes that docs section(s) are either at the site root or at most one level below the root.
- There are no changes to the generated User Guide site files.

### DIFF
--- a/layouts/partials/breadcrumb.html
+++ b/layouts/partials/breadcrumb.html
@@ -1,5 +1,9 @@
+{{ $isSingle := true -}}
+{{ with .Parent -}}
+  {{ $isSingle = .IsHome -}}
+{{ end -}}
 <nav aria-label="breadcrumb" class="td-breadcrumbs
-    {{- if .Parent.IsHome }} td-breadcrumbs__single {{- end }}">
+    {{- if $isSingle }} td-breadcrumbs__single {{- end }}">
   <ol class="breadcrumb">
     {{- template "breadcrumbnav" (dict "p1" . "p2" .) }}
   </ol>


chalin:
@emckean gave a thumbs up out-of-band.
I'll go ahead a merge.
@LisaFC - comments are still welcome.

PR #801, opened 2021-12-10
Turning docsy theme into a  hugo module

deining:
This is my final approach to bringing hugo modules to doscsy theme. This PR supersedes with previous attempt #520. This PR ist the module part only, docu comes shortly with a separate PR.
Motiavation behind this was already I you explained [here](https://github.com/google/docsy/pull/520#issue-859965814)
From the discussion at #535, I learned that this PR will be (hopefully!) appreciated by the Docsy community.
A recent [blog post](https://chrisshort.net/from-git-submodule-to-hugo-modules-using-netlify/) can serve as additional reference, I agrre with what is stated there.
Please note, that this PR is not ready for review yet, I'm missing a commit hash that I can refer to.
There are different approaches to get this PR ready to that reviewers and/or community can try it out:

- Merge #796 into master. Once we are reluctant to do so, we can:
- create a new branch `modules` in the dosy repo, I will submit then resubmit the patch to this branch.

I hope that my PR will be well received and that it will be accepted eventually.

PR #811, opened 2021-12-17
Make hugo modules run successfully

deining:


deining:
This is now deprecated in favor if #801.

Issue #822, opened 2022-01-07
Unable to check out submodules


killermoehre:
Hi,

after a fresh `clone` I hit following error for checking out the submodules

```shell
$ LANG=C git clone https://github.com/google/docsy.git
$ LANG=C git submodule update --init --recursive --remote --force
Submodule 'assets/vendor/Font-Awesome' (https://github.com/FortAwesome/Font-Awesome.git) registered for path 'assets/vendor/Font-Awesome'
Submodule 'assets/vendor/bootstrap' (https://github.com/twbs/bootstrap.git) registered for path 'assets/vendor/bootstrap'
Submodule path 'assets/vendor/Font-Awesome': checked out '7cbd7f9951be31f9d06b6ac97739a700320b9130'
fatal: Needed a single revision
fatal: Unable to find current origin/v4-dev revision in submodule path 'assets/vendor/bootstrap'
```

As we use `docsy` as submodule in our own Hugo repo, this is quite annoying as we can't build our site.

deining:
> Hi,
> ...
> ```shell
> $ LANG=C git submodule update --init --recursive --remote --force
> ```

Why are you using the --remote option?

Simply omit this option and it will work:

```shell
git submodule update --init --recursive

**Comment by killermoehre** [comment]
I require the `--remote` option to actually update some other repository where `docsy` itself is a submodule of.

**Comment by deining** [comment]
You can overwrite the branch in your `.git/config`, the settings there take precedence over the settings given in `.gitmodules`. (either use `branch = main` or `branch = HEAD`).

```
[submodule "assets/vendor/bootstrap"]
	active = true
	url = https://github.com/twbs/bootstrap.git
	branch = HEAD
```

Be aware however, since docsy pulls in bootstrap version 4.6.1, your submodule will be out of sync.

**Comment by killermoehre** [comment]
Well, `docsy` doesn't find the `v4-dev` branch it is pointing to. I don't even want `HEAD`, I just want docsy to pull from the current `v4-dev` branch as stated in the `.gitmodule`-

**Comment by killermoehre** [comment]
The linked PR is closed. Therefore this issue can be closed.

**Comment by killermoehre** [comment]
My bad, the PR #823 is not yet merged.

**Comment by deining** [comment]
Can you please try:

```
git submodule sync
```

Does this resolve your problem?

killermoehre:
I'm switching to proper hugo modules with
```toml
[module]
[[module.imports]]
  disable = false
  ignoreConfig = false
[...]

PR #823, opened 2022-01-07
Update submodules to point to current branch

killermoehre:
Updating both submodules `bootstrap` and `Font-Awesome` to their current commit in the respective branch.

Closes: #822


killermoehre:
I signed the CLA. Please re-scan.

LisaFC:
Looks good, @deining can you take a look as well? Preview looks fine.

deining:
Docsy pulls in bootstrap 4.6.1 in layouts/partials/scripts.html`:

```
<script src="https://cdn.jsdelivr.net/npm/bootstrap@4.6.1/dist/js/bootstrap.min.js"
    integrity="sha512-UR25UO94eTnCVwjbXozyeVd6ZqpaAE9naiEUBK/A+QDbfSTQFhPGj5lOR6d8tsgbBk84Ggb5A3EkjsOgPRPcKA=="
    crossorigin="anonymous"></script>
```

Dosy repo, HEAD has submodule bootstrap @ 043a03c
Revision 043a03c is tagged as `v4.6.1`, so everything looks fine to me here.

This PR wants to move submodule forward to 349a373f, current HEAD of branch `v4-dev`, 17 commits ahead of tag `v4.6.1`. I don't see a real reason for this change since script and submodule are out of sync then. Probably it won't do harm, but what is the motivation behind this? Resolving #822? I don't know the exact setting there, so it's hard to tell what's going on and what causes the problems described in #822. However, my feeling is that there must be another way to resolve issue #822. Unfortunately, without knowing further details, I'm not really able to help out to resolve #822. I guess we should implement modules and @killermoehre should discard submodules in favour of Hugo module 😉 .

killermoehre:
Well, if the branch shouldn't follow `v4-dev`, but should point to the tag `v4.6.1`, than I think this is an issue in of itself. I mean, what is `.gitmodules` if not a clear pointer to which code should be used?

I will check out Hugo modules. I took over some code from a colleague, so I don't know about it, yet.

LisaFC:
Hmm, this kind of thing is definitely an argument for using Hugo modules...

Let me check when we updated to 4.6.1 for Bootstrap. Our major concern with Bootstrap is not going on to v5 yet as we don't quite support it yet (one of our roadmap items!).

deining:
Frankly spoken, I'm a bit skeptical whether we really should merge this PR. I would recommend to merge #801 instead, which was just finalized by me. Once #801 gets merged into `master` using docsy theme as Hugo module should work, so @killermoehre can switch his site to make use of docsy as module. As side effect, we get immediate feedback whether #801 works as expected (I'm confident it does).

PR #824, opened 2022-01-08
Remove a .DS_Store file

11nu:
`.DS_Store` has already been added to `.gitignore`, so I would say it should be removed.

I have checked that it does not exist in any other directory using the following shell command.

```bash
# Run in the root directory of the repository.
find . -name ".DS_Store"
```


LisaFC:
Thanks!

PR #828, opened 2022-01-12
gh-827 Change height of rh sidebar to use all of screen

### LINKED ISSUE #827: Right hand TOC sidebar not using full screen height
As you can see in this screen grab (of the bottom right corner of the screen) the right hand TOC sidebar is not using all the available screen when not scrolled to the bottom, so the vertical divider is short of the bottom.

image

If you scroll all the way down it is fine as the footer takes up the space. 

image

Will raise a PR for the fix.

at055612:
Fixes #827


chalin (approved):
Thanks!

PR #841, opened 2022-01-17
Simplify setup: install Hugo automatically (and get the right version) using NPM

chalin:
This PR introduces the use of the [hugo-extended](https://www.npmjs.com/package/hugo-extended) package, a "Plug-and-play binary wrapper for Hugo Extended".

By doing so we **reduce the burden on contributors**: the _right_ version and variant (_extended_ edition), gets fetched automatically when `npm install` is run. Also, from now on, every version of the website will be associated to the version of Hugo needed to build it, and npm will fetch the proper version on an install.

@LisaFC @emckean et all - WDYT? If you agree with this change, I'll make followup PRs to update the README, etc.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,11 +3,11 @@
 
 [build]
 publish = "userguide/public"
-command = "npm install && npm run build:preview"
+command = "npm run build:preview"
 
 [build.environment]
 GO_VERSION = "1.17.6"
 HUGO_THEME = "repo"
 
 [context.production]
-command = "npm install && npm run build:production"
+command = "npm run build:production"
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,7 +6,6 @@ publish = "userguide/public"
 command = "npm install && npm run build:preview"
 
 [build.environment]
-HUGO_VERSION = "0.92.0"
 GO_VERSION = "1.17.6"
 HUGO_THEME = "repo"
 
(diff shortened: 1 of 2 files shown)


chalin:
@LisaFC @emckean - can either of you grant me the permissions to clear the Neltify deploy caches, or clear the cache and restart the deploy for this PR? I'd like to ensure that the build will pass (which it is now), even when run from an empty deploy cache. Thanks!

emckean:
glorious! I like the idea of making the default easy :) 

I don't have Netlify privileges (I don't think ...) though

chalin:
Thanks for the feedback @emckean.

We can merge this w/o the cache-clearing test.

Can I go ahead and merge?

chalin:
Just to be clear: if ever there's an issue with the cache, it won't have to do with the `hugo-extended` package, but rather the simplifications I've done to `netlify.toml`, which I can easily revert if ever an issue surfaces.

chalin:
I'm going to take @emckean's "glorious" comment as an approval and merge this. :)

chalin:
And the production build and deploy is green:

> image

LisaFC:
This is great, thanks! Let me know if there's anything I need to do with my Netlify privileges, but it doesn't look like it as the production build is working.

chalin:
> If you agree with this change, I'll make followup PRs to update the README, etc.

Hmm, I just realized that the README in this project doesn't describe how to build docsy and the user guide. We should change that: let's discuss this in #846.

Issue #848, opened 2022-01-19
Ensure a smooth transition for existing docsy-as-submodule users


chalin:
To give some initial context (excerpt from https://github.com/google/docsy/discussions/535#discussioncomment-1986956):

>  - **Update of existing sites**: That's for sure the most critical part. We do have two [testimonies](https://github.com/google/docsy/issues/836#issuecomment-1014935503) that users can upgrade to HEAD of repo and use their site successfully. In it's current form, there is some manual work needed. @LisaFC @chalin Is that acceptable? Yes or no? We need to make a decision here soon.
>    That's what currently is [needed]() to make the site work again:
>    - navigate to your `themes` folder
>    - create two new directory structures (empty directories):
>     ```
>    mkdir -p github.com/twbs/bootstrap
>    mkdir -p github.com/FortAwesome/Font-Awesome
>    ```
>    No code change, simply create two empty directories and your site is up again. Is this acceptable? I don't see any **good** way to make it a seamless update without any intervention. I asked on the hugo discourse list, haven't received any answer yet. I doubt whether anyone will still come up with a good answer.
> *Disclaimer:* Yes, there actually *is* a way to make it a seamless update without any intervention. But it's not clean, and I don't like it.

chalin:
I've actually now tried to migrate one of our sites, while seeking to continue using docsy via a submodule. As @deining mentions, the following command helps ensure that I can build locally by creating some empty directories:

```console
$ mkdir -p themes/github.com/{twbs/bootstrap,FortAwesome/Font-Awesome}
```

By default, git won't commit (and in any case will prune) empty directories. So, the above solution **breaks my remote builds** (unless I include that command as part of my build process, which I don't want to do). Thoughts @deining?

deining:
> Thoughts @deining?

O.k. Then we have to expand our adaption routine a bit:

```
mkdir -p github.com/twbs/bootstrap
touch github.com/twbs/bootstrap/.gitkeep
mkdir -p github.com/FortAwesome/Font-Awesome
touch github.com/twbs/bootstrap/.gitkeep
```

Does this solve your problem?

chalin:
Yes, it would likely solve the problem, but is starting to feel a little hackish. Earlier you wrote:

> there actually _is_ a way to make it a seamless update without any intervention. But it's not clean

Would you mind sharing what this seamless update path would be?

deining:
> Yes, it would likely solve the problem, but is starting to feel a little hackish.

[...]

PR #852, opened 2022-01-19
 Ensure a smooth transition for existing docsy-as-submodule users #848

deining:
This PR solves #848. It implements the following behaviour:

**Using docsy as git submodule**

```
hugo serve
```

**Using docsy as hugo module**

```
hugo serve --environment module
```

This PR works as of hugo version [0.84](https://gohugo.io/news/0.84.0-relnotes/#themes-now-support-the-config-directory). With this version, hugo learned to make proper use of the `config` directory inside themes. This is the minimum required version when using docsy theme as module. Using docsy as submodule might work with earlier versions, too, which is good since the majority of sites will still run docsy as submodule.


chalin (commented):
Thank you for adding the READMEs and the useful links and explanations that they contain.

This solution seems promising, but I'm concerned that it won't work for projects who want to use modules in a `production` environment.  Might we have to revert to using multiple config files instead?

deining:
> This solution seems promising, but I'm concerned that it won't work for projects who want to use modules in a `production` environment.

Can you elaborate con your concerns a bit, please? What are you exactly concerned about? Why do you think that a multiple config approach might be more promising?

chalin:
> Can you elaborate con your concerns a bit, please? What are you exactly concerned about?

Some docsy code makes use of [hugo.IsProduction] (a feature that was added by https://github.com/gohugoio/hugo/pull/6953). As far as I can tell, there can be only a _single_ value for [hugo.Environment], so setting it to `module` prevents projects from using `production` in production builds, etc. Does that make sense?

Unless we can find a way to allow [hugo.IsProduction] to be true while still using Docsy as a module, we won't be able to use the config directory approach.

> Why do you think that a multiple config approach might be more promising?

I don't know if using multiple config files will work, but it would be worth exploring. WDYT?

[hugo.Environment]: https://gohugo.io/functions/hugo/#prose
[hugo.IsProduction]: https://gohugo.io/functions/hugo/#prose

deining:
> Some docsy code makes use of [hugo.IsProduction](https://gohugo.io/functions/hugo/#prose) (a feature that was added by [gohugoio/hugo#6953](https://github.com/gohugoio/hugo/pull/6953)). As far as I can tell, there can be only a _single_ value for [hugo.Environment](https://gohugo.io/functions/hugo/#prose), so setting it to `module` prevents projects from using `production` in production builds, etc. Does that make sense?

Yes, I see the point.

> Unless we can find a way to allow [hugo.IsProduction](https://gohugo.io/functions/hugo/#prose) to be true while still using Docsy as a module, we won't be able to use the config directory approach.

I like the config directory approach, so I tried to make in work in another way: we now do have two config directories and we can use hugo's `--configDir` option to select one. That's how it is supposed to work:

**Using docsy as git submodule**

```
hugo serve --configDir configTheme
```

**Using docsy as hugo module**

```
hugo serve --configDir configModule
```

Empty default `config` directory at the beginning, the user has to give `--configDir` to get it up and running. After successful installation, the user may symlink `config` directory to one of both directories to make it the default way of running.
Checks are failing now for this PR, this in on purpose. Not being able to start up the site in the default setting via `hugo serve` may sound harsh at first glance, but I see three advantages:
- both installation types (submdule/module) are equivalent, no need to discuss about the recommended way of installation, and
- the installation will get fail safe this way. With `--environment` flag approach, it you want to have module installation type and forget to add `-e module` at intial startup, I have seen weird things happen (pulling in bootstrap 5, ...). I'm afraid that users who haven't dealt with hugo modules already before can get easily lost here.
- inside the empty ´config` dir, we can put file `README.txt`, explaining how to make either submodules of modules the default. With multiple config files, that wouldn' t work out that well.

@chain, WDYT? If you like it, I will start testing my new approach.

chalin:
That's a very interesting proposal! Let me think about it and get back to you shortly.

chalin:
For my detailed feedback, see https://github.com/google/docsy/issues/848#issuecomment-1018478162.

LisaFC:
> I like the config directory approach, so I tried to make in work in another way: we now do have two config directories and we can use hugo's `--configDir` option to select one. That's how it is supposed to work:
> 
> **Using docsy as git submodule**
> 
> ```
> hugo serve --configDir configTheme
> ```
> 
> **Using docsy as hugo module**
> 
> ```
> hugo serve --configDir configModule
> ```


Ooh, that's a very interesting approach, though does it mean that if you just run `hugo serve` nothing will happen? Is there a way to make our current "submodules" behaviour the default as it means it won't break people's continuous deployment (including ours)?

deining:
> Ooh, that's a very interesting approach, though does it mean that if you just run `hugo serve` nothing will happen?

Yes, in its current form, nothing happens, which will break existing build. Not good, no option at all.

> Is there a way to make our current "submodules" behaviour the default as it means it won't break people's continuous deployment (including ours)?

Ther ecertainly is a way: we simply copy `config.toml` from `configTheme` the into site root. This will bring back the default submodule functionality to docsy, this is a must since we don't want to break exiszting builds. The question now is: With this file added, are we still allowed to use as hugo module. I have a way in mind, but I'm not sure that it works. I need to test it. Fortunately I do have a playground branch on my own, this wasn't the case in the past. I will come up with the results shortly.

deining:
This PR is superseded by #871.

PR #859, opened 2022-01-22
CI: reintroduce explicit call to npm install

chalin:
Reverting the change to `netlify.toml` as mentioned in https://github.com/google/docsy/pull/841#issuecomment-1015903812.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,11 +3,11 @@
 
 [build]
 publish = "userguide/public"
-command = "npm run build:preview"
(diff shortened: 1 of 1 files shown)


chalin:
I'm suspecting that there might be something in the Netlify config that is disabling automatic calls to `npm install` at the start of a build. Anyhow, this adds an explicit call, which hopefully doesn't happen twice.

chalin:
Production builds are green once again:

>

Issue #858, opened 2022-01-21
Fix the build for docsy-as-submodule


chalin:
_Originally posted by @LisaFC in https://github.com/google/docsy/discussions/535#discussioncomment-1963785_:

> Correct - feature is in but disabled by default, docs have not been updated yet.  Anyone who wants to try it can use the preview for the docs PR for more info. The docs update is sizeable and I want to take some time to do some organization work on it.

Based on the above comment, I believe that it wasn't @LisaFC's intention to merge #801 and introduce such a **serious breaking change**: #801 **broke the builds of most users**.

I believe that our first order of business is to (at least partially) **revert #801 so that all users can continue to work from HEAD**, as we continue to work through the details of how to introduce modules as an _optional_ feature (cf. #848).

We should also restore the build of the user-guide to be submodule based, since that reflects how most users are including docsy into their projects, and as such, it acts as a test that docsy@HEAD hasn't broken anything for submodule users.

@LisaFC @emckean @geriom @deining - thoughts.

LisaFC:
Yes, it might be a good idea to either revert or (if it's nearly ready) #852 looks like it might be a way to fix things for everyone while still including the modules option (ie do what we thought #801 would do) - though it would need to have the submodule config as the default rather than one of two "equal options" to not break people's continuous deployment.

chalin:
> #852 looks like it might be a way to fix things for everyone while still including the modules option (ie do what we thought #801 would do) - though it would need to have the submodule config as the default rather than one of two "equal options" to not break people's continuous deployment.

That works for me!

(We'll still need to refer the UG so that it builds from the local repo -- not a Hugo module.)

deining:
> Based on the above comment, I believe that it wasn't @LisaFC's intention to merge #801 and introduce such a **serious breaking change**: #801 **broke the builds of most users**.

It wasn't my intention either, and I wasn't aware of the fact that it **would** break builds. Anyway, that's history now, let's look forwards.

I think finalizing #852 is the way to go, I just presented a [further idea](https://github.com/google/docsy/pull/852#issuecomment-1018728448) how to make it work in the desired way. Due to other private and job-related oblgaions, I couldn't work on this during the week as wanted. I will investigate the opportunities during the weekend and will post my findings shortly.

chalin:
> It wasn't my intention either, and I wasn't aware of the fact that it **would** break builds.

Yes, of course, I believe you.

> I think finalizing #852 is the way to go, ... will present my findings shortly.

[...]

PR #862, opened 2022-01-22
Fix the build for docsy-as-submodule (#858)

deining:
I see our commits crossed. I think essentially, they were the same.
Prepared a patch for reverting the userguide to git submodules already, stay stuned, it will come in shortly.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -29,22 +29,4 @@ permalinkable = false
 [module]
 [module.hugoVersion]
 extended = true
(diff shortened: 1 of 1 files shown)


chalin:
> I see our commits crossed. I think essentially, they were the same.

Yes. Sorry about that. I figured I could do this part while you focused on getting modules up and running :). Thanks.

deining:
Patrice was faster, thank you!

PR #863, opened 2022-01-22
Fix the build for docsy-as-submodule (#858)

### LINKED ISSUE #850: User-guide build: revert changes so that it uses the docsy as a submodule
This will ensure that we have at least one test case in support of using docsy as a submodule. This is in support of testing for #848.

For context, the UG build was changed in #801.

/cc @deining

deining:
This PR resolves #850. Also doing some whitespace cleanup.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -25,3 +25,8 @@ isHTML = true
 mediaType = "text/html"
 path = "_print"
 permalinkable = false
+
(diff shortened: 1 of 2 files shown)


deining:
#860 covers this too, so I'm closing this PR.

PR #864, opened 2022-01-26
i18n support: community page 

h1z3y3:
I make community page support i18n.

*I don't know other languages so I fill over them by English.*

### DIFF
--- a/layouts/community/list.html
+++ b/layouts/community/list.html
@@ -4,9 +4,9 @@
 <section class="row td-box td-box--1 position-relative td-box--gradient td-box--height-auto">
 	<div class="container text-center td-arrow-down">
 		<span class="h4 mb-0">
-<h1>Join the {{ .Site.Title }} community</h1>
+<h1>{{ T "community_join" }} {{ .Site.Title }} {{ T "community" }}</h1>
 
-<p>{{ .Site.Title }} is an open source project that anyone in the community can use, improve, and enjoy. We'd love you to join us! Here's a few ways to find out what's happening and get involved.
+<p>{{ .Site.Title }} {{ T "community_introduce" }}
 </span>
 	</div>
 </section>
--- a/layouts/partials/community_links.html
+++ b/layouts/partials/community_links.html
@@ -2,19 +2,19 @@
 
 <section class="row td-box td-box--4 td-box--gradient td-box--height-auto linkbox">
 <div class="col-xs-12 col-sm-6 col-md-6 col-lg-6">
-<h2>Learn and Connect</h2>		
-<p>Using or want to use {{ .Site.Title }}? Find out more here:
+<h2>{{ T "community_learn" }}</h2>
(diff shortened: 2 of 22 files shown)


LisaFC:
Oh that's great, thanks! Hopefully some of the other users will update the files for their languages.

layouts/partials/community_links.html
```diff
@@ -2,19 +2,19 @@
...
+<p>{{ T "community_using" }} {{ .Site.Title }}? {{ T "community_find_out" }}:
 {{ with index $links "user"}}
 {{ template "community-links-list"  . }}
 {{ end }}
 </div>
 <div class="col-xs-12 col-sm-6 col-md-6 col-lg-6">
-<h2>Develop and Contribute</h2>
-<p>If you want to get more involved by contributing to {{ .Site.Title }}, join us here:
+<h2>{{ T "community_develop" }}</h2>
+<p>{{ T "community_contribute" }} {{ .Site.Title }}, {{ T "community_join_us" }}:
```
deining, line 13: I don't like the way translations are implemented here: `.Site.Title` is always at the end of the sentence, which is true for English, but may not be true for other languages. I think we should pass the `.` context when calling T here. I would like to see something like:

```
<p>{{ T "community_contribute" . }}, {{ T "community_join_us" }}:
``` 

Inside my de.toml, I could then use:

```
[community_contribute]
other = "Wenn Du aktiv am {{ .Site.Title }}-Projekt beteiligen und einen Beitrag dazu leisten möchtest"
```

layouts/community/list.html
```diff
@@ -4,9 +4,9 @@
 <section class="row td-box td-box--1 position-relative td-box--gradient td-box--height-auto">
 	<div class="container text-center td-arrow-down">
 		<span class="h4 mb-0">
-<h1>Join the {{ .Site.Title }} community</h1>
+<h1>{{ T "community_join" }} {{ .Site.Title }} {{ T "community" }}</h1>
 
-<p>{{ .Site.Title }} is an open source project that anyone in the community can use, improve, and enjoy. We'd love you to join us! Here's a few ways to find out what's happening and get involved.
+<p>{{ .Site.Title }} {{ T "community_introduce" }}
```
deining, line 9: Again, I don't like the way translations are implemented here: Passing the . context when calling T seems the way to go here, too.

layouts/community/list.html
```diff
@@ -4,9 +4,9 @@
 <section class="row td-box td-box--1 position-relative td-box--gradient td-box--height-auto">
 	<div class="container text-center td-arrow-down">
 		<span class="h4 mb-0">
-<h1>Join the {{ .Site.Title }} community</h1>
+<h1>{{ T "community_join" }} {{ .Site.Title }} {{ T "community" }}</h1>
```
deining, line 7: Again, I don't like the way translations are implemented here: Passing the . context when calling T seems the way to go here, too.

deining (commented):
Translation strings are not implemented in a clean way. I think we can and should do better here.

Issue #818, opened 2022-01-01
Copy only necessary items from the code block


iamrajiv:
## Description

When I use the copy button to copy the codes then unnecessary characters are also copied like I want to copy the below code but do not want the `$` character to be copied. I want to copy only `echo "Hello World"` after clicking the copy button.

```shell
$ echo "Hello World"
```

iamrajiv:
@LisaFC

chalin:
This is only when Prism is used, correct?

I've been able to fix this in other projects by adding [`user-select: none`][us] to the prompt class(es). Might you be willing to give that a try?

[us]: https://developer.mozilla.org/en-US/docs/Web/CSS/user-select

iamrajiv:
Yeah sure @chalin. Can you send me the link to the project where you have fixed this issue? also, what does prompt class(es) mean?

chalin:
This was fixed in the Flutter website. For example, see the code blocks on this page: https://docs.flutter.dev/get-started/install/macos#get-sdk

> what does prompt class(es) mean?

I mean the CSS classes uses to style prompts in code blocks, such as the `gp` class in the case of the Flutter pages:

```html
<span class="gp">$</span>
```

chalin:
For the record, the folks at Stripe use CSS `:before` to insert prompts. For example, see https://stripe.com/docs/api/authentication:

> ![image](https://user-images.githubusercontent.com/4140793/151223885-9fd6067f-20a5-452e-bba3-0b162cd769a0.png)

That probably works best when there's a single command though.

at055612:
Prism has a command-line plugin https://prismjs.com/plugins/command-line/ that adds a prompt for you which is not user selectable.  You just need to build a new pair of `prism.{css,js}` files with the command-line plugin added in, as described [here](https://www.docsy.dev/docs/adding-content/lookandfeel/#code-highlighting-with-prism)

I have made my own shortcode for it like so
```html
<!-- 
{{/*
Short code to add command line input/output
Usage:
command-line username hostname language
username - The username to appear in the prompt, defaults to 'user'
hostname - The hostname to appear in the prompt, defaults to 'localhost'
language - The language of the content, defaults to bash

Prefix lines with '(out)' and it will render them as output lines

See https://prismjs.com/plugins/command-line/

e.g.:
{{% command-line "jdoe" "bigserver" %}}
echo "hello world"
(out) class
id
(out) uid=1000(jdoe) gid=1000(jdoe)
{{% /command-line %}}

[...]

PR #865, opened 2022-01-27
fix: Copy only necessary items from the code block

iamrajiv:
Fixes: #818


iamrajiv:
@chalin Can you check this?

PR #802, opened 2021-12-10
User guide: adding instructions on how to use and update docsy as hugo module

deining:
Instructions on how to make use of Docsy as hugo module, as submitted via #801.
More or less a complete rewrite of the chapters `Getting started` (now: [`Installation`](https://deploy-preview-802--docsydocs.netlify.app/docs/theme-installation/)) and `Updating Docsy` (now: [`Update or Conversion of Docsy Theme`](https://deploy-preview-802--docsydocs.netlify.app/docs/updating/)) from the user guide. Given links point to the preview of the chapters on Netlify. Enjoy reading!

Issue #870, opened 2022-01-31
Use NPM packages for Bootstrap and  Font Awesome


chalin:
As discussed during the 2022-01-31 PSC meeting, we'l be exploring whether we can eliminate the git submodules to Bootstrap and  Font Awesome. @geriom has done this for another project. We believe the use of submodules is historical, and docsy use would be simplified if we switched to using NPM packages -- which we're already doing to get Hugo.

LisaFC:
Would it make sense to see how we get on with the Hugo modules initiative before prioritizing implementing this - it seems if the modules approach takes off we may not need it (I'm not sure how many people absolutely won't be able to use or migrate to modules). Unless it's absolutely trivial, of course.

We use npm for our own user guide Netlify build, but it's not actually our recommended procedure for installing Hugo in our user guide, we just list it as one of several options along with eg Homebrew for Mac - the only thing we actively tell people to install with npm is postcss as I'm not sure there's any other way to do it.

chalin:
Use of **Hugo modules requires use of Go**. I don't want to force all of our docsy-dependent projects to use Go. For those projects, I also want the option to **continue using docsy via a git submodule** -- because that is part of our **workflow** for many projects.

We confirmed during the PSC meeting that docsy doesn't _need_ BS and FA to be included via submodules. Submodules are a **historical artifact** that introduces **unnecessary complexity**. The best (non-Hugo-module) way to include those dependencies is via NPM packages.

Independently of the introduction of Hugo modules, eliminating the git submodules is a very good idea IMHO.

Thoughts?

deining:
> Use of **Hugo modules requires use of Go**.

No, that's not true. I just authored #877: follow the instructions given there and you will see that you can easily come up with a skeleton hugo site using docsy as module **without** go and even **without** git.

> Independently of the introduction of Hugo modules, eliminating the git submodules is a very good idea IMHO.

I fully agree.

What we could do: for each release of Docsy, we could offer not only docsy itself as .zip download, but also the required dependencies. This way, users that don't have to install `go`, they can add their dependencies via thois download instead. We could even think about adding a dependencies directory `_vendor` to our repo. Maybe this would even allow for a seamless switch from git submodules to hugo modules without any manual intervention. I'm not sure though that adding a `_vendor` directory to the repo is a wise decision. The `vendor` directory takes precedence over any donwloaded module in hugo's module cache, this might not be what everyone wants.

Thoughts?

chalin:
[...]

PR #871, opened 2022-01-31
Enable to pull in docsy as hugo module

deining:
This is my final PR that enables users to pull in docsy as hugo module.
Having learned from the experiences that early adopters of docsy as hugo module made, I came up with the following solution:

We do have two modules now:

- root module `github.com/google/docsy` this is the theme
- submodule `github.com/google/docsy/module`, this is needed only if you use docsy as hugo modules, it pulls in dependencies `bootstrap` and `Font-Awesome`.

With all dependendies in place, one can now switch back and forth between traditional git submodules

```
theme = "docsy"
```

and modern docsy hugo module, pulled in from github:

```
theme = ["github.com/google/docsy", "github.com/google/docsy/module"]
```
Both modules were tagged as `v0.2.0-pre` and `module/v0.2.0-pre` resprectively, we should remove `-pre` once we merged this PR.

The corresponding documentation can be found in #802, it was updated and now reflect this latest state of module code as submitted with this PR.

This PR hopefully brings the module venture to a good end.


LisaFC:
This looks great! @chalin @emckean @geriom want to take a look?

LisaFC:
Will also need to review and potentially reorganize #802 when we're happy to add this to the main public docs.

go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/google/docsy
+
+go 1.12
```
chalin, line 3: Is this the minimum required version of Go or a precise version constraint -- i.e., v1.12 must be used?

go.mod
```diff
@@ -0,0 +1,11 @@
...
+
+//replace github.com/google/docsy/module v0.0.0-20220129185551-e4b42eecdf1d => e4b42eecdf1d
```
chalin, line 5: Is this comment useful?

go.mod
```diff
@@ -0,0 +1,11 @@
...
+
+require (
+	github.com/FortAwesome/Font-Awesome v0.0.0-20210804190922-7d3d774145ac // indirect
+	github.com/google/docsy/module v0.2.0-pre // indirect
+	github.com/twbs/bootstrap v4.6.1+incompatible // indirect
```
chalin, line 10: What's the `+incompatible` for?

userguide/config.toml
```diff
@@ -3,8 +3,26 @@ title = "Docsy Example"
...
+#  # replacements = "github.com/google/docsy/module -> ../../docsy"
+#  [module.hugoVersion]
+#    extended = true
+#    min = "0.73.0"
+#  [[module.imports]]
+#  # pull in docsy as hugo module
+#    path = "github.com/google/docsy"
+#  [[module.imports]]
+#  # pull in workaround and dependencies
+#    path = "github.com/google/docsy/module"
```
chalin, line 25: I'd rather that there be no changes to the `userguide` directory.

The lines above should find their way into doc pages, if anywhere IMHO.

module/config.toml
```diff
@@ -0,0 +1,40 @@
+[module]
+_merge = "deep"
+# work around https://github.com/golang/go/issues/37397
+[[module.mounts]]
+  source = "assets/bootstrap/scss/_vendor"
+  target = "assets/vendor/bootstrap/scss/vendor"
+# mount user guide so that we can copy it to skeleton site
+[[module.mounts]]
+  source = "userguide"
+  target = "assets/userguide"
```
chalin, line 10: Do we ever copy the user guide to a skeleton site? Is this a documented "feature"?

Otherwise, can we drop the `userguide` mount?

config.toml
```diff
@@ -25,3 +25,9 @@ isHTML = true
 mediaType = "text/html"
 path = "_print"
 permalinkable = false
+
+[module]
+_merge = "deep"
+[module.hugoVersion]
+extended = true
+min = "0.73.0"
```
chalin, line 33: Ok, I guess that we'll be maintaining a min Hugo version afte rall 🤷

chalin (commented):
Thanks for your efforts!
See inline comments and questions.

go.mod
deining: This is the minimum required version of Go. Higher versions of Go are fine.

go.mod
deining: No, that's an artifact and should be removed. I will  remove the line shortly.

chalin:
Thanks again for all your efforts @deining. I have a few questions -- see below.

1. Is it correct to say that the Hugo module `github.com/google/docsy/module` contains docsy's dependencies, and only the dependencies (and the file workaround -- I like how we can localize the workaround)? If so, I think that a better name could be `github.com/google/docsy/dependencies` don't you think? Regardless, I think that to have a module named `module` doesn't feel right to me :) ... but that's a small detail. 

2. I don't know about Hugo modules or whether this is a good configuration. I know that you've been thinking about and experimenting with this for a while, but have you asked on the Hugo mailing list if this Hugo module config makes sense -- i.e., separating out a project's dependencies?

3. My guess is that, some time in the future, if we decide that we no longer need to support docsy use via git submodules, then we'll be able to eliminate the dependencies module so that users can specify only `theme = ["github.com/google/docsy"]`, is that correct?

4. While this work in any way be impacted by #870 if/when it gets resolved? (/cc @geriom)

5. What are your thoughts on the relative ease (or not) of keeping Bootstrap and Fontawesome versions in sync across supported methods of using docsy (and the explicit assets imports in the files)? I'm asking because this PR introduces an extra set of dependencies to keep in sync with the others. (Maybe #870 will make it a bit easier.)

That's it for now. Thanks for entertaining my questions.

go.mod
deining: This is automatically generated when issueing `hugo mod get github.com/twbs/bootstrap`.
`+incompatible` means the dependency (=bootstrap) is not a Go module (it doesn't have go.mod in its source code). This doesn't prevent us from pulling it in as dependency, fortunately.

module/config.toml
deining: > Do we ever copy the user guide to a skeleton site?

No. But I saw that other hugo themes are doing this and I liked it.

> Is this a documented "feature"?

No not yet. But I have some ideas how to make installation process easier. I will start a discussion on this shortly. If we agree upon that copying over the user guide is a good thing to do, I'm willing to document this.

> Otherwise, can we drop the `userguide` mount?

We can drop it and readd it later.

deining:
> Thanks again for all your efforts @deining.

You are welcome.

> 1. Is it correct to say that the Hugo module `github.com/google/docsy/module` contains docsy's dependencies, and only the dependencies (and the file workaround -- I like how we can localize the workaround)?

Yes.

> If so, I think that a better name could be `github.com/google/docsy/dependencies` don't you think? Regardless, I think that to have a module named `module` doesn't feel right to me :) ... but that's a small detail.

I agree that `github.com/google/docsy/module` wasn't the best chocie. Let me reflect on this a bit.
 
> 2. I don't know about Hugo modules or whether this is a good configuration. I know that you've been thinking about and experimenting with this for a while, but have you asked on the Hugo mailing list if this Hugo module config makes sense -- i.e., separating out a project's dependencies?

Yes, I asked on the hugo discourse list, but I that didn't help me much. I agree that it's a bit akward to separate out dependencies into a separate module. But I really think is quite clean and I like it much more than the `--config` or `--configDir` route. I would like to keep and to go with this approach. 

> 3. My guess is that, some time in the future, if we decide that we no longer need to support docsy use via git submodules, then we'll be able to eliminate the dependencies module so that users can specify only `theme = ["github.com/google/docsy"]`, is that correct?

Yes, that's correct. Then we can merge both modules into one and we are set.

> 4. While this work in any way be impacted by [Use NPM packages for Bootstrap and  Font Awesome #870](https://github.com/google/docsy/issues/870) if/when it gets resolved? (/cc @geriom)

As long as we don't change the directory structure inside `assets/vendor`, I don't expect problems.
But that certainly depends on how #870 will be implemented. Once #870 will be filled with implementation details, I will have a look and come up with a more informed answer to your question.
 
> 5. What are your thoughts on the relative ease (or not) of keeping Bootstrap and Fontawesome versions in sync across supported methods of using docsy (and the explicit assets imports in the files)? I'm asking because this PR introduces an extra set of dependencies to keep in sync with the others. (Maybe [Use NPM packages for Bootstrap and  Font Awesome #870](https://github.com/google/docsy/issues/870) will make it a bit easier.)

Frankly spoken, I was a bit surprised to see that NPM packages will be implemented to replace git submodules. As pointed out by you, we then have several competing approaches and keeping them in sync may be a challenge indeed. I'm not against `npm` at all, it's popular for a good reason. But for me, hugo modules are the way to go, that's what `hugo` team is pushing forward. And docsy is a hugo theme, so it might be a good decision to follow this given path. As pointed out earlier, using the `_vendor` directory, we can come up with a go-less (and if wanted even a go- and git-less installation). Then installation is as easy as:

```
wget https://github.com/google/docsy/archive/refs/tags/v0.1.0-go-git-less.zip
unzip v0.1.0-go-git-less.zip
cd docsy
hugo serve
```

or

```
git clone -b skeleton https://github.com/google/docsy/quickstarter.git
cd docsy
hugo serve
```

No (extra) npm installation needed in this case. That's my vision, and time permitting, I will push this forward.

module/config.toml
LisaFC: So I think we didn't consider automating pushing the user guide to a skeleton site because we already have docsy-example (we do suggest users can manually copy it if the structure suits them better than docsy-example - it's smaller, for one). Might be one to ask the community if it'd be useful or just kind of redundant? 

(we *could* consider making the user guide the example site but then you don't get all the various boilerplate bits of text/structure we've put into it)

chalin: Right. I'd rather we continue to use the `docsy-example` as a starter/base instead of repurposing the user guide.

chalin:
Thanks for answering my questions @deining. I'm looking forward to the updates to this PR, after which I believe that it'll be good to go.

As for dropping git submodules for BS & FA in favor of NPM packages, see my comments in https://github.com/google/docsy/issues/870#issuecomment-1029136405.

go.mod
deining: Removed in e444964a013e.

userguide/config.toml
deining: Lines 10-25 were removed in e444964a013e.

module/config.toml
deining: Mount was removed in e444964a013e.

deining:
> > 1. Is it correct to say that the Hugo module `github.com/google/docsy/module` contains docsy's dependencies, and only the dependencies (and the file workaround -- I like how we can localize the workaround)?
> 
> Yes.
> 
> > If so, I think that a better name could be `github.com/google/docsy/dependencies` don't you think? Regardless, I think that to have a module named `module` doesn't feel right to me :) ... but that's a small detail.
> 
> I agree that `github.com/google/docsy/module` wasn't the best choice. Let me reflect on this a bit.

Changed to `github.com/google/docsy/dependencies` in e444964a013e, as requested.

userguide/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/google/docsy
+
+go 1.12
```
chalin, line 3: Why is this needed here?

userguide/config.toml
```diff
@@ -1,21 +1,22 @@
...
 
+enableRobotsTXT = true
+
+# set docsy as theme, either locally or as hugo module
+theme = ["docsy"]
+# theme = ["github.com/google/docsy", "github.com/google/docsy/dependencies"]
+
+# Will give values to .Lastmod etc.
+enableGitInfo = true
+
```
chalin, line 19: If we decide to keep this change, then I have one nit: please move this block of lines back to where it was, that is, starting in line 4 -- this will help keep blame history clean.

userguide/config.toml
```diff
@@ -1,21 +1,22 @@
...
 # Language settings
 contentDir = "content/en"
 defaultContentLanguage = "en"
 defaultContentLanguageInSubdir = false
 # Useful when translating.
 enableMissingTranslationPlaceholders = true
 
+enableRobotsTXT = true
+
+# set docsy as theme, either locally or as hugo module
```
chalin, line 13: If we decide to keep this change, then maybe:

```toml
theme = ["docsy"] # Use this repo as the theme
# theme = ["github.com/google/docsy", "github.com/google/docsy/dependencies"] # Use docsy Hugo modules
```

deining:
> Thanks for the updates @deining, we're almost there! See inline questions and comments.
> 
> As I [mentioned earlier](https://github.com/google/docsy/pull/871#discussion_r798024391), I'd rather that there be no changes to the `userguide` directory.

I thought about this and come up with a different solution that I would like to propose:

Why don't we turn the userguide into a hugo module the go-less way? I implemented this in a new branch `userguide-module-goless`, ahead of branch `feature-hugo-module` by one single commit. With this branch/commit in place, a preview of the userguide is now as easy as

```
git clone -b userguide-module-goless https;//github.com/google/docsy
cd docsy
hugo serve
```

No `--themesDir= ../..` needed, no `go`or `npm` needed, couldn't be any easier. This way we are following the new module route that hugo maintainers dedicate themselves to and we do have a (more or less) seamless transition for existing users.
Apart from that, this apporach has another advantage: we now have the directory `userguide/_vendor` inside our repo, and we can easily reuse it by simply copying it over to the site root of new, go-less installations of any kind (moght be skeleton, based on example site or even on user guide). If we agree to take that approach over, I will go through the module user docs (#802) again and will check how well this works out. If it works out, I will add an additional chapter explaining how to setup docsy using hugo modules in a go-less way (thought about this for a while already).
In general, I like the idea of having the `_vendor` directory inside the repo, I think this is superior to distributing the `_vendor` directory as `.zip` or whatever archive with each new release.

> In particular, why is there a `go.mod` file specifying that the userguide is a part of the `github.com/google/docsy` module? (As for the comment in the config file, I can understand why we _might_ want to keep it there.)

If we stick to this proposed module route, I think it's natural to mark the userguide itself as module. Apart from that,  if we go this route, I would prefer to use the modern `[module]` notation inside `userguide/config.toml` rather than sticking to the traditional `theme = "..."` approach.

config.toml
```diff
@@ -25,3 +25,9 @@ isHTML = true
 mediaType = "text/html"
 path = "_print"
 permalinkable = false
+
+[module]
+_merge = "deep"
```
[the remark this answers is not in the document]
deining: Seems like `_merge = "deep` is not needed, so if removed this line (branch `userguide-module-goless`).

userguide/go.mod
deining: If me follow my proposal given in my [comment](https://github.com/google/docsy/pull/871#issuecomment-1036220370) below, I think it's natural to turn the user guide into a hugo module.

userguide/config.toml
deining: Changing the order of blocks here was done on purpose, and I would like to keep it. See my [PR](https://github.com/google/docsy-example/pull/162) submitted to the `docsy-example`project for an explanation.

userguide/config.toml
deining: If me follow my proposal given in my [comment](https://github.com/google/docsy/pull/871#issuecomment-1036220370) below, then we should have the `[module]` declaration only, superseding the traditional `theme = "..."` approach.

LisaFC:
I think for now I'd prefer to leave the userguide as-is - its current setup works, it works with our continuous deployment/previews from Netlify, and I think the likelihood of anyone wanting to just pull down the user guide to edit/serve minus the rest of the theme is pretty small (generally community members who edit the docs do so along with a theme update). What are the advantages of making it into a module, other than "it is the newest Hugo way"? 

We can then look at refactoring it as a module itself once we've sorted the general "use Docsy as a Hugo module".

userguide/config.toml
chalin: Thanks for the link to the explanation.

config.toml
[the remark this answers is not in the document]
chalin: If it isn't needed, please remove the line from this branch. Thanks.

chalin (changes requested):
@deining: the PSC was in agreement that we'd like to bring in the docsy-as-hugo-module feature without any changes to the userguide -- please file a separate issue for the config rile refactoring. Can you make the remaining changes that were requested for this PR?

config.toml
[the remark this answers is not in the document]
deining: Removed this line from branch `feature-hugo-module` too, as requested.

userguide/go.mod
deining: Removed file `userguide/go.mod`, as requested.

userguide/config.toml
deining: Put in the theme assignment as found in HEAD of repo (`master` branch), as requested.

chalin (approved):

chalin:
@LisaFC - this has been merged, so (re-)work of the docs can being :), thanks!
@deining - great work! It's finally merged! 🎉

LisaFC:
Fantastic - I have some time this week and will revise the docs!!

PR #892, opened 2022-02-14
README: markdown cleanup, line-wrapping, etc

chalin:
No significant change in content.


chalin:
Will merge now since I want to make followup changes.

PR #896, opened 2022-02-18
Upgrade to Hugo 0.92.2 and add NPM package name, etc

chalin:
- No change in the generated site files except for the Hugo version.


chalin:
@deining - FYI, I'm doing this as a separate PR (as opposed to being part of the Hugo module PR) to ensure that there are no changes to the generated site files.

PR #898, opened 2022-02-18
User guide config: use mounts rather than the old contentDir param

### LINKED ISSUE #897: Config: clean up language settings
- Drop settings that are unnecessarily explicitly setting a parameter to an already default value
- Drop `contentDir` in favor of using mounts.

chalin:
- Closes #897
- There are no changes to the generated site files:
  ```console
  $ npm run build && gb
  ...
  $ (cd public && git diff)
  $
  ```

/cc @LisaFC @geriom @emckean

### DIFF
--- a/userguide/config.toml
+++ b/userguide/config.toml
@@ -1,13 +1,6 @@
 baseURL = "/"
 title = "Docsy Example"
 
-# Language settings
-contentDir = "content/en"
-defaultContentLanguage = "en"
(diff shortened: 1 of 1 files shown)


deining:
> @deining - can you confirm that this doesn't break anything for the Hugo modules support?

Fine for me: after applying your changes, I still could run successfully the userguide using docsy as hugo module.

chalin:
Thanks for checking @deining!

PR #904, opened 2022-02-22
Add support for Bootstrap and FontAwesome via NPM packages

### LINKED ISSUE #870: Use NPM packages for Bootstrap and  Font Awesome

chalin:
- This is a version of #889 that doesn't remove the git submodules, and hence isn't a breaking change. 
- Closes #870

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -30,3 +30,21 @@ permalinkable = false
 [module.hugoVersion]
 extended = true
 min = "0.73.0"
+  [[module.mounts]]
+    source = 'assets'
+    target = 'assets'
+  [[module.mounts]]
+    source = 'node_modules/bootstrap'
+    target = 'assets/vendor/bootstrap'
+  [[module.mounts]]
+    source = 'node_modules/@fortawesome/fontawesome-free'
+    target = 'assets/vendor/Font-Awesome'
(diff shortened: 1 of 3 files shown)


chalin:
I've tested this on one of our projects, and I'm able to use this PR w/o it being a breaking change.

Projects can continue to use docsy via a git module with or without recursive submodule dependencies.

To use docsy dependencies via NPM packages one needs to:

 1. Clean out `themes/docs` by running, e.g., `rm -Rf themes/docsy`.
 2. Fetch the docsy theme as a submodule again but without the `--recursive` flag.
 3. Fetch docsy's NPM dependencies by running `cd themes/docsy && npm install`.

From this point on, build the project as you did before.

chalin:
The production build is now using BS + FA via NPM packages:

> image

For details, see the log: https://app.netlify.com/sites/docsydocs/deploys/6214f1b914a72c0007d91062#L105

PR #905, opened 2022-02-24
Fix: shortcode 'lead' produces invalid HTML (#527)

### LINKED ISSUE #527: Shortcode 'lead' produces invalid HTML
Related, but not similar to #526:

https://github.com/google/docsy/blob/f03cba057410f5f94afb5eb21059c3874373e0c4/layouts/shortcodes/blocks/lead.html#L6-L16

This shortcode is used within `https://www.docsy.dev` and produces the HTML output below:

```
<a id="td-block-1" class="td-offset-anchor"></a>
<section class="row td-box td-box--primary position-relative td-box--gradient td-box--height-auto">
	<div class="container text-center td-arrow-down">
		<span class="h4 mb-0">
				<p>Docsy is a theme for the Hugo static site generator that’s specifically designed for technical documentation sets. Our aim is to help you get a working documentation site up and running as easily as possible, so you can concentrate on creating great content for your users.</p>
		</span>
	</div>
</section>
```

The problem with this code is the fact that the `span` tag is an inline element and therefore cannot held `<p>` tags. The above HTML code is therefore invalid.

**Possible solution:**

I would propose to replace the `<span>` tag with `<div>` tag which is a block-level element and can therefore held `<p>` tags. If desired, I can author a PR for this issue.

deining:
This PR fixes #527.

### DIFF
--- a/layouts/shortcodes/blocks/lead.html
+++ b/layouts/shortcodes/blocks/lead.html
@@ -5,12 +5,12 @@
 <a id="td-block-{{ .Ordinal }}" class="td-offset-anchor"></a>
 <section class="row td-box td-box--{{ $col_id }} position-relative td-box--gradient td-box--height-{{ $height }}">
 	<div class="container text-center td-arrow-down">
-		<span class="h4 mb-0">
+		<div class="h4 mb-0">
 			{{ if eq .Page.File.Ext "md" }}
 				{{ .Inner | markdownify }}
 			{{ else }}
 				{{ .Inner | htmlUnescape | safeHTML }}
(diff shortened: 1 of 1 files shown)


deining:
Netlify preview fails at a very early stage, for whatever reason.

chalin (approved):
LGTM, thanks.

(I'm sure that I saw a fix like this before, but I can't track it down.)

chalin:
Thanks for restarting the build @LisaFC, and thanks for the fix @deining. FYI, I've created followup/related issues:

- #906 
- #907

Issue #901, opened 2022-02-20
Latest netlify deploy has new warning message "calling IsSet with unsupported type"


deining:
With latest commit 66818eb0905c014f342432c5fc918fc4b1c04b57 in place, I now see a warning when serving the user guide.

```
8:27:49 PM: > hugo --cleanDestinationDir --themesDir ../.. "--minify"
8:27:49 PM: Start building sites …
8:27:49 PM: hugo v0.92.2-CDF6A0D6+extended linux/amd64 BuildDate=2022-02-11T14:17:39Z VendorInfo=gohugoio
8:27:50 PM: WARNING: calling IsSet with unsupported type "invalid" (<nil>) will always return false.
```

This warning can be seen also in the [netlify log](https://app.netlify.com/sites/docsydocs/deploys/620ff2f447456d000765ce21) for this commit (line 95).

@raum51: is this warning related to the changes you submitted with #895?

raum51:
I'm afraid, that its related to my changes from #895, because I've build in isset on 2 places. Sorry.
On my test system there was no problems.

I'll take a look.

raum51:
But the issue with isset rises only a warning and in the next line of the log it shows an error:

8:27:53 PM: Error: Error building site: POSTCSS: failed to transform "scss/main.css" (text/css). Check your PostCSS installation; install with "npm install postcss-cli". See https://gohugo.io/hugo-pipes/postcss/: this feature is not available in your current Hugo version, see https://goo.gl/YMrWcn for more information

Look more like a problem with SCSS/CSS, I think.

deining:
> But the issue with isset rises only a warning and in the next line of the log it shows an error:

These are 2 different, unrelated issues:

- a warning, which doesn't prevent successful deploy of the user guide. We should get rid of this warning, though.
- a failure, adressed in #900.

For me, these are separate, independent issues, that's why a raised two separate issues (#900 / #901).
Thanks for investigating the warning message adressed with this ticket.

chalin:
The build failure was my bad, and has been fixed. Thanks for reporting those @deining.

raum51:
Thanks @chalin for the info.

So also the warning caused by my PR #895 is solved or do I have to check it?

chalin:
> So also the warning caused by my PR #895 is solved or do I have to check it?

The warning is still there, for example, see the [latest production log](https://app.netlify.com/sites/docsydocs/deploys/62140dcbd44e830008db6061#L119).

[...]

PR #895, opened 2022-02-18
Taxonomy bugfix

raum51:
As reported in #888 the option to remove the taxonomy clouds in the sidebar didn't work anymore.
This PR is a bugfix for issue #888 and fixes also the same problem with the option to remove the taxonomy infos on each article `<header>` holding the meta infos of the current article.

My old github username was @narrenfrei. I'm sorry, but I had to change it.

PR #909, opened 2022-02-28
Fix taxonomy cloud and article terms

raum51:
This should fix a bug of #895 as discussed in #901.


chalin (approved):
This works for me, thanks!

PR #920, opened 2022-03-08
Lock npm dependencies with package-lock.json

candrews:
The benefit is that all developers/users will use the same dependency tree.

Problems where one user can reproduce a problem but another cannot due
to using two different versions of a dependency are eliminated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 node_modules/
-package-lock.json
 
 # Hugo
 .hugo_build.lock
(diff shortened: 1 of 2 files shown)


jalvarez-raft:
this would be great to have...

geriom:
We discussed this during the Docsy PSC meeting. The described situation where users can't run the site because of version drift hasn't been a common occurrence.

I think the benefits of including this file don't offset the extra work to keep it up-to-date and the potential merging conflicts that often happen when the lock file is checked in.

chalin:
Closing for the reasons given by @geriom above.

Issue #907, opened 2022-02-24
Clean up invalid HTML


chalin:
For reported cases, see

- #526
- #527

But there are more similar instances of these (e.g., invalid use of `<span>`) as a quick regex search can confirm. I'm opening this issue as a reminder to track down similar issues found in other files.

PR #924, opened 2022-03-09
Fixing invalid html

deining:
This PR deals with occurrences of improper html as addressed in #907. It fixes two errors reported from [W3C validation service](https://html5.validator.nu/?doc=https%3A%2F%2Fwww.docsy.dev%2Fdocs%2F) when checking the user guide:

- The “aria-controls” attribute must point to an element in the same document.
- The `itemprop` attribute was specified, but the element is not a property of any item.


geriom (approved):
LGTM. Thank you!

Issue #960, opened 2022-04-06
Link to Getting-started page gives 404


chalin:
It seems that #802 rendered the following link invalid (which is referred to by the README, among others: https://docsy.dev/docs/getting-started/

The new link is https://www.docsy.dev/docs/theme-installation/, but it still has a ** Get started** title.

There are a number of ways to fix this, but I suggest:

- Renaming the file so that the following URL is valid: https://docsy.dev/docs/get-started/
- In that file, declare an alias for https://docsy.dev/docs/getting-started/

/cc @LisaFC @deining

deining:
> There are a number of ways to fix this, but I suggest:
> 
>     * Renaming the file so that the following URL is valid: https://docsy.dev/docs/get-started/
> 
>     * In that file, declare an alias for https://docsy.dev/docs/getting-started/
`
Agreed, implemented as suggested in PR #963.

PR #961, opened 2022-04-06
NPM package: change version to 0.2.0

chalin:
This also contains initial edits to the README.


chalin:
FYI, more changes are required for the README, but we'll need to fix #960 first.

PR #970, opened 2022-04-09
fix docu TL;DR: second module import path

Marzal:
path = "github.com/google/docsy" is duplicated, in 'Detailed conversion instructions' the right path is '"github.com/google/docsy/dependencies"'


deining:
Your PR is valid, thanks for spotting this!
The same change should be applied in line 42 (Windows instructions).
Can you please add this change to your PR as well? Thanks!
Afterward this PR will be ready for merging.

PR #972, opened 2022-04-09
User guide, package.json: add name and version

deining:
From the [npm docs](https://docs.npmjs.com/creating-a-package-json-file):

```
A package.json file must contain "name" and "version" fields.
```

`package.json` inside the `userguide` directory does not have this fields.

As a consequence users may be confronted with poor output like this:

```
/path/to/docsy/userguide$ npm run-script
Scripts available in undefined via `npm run-script`:
  _build
  ...
```

This PR corrects this.

### DIFF
--- a/userguide/package.json
+++ b/userguide/package.json
@@ -1,4 +1,6 @@
 {
+  "name": "docsy-user-guide",
+  "version": "0.2.0",
   "scripts": {
     "_build": "npm run _hugo-dev",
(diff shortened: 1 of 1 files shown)


chalin (approved):
Interesting, I had not noticed that behavior of "undefined" being printed when NPM lists the scripts.

I doubt that we need a version since the UG isn't a published package, but I'm ok with this PR. (We might consider removing the version field later.)

Issue #992, opened 2022-04-27
Example site, Go commands don't appear to be using proxy


jasonday:
Running `hugo server` and getting the following errors:

```
hugo: downloading modules …
go: github.com/FortAwesome/Font-Awesome@v0.0.0-20210804190922-7d3d774145ac: invalid version: git ls-remote -q origin in /var/folders/56/cm8gd_sj6vx5wwwl3x3wyc_cmskj1x/T/hugo_cache/modules/filecache/modules/pkg/mod/cache/vcs/a81254f0cf90f611158215d1cb50586eccc323b54ab825bb7e9c8b2580ac9fb3: exit status 128:
	fatal: unable to access 'https://github.com/FortAwesome/Font-Awesome/': Failed to connect to github.com port 443: Operation timed out
go: github.com/google/docsy@v0.2.0: invalid version: git ls-remote -q origin in /var/folders/56/cm8gd_sj6vx5wwwl3x3wyc_cmskj1x/T/hugo_cache/modules/filecache/modules/pkg/mod/cache/vcs/98f59a39bbbc2d4de9185a32a3d8790111161ae9778e3e5e32a90c95f026ac9b: exit status 128:
	fatal: unable to access 'https://github.com/google/docsy/': Failed to connect to github.com port 443: Operation timed out
go: github.com/twbs/bootstrap@v4.6.1+incompatible: invalid version: git ls-remote -q origin in /var/folders/56/cm8gd_sj6vx5wwwl3x3wyc_cmskj1x/T/hugo_cache/modules/filecache/modules/pkg/mod/cache/vcs/452026f8ece4b6e04adce745be52bb14c2f2b65baffde3e5ebfe0112a044536d: exit status 128:
	fatal: unable to access 'https://github.com/twbs/bootstrap/': Failed to connect to github.com port 443: Operation timed out
hugo: collected modules in 106159 ms
Error: failed to download modules: failed to execute 'go [mod download]': failed to execute binary "go" with args [mod download]: go: github.com/FortAwesome/Font-Awesome@v0.0.0-20210804190922-7d3d774145ac: invalid version: git ls-remote -q origin in /var/folders/56/cm8gd_sj6vx5wwwl3x3wyc_cmskj1x/T/hugo_cache/modules/filecache/modules/pkg/mod/cache/vcs/a81254f0cf90f611158215d1cb50586eccc323b54ab825bb7e9c8b2580ac9fb3: exit status 128:
[...]

PR #994, opened 2022-04-28
User guide: module definition: add 'proxy' attribute

### LINKED ISSUE #993: User guide, module installation: add instructions how to deal with a proxy server
See #992 for details.

deining:
#992 addressed how to deal with proxies, so it's a good thing to extend the documentation in order to alleviate the use of a proxy.
This PR closes #993.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,7 +6,7 @@ publish = "userguide/public"
 command = "npm run docs-install && npm run build:preview"
 
 [build.environment]
-GO_VERSION = "1.18"
+GO_VERSION = "1.18.2"
 HUGO_THEME = "repo"
 
 [context.production]
--- a/package.json
+++ b/package.json
@@ -17,6 +17,6 @@
     "bootstrap": "^4.6.1"
   },
   "devDependencies": {
-    "hugo-extended": "0.96.0"
+    "hugo-extended": "0.98.0"
   }
 }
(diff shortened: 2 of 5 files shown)


LisaFC:
This is great! Should we add something in the "Add theme module configuration..." intro section linking to the Hugo docs so users will know if/when they might need to change things? Something like:

You can find details of what these configuration settings do in the [Hugo modules documentation](https://gohugo.io/hugo-modules/configuration/#module-config-top-level). Depending on your environment you may need to tweak them slightly, for example by adding a `proxy` to use when downloading remote modules.

Also is the proxy issue likely to show up for users who have copied the example site instead, in which case we should probably put a note in those docs too.

I can update it myself in this PR or a subsequent one?

deining:
> This is great!

Glad you like it!

> Should we add something in the "Add theme module configuration..." intro section linking to the Hugo docs so users will know if/when they might need to change things? Something like:
> 
> You can find details of what these configuration settings do in the [Hugo modules documentation](https://gohugo.io/hugo-modules/configuration/#module-config-top-level). Depending on your environment you may need to tweak them slightly, for example by adding a `proxy` to use when downloading remote modules.

Yes, good idea. Wording sounds clear to me.

> Also is the proxy issue likely to show up for users who have copied the example site instead, in which case we should probably put a note in those docs too.

Yes. And we should add `proxy = "direct"` to the `config.toml` used by the example site. I will author a PR for this shortly.

> I can update it myself in this PR or a subsequent one?

Yes, go ahead. I'm glad if you update my exisiting PR with your useful suggestions. Looking forward to get that merged!

LisaFC:
Excellent!

PR #996, opened 2022-04-29
[offline search configuration] Specify attributes for offline search belongs to the params key 

ArtieReus:
This PR strictly specifies that attributes in the code snippets belongs to the params key. There is no other notice in text that clarifies this.


ArtieReus:
Any comments / actions on this?

LisaFC:
That's really helpful, thanks - should make it clearer where these settings need to go in the config file.

PR #1000, opened 2022-05-05
feat: support gitlab page meta links

### LINKED ISSUE #375: Add integration with GitLab
Currently Docsy has very good integration with GitHub including but not being limited to "edit this page" and "create child page" links.

Would it be possible to implement the same for GitLab? Or at least a subset of the features should there be any restrictions on GitLab side. Something similar has already been done in the [Vuepress](https://vuepress.vuejs.org/) project.

valdisrigdon:
Fixes #375

An alternative implementation here would be to deprecate the existing `github_` fields, rename them to `git_` instead of `github_` and add a 5th field of `git_repo_type` where `github` or `gitlab` is supported.

I chose to avoid deprecation and add the new fields, but I'm open to implementation suggestions.

### DIFF
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -1,47 +1,66 @@
 {{ if .File }}
 {{ $pathFormatted := replace .File.Path "\\" "/" -}}
-{{ $gh_repo := ($.Param "github_repo") -}}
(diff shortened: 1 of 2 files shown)


valdisrigdon:
@chalin ready for a review

chalin:
- Superseded by the repo-links unification tracked in #1841. Thanks.

Issue #605, opened 2021-06-18
Disable CDN resources


fnetX:
Downloading resources from CDNs (esp. those who are behind Cloudflare) is questionable from a users' privacy perspective. Technically, they even have to be mentioned in the privacy statement as of German law here.

What do you think about adding a config option for not relying on external providers. When you don't want to update all dependencies in your repository, I'd even be fine manually overriding the URLs and updating them manually.

It'd be nice to hear your opinion and if this project is suitable for people caring about privacy at all, or if we should rather avoid it.

LisaFC:
I'm not sure how this would work as a config option, but you can always create your own copies of the relevant layouts and just put in your own URLs.

@gwatts, you mentioned this in #432, interested to hear your thoughts.

Natureshadow:
Oh great. I just migrated my site to this theme, only to find it to load an incredible amount of external resources, completely ignoring any privacy regulations.

I should have known when seeing that it is made by Google…

fnetX:
For everyone landing here, I just want to recommend having a look at https://squidfunk.github.io/mkdocs-material/ - we went with it after docsy wasn't what we were looking for. It's awesome, somewhat lightweight compared to the amount of features, and the community is nice. :wink:

heinrich-ulbricht:
As heads-up: this topic currently gets some traction in Germany. People are getting fined for pulling in fonts from Google servers because this gives their IP address (which seems to be [PII](https://en.wikipedia.org/wiki/Personal_data)) to Google, without user consent. -> https://www.golem.de/news/google-fonts-abmahnungen-an-webseitenbetreiber-mit-google-schriftarten-2208-167472.html

chalin:
Hi @fnetX! IMHO, this is a very reasonable request, and quite doable (since Hugo v0.90+). Sorry to see you go. Thanks for the link to [Material for MkDocs], the resulting sites look amazing! I'm hoping that we can get inspiration from them.

[Material for MkDocs]: https://squidfunk.github.io/mkdocs-material/

sftim:
Hugo allows fetching resources at build time, so I think Docsy could actually have a configurable option that allows either:

- fetch from CDN at build time, validate hash, then place fetched resource into `static` folder
- reference from CDN

chalin:
Thanks for the feedback @sftim. Could you upvote the issue to help increase its priority?

PR #1006, opened 2022-05-08
Make asset urls configurable

magicbrothers:
Allows configuration (in `config.toml` and `assets/scss/_variables_project.scss`) of own URLs instead of public CDNs.

advances #605 

One font is still loaded from jsdelivr.net, but I can't find it in the code.

### DIFF
--- a/assets/scss/_variables.scss
+++ b/assets/scss/_variables.scss
@@ -78,6 +78,10 @@ $font-family-monospace: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mon
 $font-family-base: $font-family-sans-serif !default;
 $font-size-base: 1rem !default;
 
+$font_vazir: "https://cdn.jsdelivr.net/gh/rastikerdar/vazir-font@v27.0.1/dist/font-face.css";
+$font_rubik: "https://fonts.googleapis.com/css2?family=Rubik:wght@300;400;500;600;700&display=swap";
+$font_tajawal: "https://fonts.googleapis.com/css2?family=Tajawal:wght@300;400;500;700&display=swap";
+
 // Font weights
 
 $font-weight-light: 300 !default;
--- a/assets/scss/rtl/_main.scss
+++ b/assets/scss/rtl/_main.scss
@@ -31,20 +31,20 @@ body:lang(ur) {
 }
 
 body:lang(fa) {
-    @import url('https://cdn.jsdelivr.net/gh/rastikerdar/vazir-font@v27.0.1/dist/font-face.css');
+    @import url($font_vazir);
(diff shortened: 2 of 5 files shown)


magicbrothers:
How can I sign the CLA without a Google account?

emckean:
Unfortunately it's not possible to sign the Google CLA without a Google account.

magicbrothers:
Then you can just accept my contribution without a CLA. As long as you stick to the terms of the Apache License, there is no need to grant Google more permissions.

emckean:
Unfortunately our rules do not allow contributions without a signed CLA.

Natureshadow:
I can only shake my head at this. Well, ok, this is Google – so I probably should not be too surprised.

This contribution is covered by the Apache license, and this has to be enough. What Google wants here is a permission to use the contribution under other terms, which is plain unacceptable. The Apache license, in contrast to other free software licenses, even has a paragraph expressly guaranteeing that contributions are covered by it.

This change makes Docsy GDPR compliant, and apart from GDPR being a law for a major part of the world, it should be more or less self-evident that users' privacy protection matters more than some stupid CLA, especially when there is no single reason for requiring this CLA.

And, as already stated: It seems that @magicbrothers would even grant a CLA if it could be sent without accepting Google's ToS on top before.

emckean:
Thank you for your feedback! I will take up with the folks managing the CLA process.

heinrich-ulbricht:
@magicbrothers Good job and right on time, as people are getting fined now in Germany! -> https://github.com/google/docsy/issues/605#issuecomment-1210249625
I hope this CLA issue can be solved swiftly. (But I doubt it.)

Nino-Bock:
I too want to say that this topic is important for me.
Please make a website which allows a full privacy mode. Thanks.

Nino-Bock:
@magicbrothers, how do I set the link to a locally downloaded fille?
I applied your changes to my local files and downloaded the scripts to `themes/docsy/assets/js/SCRIPT`.

Putting the following into my `config.toml` does not make my files to find the locally downloaded file:
`lunr = { src = "/js/extern/lunr.min.js", integrity = "sha384-203J0SNzyqHby3iU6hzvzltrWi/M41wOP5Gu+BiJMz5nwKykbkUx8Kp7iti0Lpli" }`

This is probably some HUGO related issue I do not understand and all I need to do is to put the files into the correct folder and use the correct path

Natureshadow:
I have given up on this, due to the general foulness of the Go and Hugo ecosystems (which is demonstrated perfectly in this issue, its ignorance by the maintainers, and the refusal to let people contribute without unnecessarily subjugate themselves to Google.

For anyone looking for a a sane alternative, I recommend looking at [Zola](https://www.getzola.org/) and its [Adidoks](https://www.getzola.org/themes/adidoks/) theme.

naturedamends:
I have a Google account for now. I will copy this code and commit it under myname. I am allowed to if I reference the author (if he cares lol). Can the author not do this. I want this change now more than ever before with AI on the prowl

PR #1010, opened 2022-05-16
Completion of Bengali translation

MoshiulRabbi:
Update bn.toml

### DIFF
--- a/i18n/bn.toml
+++ b/i18n/bn.toml
@@ -16,6 +16,10 @@ other = "এই সাইটে খোঁজ করুন…"
 [ui_in]
 other = "মধ্যে"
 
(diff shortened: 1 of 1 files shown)


deining:
Thanks for your contribution! I realized that `bn.toml` is not in sync with latest `en.toml`. I just added two recently added strings, can you please translate these strings, too? Thanks.
Also, these lines are not translated yet:

https://github.com/google/docsy/blob/3641b248f2ee8ed97837d9104212f6addf6a298d/i18n/bn.toml#L59-L75

Can you please translate these lines as well? Thanks!

deining (approved):
LGTM

deining:
Thanks for your contribution!

PR #1011, opened 2022-05-17
Add breadcrumbs to taxonomy result list

### LINKED ISSUE #1004: Add breadcrumbs to taxonomy result list
When you click a taxonomy term, e.g. in the Tags cloud, you get the list of matching pages for that term.  It is possible to have multiple pages with the same name.  It would be clearer if each item had its breadcrumb displayed so the user can see where the page is.  The user also then has the option to select an ancestor of one of the matched pages.

I have done this on my site.  Happy to craft a PR for the change to `_default/taxonomy.html` if you feel it warrants being added into Docsy.  I also put each item in a card to make each item more distinct from each other.

It would need to respect the `breadcrumb_disable` config property or have its own, e.g. `taxonomy_list_breadcrumb_disable`.

The coloured tags are a local customisation so ignore that.

image

The un-modified docsy theme looks like this for comparison:

image

at055612:
Fixes #1004 

* Adds a card around each item in the taxonomy result list
* Adds a breadcrumb line to each taxonomy result
  * Makes it clear the location of the item
  * Provides the means to open an ancestor of a result item
  * Can be disabled using site param `taxonomy_breadcrumb_disable`, enabled by default
* Changed the breadcrumb so that the leaf item is not a clickable link

I think the cards could do with some drop shadow but I haven't added it as I wasn't sure it was consistent with the rest of the theme.

Making the leaf item not a link is possibly contentious but a quick google suggests this is an accepted practice.  I think having a link to the page you are on is confusing and offers no value.

This is how it now looks on the example site

image

This is how it looks for a result that is at the root level.

image

But if `.td-breadcrumbs__single { display: none !important; }` is set in the project scss then you will see:

image

The page breadcrumbs now look like this:

image


LisaFC:
This is great! @narrenfrei if you're free do you want to take a closer look at this as you did a lot of the taxonomy support? I'll take a look as well.

raum51:
Looks fine - thank you for the great work!

LisaFC:
Follow-up: can you add a quick update to the docs telling users how to disable the taxonomy breadcrumb feature if required?

at055612:
>Follow-up: can you add a quick update to the docs telling users how to disable the taxonomy breadcrumb feature if required?

Thanks for merging.  See #1048 for the docs update.

layouts/partials/breadcrumb.html
```diff
@@ -18,8 +18,11 @@
     {{ template "breadcrumbnav" (dict "p1" .p1.Site.Home "p2" .p2 )  -}}
   {{ end -}}
   {{ $isActive :=  eq .p1 .p2 }}
-  <li class="breadcrumb-item{{ if $isActive }} active{{ end }}"
-      {{- if $isActive }} aria-current="page"{{ end }}>
```
chalin, line 22: @at055612 - why did you remove `aria-current`?

at055612: Good question.  I can't remember why I took that out.  Maybe because I don't think it is applicable when the breadcrumb is shown on the taxonomy list, however I agree it ought to be there for the page breadcrumbs.  Will take a look.

chalin: For now I've reverted this change -- see #1068. We would need a bit more logic to remove (all?) the Aria related labels for the taxonomy-list usage.

PR #1016, opened 2022-05-21
Add Docker to run docsy user guide locally

geriom:
Add support to run the docsy website server locally inside a container.
With support for both `docker` and `docker-compose`.

The folder is mounted as a shared volume, allowing changes to be picked up
by the container. The `node_modules` folder is kept in a volume, to
avoid surfacing those files in the host folder.

The documentation explains how to run the container as the host user, so
you don't end up with files like `package.json` and `.hugo-build.lock`
owned by root on the host os.

[Docs preview](https://deploy-preview-1016--docsydocs.netlify.app/docs/contribution-guidelines/#preview-your-changes-using-a-docker-container)


geriom:
Tested only on Linux, please test on Mac.

LisaFC:
Works fine on my Mac!

Issue #257, opened 2020-04-21
Sidebar menu reloads to the top and user has to scroll down to find an item


andreamussap:
In a long sidebar menu, when the user wants to navigate in a doc set, which is way down in the menu structure, every time the user clicks a menu item, the entire menu is reloaded (??) to the top.
Then, the user has to scroll all the way down in the menu to find the item that they've just clicked.

https://www.loom.com/share/a259028d884d40ec96b98cc2165d6df8

It's tiring and confusing having to scroll down the menu every time that you click an item. So I'd like to ask if you could show/stuck the menu where the user is, like for example, on MS docs, no matter how long the menu is, it always shows the part of the menu where I am navigation at the moment.

See [screenshot below](https://docs.microsoft.com/en-us/azure/sql-database/sql-database-elastic-scale-upgrade-client-library?view=sql-server-ver15), although I'm way down in the menu, it didn't scrolled to the top, it's stuck, showing the part where I am navigating. If I click any item inside 'Database sharding', the menu stays where it its, I don't have to scroll to find the item that I had clicked.

![Docsy_SidebarMSexample](https://user-images.githubusercontent.com/10401080/79869316-13b51d80-83d9-11ea-90ff-12dbae43d3b4.png)

BTW, Docsy is an amazing project, guys. Thank you!!

alexearnshaw:
@LisaFC A couple of our users have reported this problem now for our Docsy site https://axway-open-docs.netlify.app/docs/

I presume it's only an issue for larger sites like ours with a long left navigation? Any suggestions what might be causing it/how we can fix?

prdsoftware:
Yep, I've noticed this also.  It's not friendly at all.  Is there any setting / fix that can be done?

Prachi-Srivastava-13:
We are also facing this issue in our internal site based off of docsy. Its pain to navigate through the pages with scroller being reset to the top. Can we have a fix for this?

narrenfrei:
I had already made a working solution (mentioned in #348) for this issue in my pipeline, but I'm afraid that this tweak got lost by some code merging problems. The only fragment of my code I can find is here: https://github.com/narrenfrei/docsy/commit/09fb1433f2d25d149e014a1fc582febf12bb01ed

I can take once more a look on this issue, but I'm afraid I won't find time in the next 2 weeks. But if anyone else wants to build a PR it's fine for me.

fekete-robert:
@huanlin 's solution worked for me.

chalin:
[...]

PR #1023, opened 2022-05-25
Scroll left toc to active entry

fekete-robert:
Useful when the left-side toc is longer than the screen.
Fixes https://github.com/google/docsy/issues/348 and https://github.com/google/docsy/issues/257

This is an update of the original patch of narrenfrei

Notes: 
- It seems that on Firefox there is a side-effect of using https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView that causes the page to scroll its main content even if it's not needed. I haven't seen this problem in Chrome and Safari
- It might be better to use https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded but that's not yet supported in every browser


fekete-robert:
Hi @LisaFC and @narrenfrei, have you had a chance to review this PR?

raum51:
I think in the netlify preview the functionality doesn't work, because the sample page ist built with the cached sidebar `sidebar_cache_limit: 10`.

Also the script that adapts the sidebar menu when the sidebar cache is use have to set the id of the span to `td-sidebar-nav-active-item`. Actually only a class get added.
https://github.com/google/docsy/blob/64883607d575921291ec09485dde77e61698d828/layouts/partials/sidebar.html#L11

fekete-robert:
Many thanks for the comments, I'll try to fix them later this week

Issue #1022, opened 2022-05-24
With `ui.sidebar_menu_foldable = true` enabled keyboard focus moves to hidden menu items


jasonday:
Current behavior:
With `ui.sidebar_menu_foldable = true` enabled, when using the keyboard only, focus is placed on submenu items that are not visible. 

Expected behavior:
Only visible elements in the menu are able to receive focus.

raum51:
I'll take a look on this issue (hopefully next week).

kimberleybrown:
Bumping this topic up to see if a fix was applied for this. We recently had an accessibility audit and this issue in particular was still being flagged.

jasonday:
yes this was merged - https://github.com/google/docsy/pull/1028

chalin:
Is this closed by #1028?

jasonday:
Yes. Verified code at line 151 - https://github.com/google/docsy/blob/main/assets/scss/_nav.scss#L151

Issue #1005, opened 2022-05-06
imgproc should follow a <figure/> model to be more semantically correct


jasonday:
Current output of `imgproc` from docsy-example [site](https://example.docsy.dev/blog/2018/10/06/easy-documentation-with-docsy/):
```
<div class="card rounded p-2 td-post-card mb-4 mt-4" style="max-width: 610px">
	<img class="card-img-top" src="/blog/2018/10/06/easy-documentation-with-docsy/featured-sunset-get_hu69849a7cdb847c2393a7b3a7f6061c86_387442_600x300_fill_catmullrom_smart1_3.png" width="600" height="300">
	
	<div class="card-body px-0 pt-2 pb-0">
		<p class="card-text">
Fetch and scale an image in the upcoming Hugo 0.43.
<small class="text-muted"><br>Photo: Riona MacNamara / CC-BY-CA</small></p>
	</div>
	
</div>
```

To be more semantically aligned and to improve accessibility, I propose using `<figure>` and `<figcaption>` for images included in this manner, which creates a programmatic association between the image and the additional text, providing more context for screen readers :

```
<figure class="card rounded p-2 td-post-card mb-4 mt-4" style="max-width: 610px">
	<img class="card-img-top" src="/blog/2018/10/06/easy-documentation-with-docsy/featured-sunset-get_hu69849a7cdb847c2393a7b3a7f6061c86_387442_600x300_fill_catmullrom_smart1_3.png" alt="embedded hugo image with obscured code snippet" width="600" height="300">
	
	<figcaption class="card-body px-0 pt-2 pb-0">
		<p class="card-text">
Fetch and scale an image in the upcoming Hugo 0.43.
<small class="text-muted"><br>Photo: Riona MacNamara / CC-BY-CA</small></p>
	</figcaption>
	
</figure>
```

LisaFC:
That's a nice idea! Is it likely to break anything (I can't see anything obvious...)?

jasonday:
@LisaFC does the imgproc shortcode need to have alt text added (or is there another way to add alt text that marries this shortcode with Hugo's approach)? Note: images either need `alt=""` to indicate that it's a decorative image or `alt="image description"` to ensure screen readers can parse the image. `figcaption`s are not a substitute for alt text. 

For example, I could see:
```
resources:
- src: "**spruce*.jpg"
  params:
    byline: "Photo: Bjørn Erik Pedersen / CC-BY-SA"
[...]

PR #1028, opened 2022-05-26
Improve accessibility of foldable sidebar nav and imgproc

jasonday:
The foldable sidebar nav is using `max-height` transitions to hide subnavs visually, but this approach does not hide visually hidden content from keyboard and screen reader users. Moving to a simple `display: [block/none]` toggle solves the issue.

See #1022


jasonday:
#1005 Addresses imgproc accessibility.

LisaFC:
This LGTM, @narrenfrei do you want to take a look as you did a lot of the navbar work?

PR #1031, opened 2022-05-30
Update feedback.md

derrygeoff:
Added instructions on how users can get feedback via contact form submission.


LisaFC:
Thanks for that! I moved the content down a bit and expanded the intro, looks good!

PR #1036, opened 2022-06-01
fix: add link to migration guide

317brian:
I'm not sure if I should be using the filepath or a URL. The same link to the migration guide on "Get started: Migration and backward compatibility" uses the URL

```
If you have an existing site that uses Docsy as a Git submodule, and you would like to update it to use Hugo Modules, follow our [migration guide](https://www.docsy.dev/docs/updating/convert-site-to-module/)
```


deining (approved):
LGTM. Using the filepath is a good decision.

deining:
Thanks for spotting and fixing this broken link!

PR #1038, opened 2022-06-03
User guide: remove content related to Blackfriday (unsupported)

deining:
As of version 0.100, hugo [dropped support](https://github.com/gohugoio/hugo/issues/9944) for Blackfriday markdown engine.
This PR makes changes to `config.toml` and the user guide to reflect this.


LisaFC:
Great work, thanks!

PR #1044, opened 2022-06-10
community_links: make community_how_to adjustable

stephanlachnit:
This changes "You can find out how to contribute to these docs in our" on the Community Page to "You can find out how to contribute to {{ .Site.Title }} in our".

This make much more sense IMHO, since the right side is about the project, not the docs of the project. See for example the Docsy page itself (https://www.docsy.dev/community/): the contribution guidelines also refer to the chapter "How to contribute to Docsy" and not just "How to contribute to Docsy's documentation".

I also adjusted the translation for all the languages I'm able to understand.


LisaFC:
This is a great call, thanks! I just added a few missing "to"s in your updates, otherwise looks great.

PR #1050, opened 2022-06-15
Loading lunr.js is not critical for page rendering and can be deferred

fekete-robert:


LisaFC:
@tekezo do you want to take this one?

tekezo:
Thank you!
It's good to defer loading `lunr.js`.

I have one point of concern.
`offline-search.js` depends `lunr.js`, it requires lunr is ready when DOMContentLoaded.
If the async attribute is specified, this preparation may not be completed in time.
So, I think it's better to remove `async` attribute.

```diff
- <script
+ <script defer
```

fekete-robert:
Many thanks for the review, I've updated the PR

tekezo:
LGTM 👍

LisaFC:
Thanks for this!

PR #1051, opened 2022-06-15
Simple HubSpot integration

fekete-robert:


chalin:
- We don't have capacity to adopt and maintain a HubSpot integration in core. Thanks.

PR #1053, opened 2022-06-15
Conditional text

fekete-robert:
The `conditional-text` shortcode allows you to show or hide parts of your content depending on the value of the `buildCondition` parameter set in your configuration file. This can be useful if you are generating different builds from the same source, for example, using a different product name. This shortcode helps you handle the minor differences between these builds.


LisaFC:
Oh this is really nice! Thanks so much for adding it (and documenting it too).

PR #1054, opened 2022-06-15
Version pagelinks

fekete-robert:
If you set the `version_menu_pagelinks` parameter to `true`, then links in the version drop-dow menu 
point to the current page in the other version, instead of the main page.
This can be useful if the document doesn't change much between the different versions.
Note that if the current page doesn't exist in the other version, the link will be broken.


LisaFC:
I wonder is there a way for this to fail more elegantly - go to the main page of the versioned site if the specific URL isn't there?

fekete-robert:
I don't think so, as there is no real way to check that the url exists - often the different versions are under different domains and different sites.

polarweasel:
This seems enormously useful even without the graceful failure mode. In our particular case, we could handle most 404s with server-side redirects.

LisaFC:
Ok, let's go with the current implementation for now, looks like people will find it useful! Thanks.

chalin:
> I wonder is there a way for this to fail more elegantly - go to the main page of the versioned site if the specific URL isn't there?

Nice feature indeed. I'd consider failing more elegantly to be essential, so I've created an issue to track this:

- #1151

/cc @nate-double-u - navigate between versions is getting closer to being supported the way we need it :)

PR #1055, opened 2022-06-15
Add missing argument

cbrownstein-lambda:
`docker build` needs as an argument the path to the Dockerfile.


geriom (approved):
Good catch. Thank you!

PR #1057, opened 2022-06-16
cards: fix max width on mobile

### LINKED ISSUE #1056: Code blocks not full width on mobile
The problem in a nutshell:

Screenshot from 2022-06-15 22-11-35

The culprit is [assets/scss/shortcodes/cards-pane.scss](https://github.com/google/docsy/blob/6769ab6e8f7d26aea222f02e86b576ee159f19c4/assets/scss/shortcodes/cards-pane.scss#L1-L5). The file hard-codes a `max-width`, which works on a desktop-like layout, but not on mobile. I think the correct fix is to change the linked lines to

```scss
.card-deck {
	@extend .td-max-width-on-larger-screens;
}
.card {
    @extend .td-max-width-on-larger-screens;
```

It looks correct on most things, but I haven't tried this with all possible card-pane configs so I'm not entirely sure. Please let me know if I should open a PR.

stephanlachnit:
Closes #1056.


LisaFC:
Looks good, thanks!

PR #1060, opened 2022-06-17
i18n: es - use var Title in community_how_to

Marzal:
Use {{ .Site.Title }} in [community_how_to] as the rest of languages


chalin (approved):
This is the ES lang counterpart to #1044. LGTM

chalin:
Thanks!

Issue #755, opened 2021-11-03
Shallow clone submodules by default


chalin:
Context:

- #269 
- https://github.com/google/docsy/pull/749#issuecomment-958990607

As I mentioned in https://github.com/google/docsy/pull/749#issuecomment-958990607, I'd vote to merge this change.

To do:

- [x] Add `shallow = true` via #269
- [ ] Update the User Guide, per https://github.com/google/docsy/pull/269#issuecomment-626995225, noting that the extra step is optional.
- [ ] Remove the `--depth 1` from the CI scripts -- assigned @chalin
- [ ] Follow through to make a similar to change to `docsy-example`, per https://github.com/google/docsy/pull/269#issuecomment-626995225

chalin:
From https://github.com/google/docsy/pull/749#issuecomment-958992545:

> For the record: on other projects, I've encountered situations where shallow cloning causes some extra headaches if you try to submit changes to submodules via the submodules themselves. I don't foresee that happening here.

There's an option for ignoring the `shallow` config setting! See [--no-recommend-shallow](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---no-recommend-shallow).

LisaFC:
So I *think* from discussions on #269 and #749 we can merge the change before updating the docs.

Do you have a preference for which of the above PRs (which do the same thing) we merge? :)

chalin:
The general practice I've seen is to merge the earliest of two PRs when they are equivalent.

johnpaulett:
I'm happy to rebase if needed, but #749 effectively already did that.  I won't be offended if my PR isn't the one to get in, as long as I get a smaller checkout, I'm happy!

LisaFC:
Merged the earlier one!

gwatts:
For anyone following along curious about what shallow cloning is, and what other cloning choices there are, I found this Github blog post to be a great read: https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

To be clear, while the above post recommends caution against shallow clone in most circumstances, i don't think they intended that to be true for circumstances like this where you're really not working directly with the cloned submodule at all and really do just want the code associated with the submodule's referenced commit.  i.e. i think shallow clones here are a good idea.

People working on updating the Docsy theme itself probably still want a full clone (or one of the other options in the above post).

chalin:
That looks like a great post! Thanks for sharing @gwatts!

chalin:
[...]

PR #1064, opened 2022-06-21
Make git submodules shallow clones to save bandwidth

fekete-robert:
and speed up builds on CI/CD systems

Unless someone is working on the underlying bootstrap or fontawesome modules,
the entire git tree is not needed


deining:
Please note PR #269 and PR #749 which are identical to this PR.
Shallow cloning was put in place with #269 already, but this change was reverted again with PR #952 due to reasons given [here](https://github.com/google/docsy/issues/755#issuecomment-1081808648).

PR #1069, opened 2022-06-24
Set NPM package version to 0.4.0-dev

chalin:


userguide/package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "docsy-user-guide",
-  "version": "0.2.0",
```
chalin, line 3: Removing the version because the UG isn't a published package, and I don't want us to have to update this as part of the release process.

Note that the issue raised in #972 is addressed by having a package name. We don't need a version.

PR #1072, opened 2022-06-24
Remove ARIA attributes from taxonomy lists

### LINKED ISSUE #1071: Remove ARIA attributes from embedded-page breadcrumbs in taxonomy lists
From https://github.com/google/docsy/pull/1011#discussion_r905886458:

> ... remove (all?) the ARIA related labels for the taxonomy-list usage ... or maybe just use a regex to strip all aria labels from breadcrumbs in taxonomy list. I'll let you follow up on that if you think that it's worth it.

/cc @at055612

chalin:
- Closes #1071

I've chosen to avoid complicating the logic of the breadcrumbs partial since it's complicated enough already. This PR implements what I proposed earlier: it strips out unnecessary classes and irrelevant ARIA labels from the "breadcrumbs" used in the taxonomy lists.

You'll notice that I've removed the `disabled` state from the last breadcrumb since it no longer makes sense to disable it (because we're not on that page anymore :)).

Preview: https://deploy-preview-1072--docsydocs.netlify.app/categories/taxonomies/

---

### Screenshots

Before:

> image

After:

> image

/cc @at055612

### DIFF
--- a/layouts/_default/taxonomy.html
+++ b/layouts/_default/taxonomy.html
@@ -10,9 +10,11 @@
         {{ $manualLink := cond (isset .Params "manuallink") .Params.manualLink ( cond (isset .Params "manuallinkrelref") (relref . .Params.manualLinkRelref) .RelPermalink) }}
         <article class="card article-teaser article-type-{{ with .Type }}{{ ( urlize . ) }}{{ end }}">
           <h3><a href="{{ $manualLink }}"{{ with .Params.manualLinkTitle }} title="{{ . }}"{{ end }}{{ with .Params.manualLinkTarget }} target="{{ . }}"{{ if eq . "_blank" }} rel="noopener"{{ end }}{{ end }}>{{- .Title -}}</a></h3>
-          {{ if not .Site.Params.ui.taxonomy_breadcrumb_disable }}
-            {{- partial "breadcrumb.html" . -}}
-          {{ end }}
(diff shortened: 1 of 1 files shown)


chalin:
@at055612 - PTAL and approve if this solution is suitable for you.

at055612:
@chalin Agree with making the leaf link enabled on the taxonomy results.
I think a comment would be useful to explain why the regex replace is there.
I did play with passing a flag into the partial to define whether it was a page breadcrumb or not and thus whether to include the aria bits, but it wasn't working for some reason.  My only worry with the regex approach is that if someone changes the implementation of the partial they may unknowingly break the the taxonomy results.

I did wander about keeping the `aria-label` on the `<nav>` but setting its value to `"breadcrumb for {{ .LinkTitle }}"`.  However I think `aria-label` should be unique and we can't guarantee that.  Setting it to something like "breadcrumb for result N" would be an option but the partial has no knowledge of its index.

LisaFC:
LGTM, though agree with @at055612 that a comment for the regex replace would be useful, especially for future users who might want to modify or override the partial.

chalin:
Added a comment and rebased. PTAL

chalin:
Rebased. Pinging @LisaFC and @geriom for a review.

chalin:
@LisaFC - just a reminder that this is ready for your review.

LisaFC (approved):

PR #1084, opened 2022-07-01
Add Node install/update instructions to prerequisites

### LINKED ISSUE #1081: Document prerequisite of the latest Node LTS
I thought that we had documented this, but apparently we don't.

chalin:
- Closes #1081

Preview: https://deploy-preview-1084--docsydocs.netlify.app/docs/get-started/other-options/#node-get-the-latest-lts-release

### DIFF
--- a/userguide/content/en/docs/Get started/other-options.md
+++ b/userguide/content/en/docs/Get started/other-options.md
@@ -64,9 +64,39 @@ npm install hugo-extended --save-dev
 
 See the [`hugo-bin` documentation](https://www.npmjs.com/package/hugo-bin) for usage details.
 
+### Node: Get the latest LTS release
+
+Install or upgrade your version of Node to the active [LTS release][]. We
+recommend using **[nvm][]** to manage your Node installation (Linux command
+shown):
+
+```console
+$ nvm install --lts
+```
+
+[lts release]: https://nodejs.org/en/about/releases/
+[nvm]:
+  https://github.com/nvm-sh/nvm/blob/master/README.md#installing-and-updating
+
+
 ### Install PostCSS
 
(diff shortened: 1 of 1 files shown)


emckean:
LGTM!

mikeBoterf:
By recommending nvm to users, you're introducing another prereq. From nodejs:

here's what I had in my MR:
### Install

> Ensure you have the [latest LTS Version](https://nodejs.org/en/about/releases/)

[Debian and Ubuntu based distributions](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions)

tl;dr

```
# Using Ubuntu
curl -fsSL https://deb.nodesource.com/setup_18.x | sudo -E bash -
sudo apt-get install -y nodejs

# Using Debian, as root
curl -fsSL https://deb.nodesource.com/setup_18.x | bash -
apt-get install -y nodejs
```

[Enterprise Linux based distributions](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions-1)

tl;dr
```
# As root
curl -fsSL https://rpm.nodesource.com/setup_18.x | bash -

# No root privileges
curl -fsSL https://rpm.nodesource.com/setup_18.x | sudo bash -
```

chalin:
Thanks for those details @mikeBoterf. You might consider sharing those details in a [discussion](https://github.com/google/docsy/discussions) so that others can more easily find them. (For now I'd rather keep such details out of the docs.)

mikeBoterf:
Cool. I’ll definitely do that next time. I agree that a discussion seems to be a more probable place

prepstarr:
does this work on computers running Berkeley Systems Distribution? :)

mikeBoterf:
> does this work on computers running variants of Berkeley Systems Distribution? :) I guess with source code? https://cgit.freebsd.org/ports/tree/www/gohugo/pkg-descr they have the port for Hugo and likely all the components. It would be good to add it to the documentation... guten nacht , auf wiedersehen.. a bientot. if we can get side tracked a little bit if I remember clearly it's BSD day in 5 days...

I’d refer you back to the docs. You can install Hugo from a tar ball.. refer to Hugo docs for that. Docsy is a theme so installing it is dependent on node and other items listed in the docs. So really you should determine if you can install the prerequisites for docsy, then things should work normally, BUT there are no guarantees.

PR #1086, opened 2022-07-01
Adding the instructions to install/upgrade Node LTS

mikeBoterf:
# [Issue](https://github.com/google/docsy/issues/1081)

## What Changed:
- added section to describe install/upgrade of Node LTS. I tried to emphasize the importance of this issue and included what may happen when the version is not up to the latest. 

Please advise if changes are necessary. Thanks!


LisaFC:
One question - are we saying you might get errors if you don't have the most recent LTS version of node, or if you don't have v16.15.1 (the current latest LTS) or higher, as these are two different things?

chalin (commented):
For the record, while this information is very useful, IMHO this is way too much detail for a Docsy guide.

PR #1087, opened 2022-07-01
User guide: fixing typos

deining:


chalin (commented):
Thanks for catching and fixing those @deining!

deining:
> Does either of you (@deining @LisaFC) remember if we have an issue open already for link checking?

Yes, we do. Please see this [issue](https://github.com/google/docsy-example/issues/161) which was raised in the docsy-example project.

Issue #1097, opened 2022-07-12
Setup Google Analytics for User Guide and Example


chalin:
The user guide config currently uses the bogus ID `UA-00000000-0`. We need a (real) analytics ID for the Docsy website. @LisaFC, @emckean can we create one? 

### Tasks

- [x] Create a community GA4 property - [G-VMZY37B6T8]
- [x] Configure the new GA4 property through [Netlify snippet injection] for:
  - [x] User guide
  - [x] Docsy example

[G-VMZY37B6T8]: https://analytics.google.com/analytics/web/#/p416098284
[Netlify snippet injection]: https://docs.netlify.com/site-deploys/post-processing/snippet-injection/

emckean:
looking into it! :)

chalin:
@LisaFC - @emckean and I were discussing this during the last PSC meeting. @emckean thought that maybe the best route would be for a non-Googler to create a new GA account, which would then become the main Docsy GA account (correct me if I'm wrong @emckean). WDYT?

chalin:
Any updates on how to move forward with this @emckean?

emckean:
I have a meeting w/someone in Analytics set for next week :)

chalin:
Any updates @emckean?

chalin:
I updated the opening comment with a list of tasks. We're all done here, so shall we close this for now @emckean?

PR #1098, opened 2022-07-12
Update Google Analytics section of UG

### LINKED ISSUE #921: Understanding config params: services.googleAnalytics.id vs. top-level googleAnalytics
There are two documentation sources on getting Google Analytics working in Hugo/Docsy
- The [Hugo google-analytics documenation](https://gohugo.io/templates/internal/#google-analytics) 
- The [Docsy google-analytics documentation](https://www.docsy.dev/docs/adding-content/feedback/)

Its unclear how these two techniques interact. They seem to use different `config.toml` variables and I found that using both at the same time created problems. There is also the possible added confusion between the concepts of general page hit tracking vs. the Docsy [params.ui.feedback] feature - can I have one or the other or both? Finally the references to the now outdated "UA-00000000-0" `tracking id` caused me some confusion as I could not generate one, as Google Analytics does not allocate these anymore (unless you burrow down into advanced settings somewhere) - instead  a ‘G-xxxxxxxxxx’ style `measurement id` is now allocated by default, and its not clear if such a measurement id can be used instead in the config. For these reasons, the Docsy documentation referred to by #92 I think needs clarification on some or all of these questions.

## Repro

To get Google Analytic page hits working, the [Hugo doco](https://gohugo.io/templates/internal/#google-analytics)  says to create and set a `googleAnalytics` variable in `config.toml` e.g.

    googleAnalytics = 'G-MEASUREMENT_ID'

This failed to work for me - it turns out it was because of the following default Docsy config in `config.toml`

```toml
[services]
[services.googleAnalytics]
# Comment out the next line to disable GA tracking. Also disables the feature described in [params.ui.feedback].
id = "UA-00000000-0"
```

Leaving the ‘id’ uncommented (as is the default) seems to prevent Hugo’s official `googleAnalytics` variable from working and thus no hits will be recorded by Google Analytics.  Or rather, the HTML that gets generated is old style `UA-xxxxxxxx` script referring to  "UA-00000000-0".  Of course "UA-00000000-0" needs to be replaced with a real UA-xxxxxx number, which may seem obvious, but since Google does not allocate these anymore (unless you burrow down into advanced settings somewhere) then a new user might decide to leave the line unchanged, thinking "UA-00000000-0" has a special meaning or something.

What ended up working for me was setting `googleAnalytics` and commenting out `id`
```toml
googleAnalytics = 'G-xxxxxxxxxxx'
# id = "UA-00000000-0"
```

## My later discoveries

It seems that using `[services.googleAnalytics] id=` overrides the plain `googleAnalytics` variable, leading me to think they are somehow two ways of setting the same thing.

It turns out you don't need  to set `googleAnalytics` variable (despite what the Hugo analytics documentation says) you can simply set `[services.googleAnalytics] id=` to either an old style `UA-xxxxxxxx` tracking code or a newer `G-xxxxxxxxxxx` measurement id.

## Documentation Suggestion

It might be helpful if the default `config.toml` relating to Google Analytics was updated to contain something like

```toml
[services]
[services.googleAnalytics]
# Comment out the 'id' line to disable GA tracking. Also disables the feature described in [params.ui.feedback].
# There is no need to set another config variable called 'googleAnalytics'
# Set id to either an old style `UA-xxxxxxxx` tracking code or a newer `G-xxxxxxxxxxx` measurement id 
# Note tracking codes are deprecated, though can still be generate it under google analytics advanced options
id = "G-xxxxxxxxx"
```

with assocated changes to the [Docsy google-analytics documentation](https://www.docsy.dev/docs/adding-content/feedback/)

chalin:
- Closes #921

Preview: https://deploy-preview-1098--docsydocs.netlify.app/docs/adding-content/feedback/#adding-analytics


userguide/config.toml
```diff
@@ -42,11 +42,6 @@ anchor = "smart"
     url = "https://example.docsy.dev"
     post = '<sup><i class="pl-1 fas fa-external-link-alt fa-xs" aria-hidden="true"></i></sup>'
 
-[services]
-[services.googleAnalytics]
-# Comment out the next line to disable GA tracking. Also disables the feature described in [params.ui.feedback].
-id = "UA-00000000-0"
```
chalin, line 48: @LisaFC @emckean @geriom - I suspect that the all-zeros ID is bogus, and that it was used only for the side-effect of having the feedback form appear on pages. So I'm removing this for now, and we can track progress via:

- #1097
- #1099

chalin: Update: I moved the ID to earlier in the file.

chalin:
FYI, I posted this question: https://discourse.gohugo.io/t/config-is-services-googleanalytics-id-an-alias-for-googleanalytics/39469. If it turns out that they are (officially) aliases, I'll add a note to that effect to the GA section.

userguide/content/en/docs/Adding content/feedback.md
```diff
@@ -57,6 +60,21 @@ googleAnalytics: PUT-YOUR-ANALYTICS-ID-HERE
...
+
+  Earlier versions of Docsy recommended setting the
+  `services.googleAnalytics.id` config parameter to enable analytics. Upgrade
+  your configuration as outlined above. **Do not** set both this parameter _and_
+  the top-level `googleAnalytics`, otherwise this will likely result in
+  [unexpected behavior][]. For more details about how these two parameters are
+  related, see [Is `services.googleAnalytics.id` an alias for
+  `googleAnalytics`][alias-discussion].
+
+  [alias-discussion]: https://discourse.gohugo.io/t/config-is-services-googleanalytics-id-an-alias-for-googleanalytics/39469?u=chalin
```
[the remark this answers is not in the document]
chalin: Oops! Done.

chalin:
All: this is ready for review. I decided to simplify this change by referring the reader to the appropriate section of the Hugo docs. The page update includes a warning (about _not_ setting _both_ `googleAnalytics` and `services.googleAnalytics.id`), but eventually (once https://github.com/gohugoio/hugoDocs/issues/1780 is addressed), we'll be able to remove the warning.

geriom (approved):

geriom:
> but eventually (once [gohugoio/hugoDocs#1780](https://github.com/gohugoio/hugoDocs/issues/1780) is addressed), we'll be able to remove the warning.

Should we open an issue to track this?

chalin:
> Should we open an issue to track this?

Done:

- #1146

Given @geriom's approval, I'm going to merge.

PR #1102, opened 2022-07-15
Upgrade to FontAwesome 6

### LINKED ISSUE #872: Font Awesome static files seem 3 years old?
Good day to whoever reads this. I was trying to load the fab fa-discord icon from Font Awesome. However I was greeted with the old discord logo. I checked which logo they supplied and since the latest version it should be the new logo. So to be sure I updated the font awesome sub module but still got the old discord logo showing up. So I dug a bit deeper and at the moment I’m thinking it’s caused by the webfonts folder in the static folder that seems to have been updated 3 years ago for the last time. I also noticed these files don’t get updated when we bump the font awesome package. So 2 questions;

1. Should I manually update the files in the webfonts folder to be able to use the latest font awesome icons?
2. Should we think of a way to automatically update these files on a version bump?

samiahmedsiddiqui:
(Edit: @chalin)

- Waiting on #1164
- Closes #872

Preview: https://deploy-preview-1102--docsydocs.netlify.app/docs/

### DIFF
--- a/assets/scss/_nav.scss
+++ b/assets/scss/_nav.scss
@@ -121,13 +121,13 @@
             font-variant: normal;
             text-rendering: auto;
             -webkit-font-smoothing: antialiased;
-            font-family: "Font Awesome 5 Free";
-            font-weight: 900; 
+            font-family: "Font Awesome 6 Free";
+            font-weight: 900;
             content: "\f0d9";
             padding-left: 0.5em;
             padding-right: 0.5em;
         }
-    }  
+    }
 }
 
 // Foldable sidebar menu
@@ -136,18 +136,18 @@ nav.foldable-nav {
     &#td-section-nav {
         position: relative;
     }
-    
+
     &#td-section-nav label {
         margin-bottom: 0;
         width: 100%;
     }
-    
+
     .td-sidebar-nav__section, .with-child ul {
         list-style: none;
         padding: 0;
         margin: 0;
     }
-    
+
     .ul-1 > li {
         padding-left: 1.5em;
     }
@@ -162,8 +162,8 @@ nav.foldable-nav {
 
     input[type=checkbox] { display: none; }
 
-    .with-child, .without-child { 
-        position: relative; 
+    .with-child, .without-child {
+        position: relative;
         padding-left: 1.5em;
     }
 
@@ -173,7 +173,7 @@ nav.foldable-nav {
         font-variant: normal;
         text-rendering: auto;
         -webkit-font-smoothing: antialiased;
-        font-family: "Font Awesome 5 Free"; font-weight: 900; content: "\f0da";
+        font-family: "Font Awesome 6 Free"; font-weight: 900; content: "\f0da";
         position: absolute;
         left: 0.1em;
         padding-left: 0.4em;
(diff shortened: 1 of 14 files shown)


samiahmedsiddiqui:
@LisaFC @narrenfrei @raum51

polarweasel:
@LisaFC and @raum51 could you have a look at this one? Fontawesome 6 would be very good to have by default, instead of us having to work around and add it manually. It looks like there's a Docsy release planned for fall, and we'd love to help get this into that.

LisaFC:
Will take a look - @chalin, can you take a look as well? Agree would be good to have the latest version for the next Docsy release. It doesn't look like it's likely to break anything but do FontAwesome themselves document any breaking changes? Don't want any users to suddenly end up with unexpected icons.

chalin:
Will take a look as soon as I can. 
Please rebase & resolve the conflicts, thanks.

samiahmedsiddiqui:
@chalin conflicts resolved.

polarweasel:
Backward compatibility information from the project: https://fontawesome.com/docs/web/setup/upgrade/whats-changed#backward-compatibility

Lots of icons got renamed, but the v5 names will still work. It'd be worth pointing to this table in the release notes, though, to make people aware that they should pick up the new names at some point.

assets/scss/_nav.scss
```diff
@@ -121,13 +121,13 @@
             font-variant: normal;
             text-rendering: auto;
             -webkit-font-smoothing: antialiased;
-            font-family: "Font Awesome 5 Free";
-            font-weight: 900; 
+            font-family: "Font Awesome 6 Free";
```
chalin, line 124: We should define a variable for the FA font name and use it here and elsewhere. This can be done in this PR or a followup PR.

layouts/partials/page-meta-links.html
```diff
@@ -37,18 +37,18 @@
   {{ $newPageQS := querify "value" $newPageStub.Content "filename" "change-me.md" | safeURL -}}
   {{ $newPageURL := printf "%s/new/%s?%s"  $gh_repo $gh_repo_path $newPageQS -}}
 
-  <a href="{{ $viewURL }}" class="td-page-meta--view" target="_blank" rel="noopener"><i class="fa fa-file-alt fa-fw"></i> {{ T "post_view_this" }}</a>
-  <a href="{{ $editURL }}" class="td-page-meta--edit" target="_blank" rel="noopener"><i class="fa fa-edit fa-fw"></i> {{ T "post_edit_this" }}</a>
-  <a href="{{ $newPageURL }}" class="td-page-meta--child" target="_blank" rel="noopener"><i class="fa fa-edit fa-fw"></i> {{ T "post_create_child_page" }}</a>
-  <a href="{{ $issuesURL }}" class="td-page-meta--issue" target="_blank" rel="noopener"><i class="fas fa-tasks fa-fw"></i> {{ T "post_create_issue" }}</a>
+  <a href="{{ $viewURL }}" class="td-page-meta--view" target="_blank" rel="noopener"><i class="fa-solid fa-file-alt fa-fw"></i> {{ T "post_view_this" }}</a>
+  <a href="{{ $editURL }}" class="td-page-meta--edit" target="_blank" rel="noopener"><i class="fa-solid fa-edit fa-fw"></i> {{ T "post_edit_this" }}</a>
```
chalin, line 41: As pointed out by @polarweasel, some icon names have changed, like: `fa-edit` --> `fa-pen-to-square`. It would be nice to update these too in this PR, but it could be in a followup PR too.

chalin (approved):
LGTM. Note that I haven't done an exhaustive search to see if any FA class names might have been missed, but what _has_ been updated here works. Thanks!

See inline comment for suggested improvements for this or a followup PR.

chalin:
Thanks @samiahmedsiddiqui for the PR. Thanks @polarweasel for the link to the backwards compatibility page.

FYI, some prep work which I'd like reviewed and merged before this PR:

- #1164

Comments are welcome over on that PR.

layouts/partials/page-meta-links.html
samiahmedsiddiqui: @chalin Names updated what I have found so far.

chalin: Great, thanks!

assets/scss/_search.scss
```diff
@@ -14,7 +14,7 @@
         border-radius: 1rem;
     }
 
-    font-family: "Font Awesome 5 Free", $font-family-base;
+    font-family: "Font Awesome 6 Free", $font-family-base;
```
chalin, line 17: This is now causing problems because it seems that the FA6 font contains regular character glyphs too:

> image

As compared to:

> image

The problem was there before this PR, so I'm ok merging this as is and fixing this problem after.

chalin (commented):
See inline comment regarding an issue with search placeholder text styling.

chalin (approved):
Very nice step forward, doing the bulk of the upgrade. Will merge now (as agreed during PSC meeting) and follow up with further changes.

assets/scss/_search.scss
chalin: Will follow up via:

- #1169

assets/scss/_nav.scss
chalin: Will follow up via:

- #1170

chalin:
I've collected the followup tasks in:

- #1171

Issue #1096, opened 2022-07-12
Support Google Analytics 4 (GA4) and migration to GA4


chalin:
**Universal Analytics (UA) is [officially deprecated]**:

- Data collection will stop on July 1st, 2023
- Six months after that, historical data will be gone too

The sooner projects migrate to Google Analytics 4 (GA4), the better.

A number of downstream projects have started migrating to GA4, including CNCF projects (https://github.com/cncf/techdocs/issues/108).

We should do what we can to help ease migration efforts.

Tasks:

- [x] #1177
- [x] #1302

[officially deprecated]: https://support.google.com/analytics/answer/11583528

chalin:
Done!

PR #1106, opened 2022-07-15
Support Universal Analytics and Google Analytics 4 IDs simultaneously

chalin:
This is a proposed solution for:

- #1096

This PRs eases migration from UA to GA4 by allowing both IDs to be specified via the `googleAnalytics` configuration parameter.

I've proposed this change to the Hugo template here:

- https://github.com/gohugoio/hugo/pull/10094

In the meantime, we might want to support a template override as given in the PR.

---

For context, we're in the midst of migrating CNCF projects. For details see:

- https://github.com/cncf/techdocs/issues/108

/cc @nate-double-u


chalin:
After much thought, I'm going to close this. I don't want to propose a Docsy feature that runs the risk of not being adopted upstream. This solution of supporting multiple IDs simultaneously, while interesting, feels like a special case of a more general need of supporting general config https://github.com/gohugoio/hugo/issues/4479.

Issue #853, opened 2022-01-20
Page TOC not showing up


juandiegopalomino:
I'm trying to use docsy for my project's documentation, but the right-hand side menu is not appearing (the one with the table of contents and "Edit this page" etc...). This looks very ugly and I don't know why it's not working.
<img width="1792" alt="Screen Shot 2022-01-19 at 5 01 16 PM" src="https://user-images.githubusercontent.com/10430635/150243315-7424d0d3-1e1d-46a2-a795-5ff4c5899160.png">

narrenfrei:
If I remember correctly without a computer to check, the Repository links have to be enabled or the content of the page must have enough headlines, otherwise there is no right sidebar built. 

So try to set more headlines in your content markdown file or/and check if the repository links are enabled.

LisaFC:
That is very weird. Agree with @narrenfrei, can you check that you're using regular Markdown/HTML headers for your section titles, as that's what the right nav is built from.

polarweasel:
It's weird indeed, and really breaking a bunch of our pages, but it's coded into Docsy's TOC generation... If the generated HTML for that TOC is less than 200 characters, the TOC doesn't get output into the page.

What is the reasoning behind the 200-character minimum?

https://github.com/google/docsy/blob/main/layouts/partials/toc.html#L3

polarweasel:
Should have added: I've fixed this locally by overriding the `toc.html` shortcode, but this behaviour should get fixed in Docsy. I lowered the minimum to 100 for now.

polarweasel:
@raum51 I wonder what the logic behind the character count is, rather than a word count (or better, heading count)?

At Yugabyte, we're doing a variety of overrides with the TOC, but I'll ask the developer who's working on our site to take a look at this issue.

mnordhaus:
Let me link my Idea here it is not similar, but related: https://github.com/google/docsy/discussions/1093 IMHO it should be possible that the reader can navigate to the contents of the righthand sidebar even on a small screen like an iPhone. I am happy to hear your thoughts on that discussion item. I am currently experimenting with that Idea.

LisaFC:
What do we think of @samiahmedsiddiqui's solution in #1110 (thanks for that!)? We'd need to document it as well obviously.

chalin:
Related:

- #326
- https://github.com/open-telemetry/opentelemetry.io/issues/712

Some thoughts on the subject:

[...]

PR #1110, opened 2022-07-22
Configurable TOC limit with default 200

samiahmedsiddiqui:
It allows users to configure TOC limit or remove the limit completely as discussed in #853.


chalin (changes requested):
Thanks for the suggested PR.

Unfortunately, I don't think that this approach is the best one. For my reasoning, see https://github.com/google/docsy/issues/853#issuecomment-1209475444.

Issue #1114, opened 2022-07-25
Broken URLs since Hugo 101 due to leading slash and relURL


Marzal:
Since [Hugo 101](https://github.com/gohugoio/hugo/commit/a5a4422aaeffe0e9df713cf09c73a9cc423a2e1e), if you use a leading slash in a path with relURL, Hugo will not prepend the baseURL directory as documented in this issue https://github.com/gohugoio/hugo/issues/10049

For context, see:

- Commit https://github.com/gohugoio/hugo/pull/10002
-  Bug https://github.com/gohugoio/hugo/issues/9994

I only have seen problems in 101 with
- https://github.com/google/docsy/blob/main/layouts/partials/scripts.html#L27
- https://github.com/google/docsy/blob/main/layouts/partials/scripts.html#L31

But I suppose that at least these could have the same problem
https://github.com/google/docsy/blob/main/layouts/partials/scripts.html#L74
https://github.com/google/docsy/blob/main/layouts/partials/head.html#L41
https://github.com/google/docsy/blob/main/layouts/partials/community_links.html#L17
https://github.com/google/docsy/blob/main/layouts/404.html#L5
https://github.com/google/docsy/blob/main/layouts/swagger/baseof.html#L6
https://github.com/google/docsy/blob/main/layouts/swagger/baseof.html#L25
https://github.com/google/docsy/blob/main/layouts/swagger/baseof.html#L26

Tasks:

- [x] #1127
- [ ] #1134

LisaFC:
Thanks for spotting that, does it look like this is a permanent breaking change in Hugo that we'll need to work around/update or something that's likely to go away in a future version?

Marzal:
> Thanks for spotting that, does it look like this is a permanent breaking change in Hugo that we'll need to work around/update or something that's likely to go away in a future version?

As far as I understand from the conversation in the links above, the old behavior (Hugo <101) was an unwanted behavior now fixed.

Here is a good summary of how things are now implemented:
https://github.com/gohugoio/hugo/issues/9994#issuecomment-1153588586

chalin:
Thanks for the detailed report @Marzal! I'm investigating.

chalin:
From https://github.com/google/docsy/pull/1127#issuecomment-1197313702, @Marzal wrote:

[...]

PR #1127, opened 2022-07-27
Hugo relURL fix for CSS and JS paths

chalin:
- Contributes to #1114, and hopefully fixes it
- Addresses CSS and JS path changes only -- though that might be the only thing that we need to address (but I'll confirm later)
- According to a very reliable source (`@dep`), the fix consists of the following -- excerpt from https://github.com/gohugoio/hugo/issues/10049#issuecomment-1165948452 (emphasis mine):
  > The gist of it is, res starting with a / is considered relative to the server host, **remove that and it will work**.

/cc @Marzal @daniel-milchev @deining

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -38,7 +38,7 @@
 {{ end -}}
 
 {{ if .Site.Params.prism_syntax_highlighting -}}
-<link rel="stylesheet" href="{{ "/css/prism.css" | relURL }}"/>
+<link rel="stylesheet" href="{{ "css/prism.css" | relURL }}"/>
 {{ end -}}
 
 {{ partial "hooks/head-end.html" . -}}
--- a/layouts/partials/scripts.html
+++ b/layouts/partials/scripts.html
@@ -24,11 +24,11 @@ window.markmap = {
 <script src="https://cdn.jsdelivr.net/npm/markmap-autoloader"></script>
 {{ end }}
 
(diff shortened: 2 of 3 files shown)


chalin:
No changes to the generated UG files, and it fixes one of the Docsy-based projects -- as reported via https://github.com/eclipse-kanto/kanto/issues/57:

> image

Note that the styling of the code blocks is working once again.

So I'll merge now and hopefully the Kanto project folks (like @daniel-milchev) can confirm the fix.

Marzal:
Thanks for the quick fix.

I only have doubts about this one:
https://github.com/google/docsy/blob/main/layouts/404.html#L5
`<p>Oops! This page doesn't exist. Try going back to our <a href="{{ "/" | relURL }}">home page</a>.</p>`
Some other projects put `""` in similar cases
https://github.com/toppers/hakoniwa/commit/ccdd25845d031ec74b3f188ccdd89474bcd4ce1c#diff-42986cf40ad96f5b7ca39d9b71ba6a9f252656de24cfe82100892ff0f4400f90L4

chalin:
Thanks for the reference and the feedback. I've moved your comment into the issue -- https://github.com/google/docsy/issues/1114#issuecomment-1197334953 -- since the issue is still open (I left it open because I knew that there might be more issues to address, like the ones that you're pointing out).

daniel-milchev:
Thank you for the quick fix!
The problem with the code snippets has been resolved.

PR #1128, opened 2022-07-27
Remove `height` attribute on `.td-outer`

### LINKED ISSUE #1116: Problem with scroll-anchoring
Using Firefox 102.0.1, and Google Chrome 103.0.5060.134: 

The `height: 100vh` property on `.td-outer` ([link](https://github.com/google/docsy/blob/main/assets/scss/_main-container.scss#L5)) seems to break scroll-anchoring, so that resizing the window causes the content to scroll to a completely different location. Removing that property restores scroll-anchoring (and doesn't seem to have any other detrimental effects, at least on our site).

It appears that the change in height counts as a [suppression trigger](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor/Guide_to_scroll_anchoring#suppression_triggers): setting `layout.css.scroll-anchoring.suppressions.enabled` to false makes scroll anchoring work as expected.

richvdh:
The presence of this attribute appears to break scroll-anchoring.

Fixes #1116, I hope.

### DIFF
--- a/assets/scss/_main-container.scss
+++ b/assets/scss/_main-container.scss
@@ -2,7 +2,6 @@
 .td-outer {
     display: flex;
     flex-direction: column;
-    height: 100vh;
 }
 
 // The outer page container for the default base template.


richvdh:
per #1116: this breaks footer positioning

PR #1132, opened 2022-07-27
UG feedback: update SEO page

### LINKED ISSUE #1113: Update user guide following #1092 - new meta description
- Code change: #1092
- Revert all/part of #662

chalin:
- Closes #1113

@LisaFC @geriom et al.: I'm on the fence for this one. I'm tempted to **just drop the SEO section**. WDYT?

Preview: https://deploy-preview-1132--docsydocs.netlify.app/docs/adding-content/feedback/#search-engine-optimization-meta-tags

### DIFF
--- a/userguide/content/en/docs/adding-content/feedback.md
+++ b/userguide/content/en/docs/adding-content/feedback.md
@@ -2,10 +2,8 @@
 title: Analytics, User Feedback, SEO
 date: 2019-06-05
 description: >-
-  Add Google Analytics tracking to your site, use the "was this page helpful?"
-  widget data, disable the widget on a single page or all pages, and change the
-  response text. See what data is used to create the `meta description` tag for
-  SEO.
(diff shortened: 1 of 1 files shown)


aimeeu:
Please don't drop that SEO section from the User Guide. I authored the fix in https://github.com/google/docsy/pull/662/. Docsy users should know how the value for the description meta tag is created as you wrote: page description, page summary, site description.`.Site.Params.description` can create duplicate meta tag content across many pages, which search engines do not like.  Granted the site description option is not likely to happen, but it's nice to have the order documented so users don't have to dig like I did when SEO scans come back full of issues. Thanks.

chalin:
Thanks for your feedback @aimeeu and @emckean. If you agree with the changes, could you (GitHub) approve?

LisaFC:
LGTM!

PR #1134, opened 2022-07-28
404 layout: fix relURL argument

### LINKED ISSUE #1114: Broken URLs since Hugo 101 due to leading slash and relURL

### LINKED ISSUE #1133: 404 layout generates invalid HTML
The current 404 layout causes nested `<main>` elements to be generated:

```html
  <main role="main" class="td-main">
    <main id="main">
      <div>
       <h1 id="title">Not found</h1>
          <p>Oops! This page doesn't exist. Try going back to our <a href="/">home page</a>.</p>
      </div>
    </main>
 </main>
```

> image

This is an instance of #907.

chalin:
- Closes #1114
- ~Won't close #1133~ moving this to another PR

Preview: https://deploy-preview-1134--docsydocs.netlify.app/some-invalid-url/

/cc @deining

### DIFF
--- a/layouts/404.html
+++ b/layouts/404.html
@@ -2,7 +2,7 @@
     <main id="main">
       <div>
        <h1 id="title">Not found</h1>
-          <p>Oops! This page doesn't exist. Try going back to our <a href="{{ "/" | relURL }}">home page</a>.</p>
+          <p>Oops! This page doesn't exist. Try going back to our <a href="{{ "" | relURL }}">home page</a>.</p>
       </div>
     </main>
 {{ end }}


chalin:
Ah, now I see why the layout was like it was. I'll need to fix it, but that will have to wait until tomorrow.

chalin:
All: this simple fix is ready for review. I'm going to handle the HTML rework through another PR. Given how trivial a change this is, I'm going to merge. Feedback is always welcome.

PR #1149, opened 2022-08-09
Update Prism JS and CSS to v1.28.0

### LINKED ISSUE #1123: PrismJS is outdated and prone to DOM XSS
A checkmarx scan of our docsy-themed documentation turned up a reported Client DOM XSS vulnerability in `static/js/prims.js`.  It reports:

> The application's o embeds untrusted data in the generated output with element, at line 120 of .../static/js/prism.js. This untrusted data is embedded straight into the output without proper sanitization or encoding, enabling an attacker to inject malicious code into the output.

The problematic 'o' method:

```
function o(e) {
    l.highlightedCode =  e, M.hooks.run("before-insert", l), l.element.innerHTML = l.highlightedCode, M.hooks.run("after-highlight", l), M.hooks.run("complete", l), t && t.call(l.element)
```

I see that `static/js/prism.js` appears to be a hard-coded copy of approximately v1.21.0 of primsjs.  But it isn't clear where exactly it came from.  Should it be pulled in dynamically as an npm dependency?  Or alternatively, a small change is made to `static/js/prims.js` to sanitize `e` before inserting it into the DOM?

chalin:
- Updates Prism JS and CSS. Files downloaded from 
 https://prismjs.com/download.html#themes=prism&languages=markup+css+clike+javascript+bash+c+csharp+cpp+go+java+markdown+python+scss+sql+toml+yaml&plugins=toolbar+copy-to-clipboard
- Closes #1123, hopefully
  WDYT @BradHolmes?

Someone able and willing to test this update?

This update might also address other Prism-related issues.


BradHolmes:
I think this will cover it - thank you!

chalin:
@LisaFC - is this good to go?

LisaFC (approved):

PR #1157, opened 2022-08-12
Card-related shortcode style fixes

### LINKED ISSUE #1154: cards-pane and tabbed-panes style import causes style overrides of Bootstrap card and Docsy highlight styles
The following PR

- #1009

introduced an override of the Bootstrap styles for `.card-deck`, `.card`, etc. through `shortcodes/cards-pane.scss`. This style overrides is an undesirable **breaking change**. 

My guess is that this problem was introduced way back via:

- #532

but that it wasn't as obviously manifested until `cards-pane.scss` got processed along with all other Bootstrap styles (as is done in #1009).

The card-related style changes should only be applied in the context of Docsy cards and card-decks.
@deining - do you agree?

**Edit**: this also happens for the `.td-content .highlight` Docsy style.

- [x] Fix style overrides for:
  - [x] card-related styles - #1157
  - [x] `highlight` - #1163
- [ ] Add a note to the [0.3.0 CHANGELOG](https://github.com/google/docsy/blob/main/CHANGELOG.md#030) entry concerning this breaking change - #1167

chalin:
- Contributes to #1154
- Adds `td-` prefixed classes to Docsy card-related shortcode outer `<div>`s
- No other style changes

If y'all agree with this change, I suggest that we submit a followup PR that informs users of the #1154 breaking change in 0.3.0. (**Edit:** I've added a task to #1154 to this effect.)

Preview: https://deploy-preview-1157--docsydocs.netlify.app/docs/adding-content/shortcodes/#card-panes

### DIFF
--- a/assets/scss/shortcodes/cards-pane.scss
+++ b/assets/scss/shortcodes/cards-pane.scss
@@ -1,19 +1,25 @@
-.card-deck {
-	@extend .td-max-width-on-larger-screens;
+.td-card-deck.card-deck {
+  @extend .td-max-width-on-larger-screens;
 }
-.card {
-	@extend .td-max-width-on-larger-screens;
-	.highlight {
-		border: none;
-	}
-}
-.card-body.code {
-	background-color: #f8f9fa;
-	padding: 0 0 0 1ex;
-}
-.card-body {
-	pre {
-		margin: 0;
-		padding: 0 1rem 1rem 1rem;
-	}
+
+.td-card {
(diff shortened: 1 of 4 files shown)


LisaFC (approved):

chalin:
Thanks for the review @LisaFC. I'll be AFK for a bit and will await to see if others (esp. @deining) comment on this. When I get back I'll merge and continue working on the other fix for #1154.

PR #1159, opened 2022-08-14
Respect ui.ul_show with ui.sidebar_menu_foldable

metafeather:
The docs are ambiguous about combining `ui.sidebar_menu_foldable` with `ui.sidebar_menu_compact` and `ui.ul_show` but the fix is trivial so that the desired number of levels of menus are automatically opened in the sidebar since `$show` already depends on `$activepath`


LisaFC:
@raum51 do you want to take a look to make sure I haven't missed anything?

PR #1160, opened 2022-08-14
Display all Sections that are descendants of Page

metafeather:
By default Hugo only return the immediate nested Section of a Page declared by `_index.md` , forcing users to add empty files to walk the directory tree to nested docs.
This partial and update to `$pages_tmp` enables the sidebar to discover and display all docs in a tree, flattening lower levels into their closest Parent Section.
This also matches the behaviour in `section-index.html`


metafeather:
This shouldn't be necessary given the Hugo docs on [Nested Sections](https://gohugo.io/content-management/sections/) but is practice I couldn't get deeply nested sections to work with Docsy.

The addition of [`RegularPagesRecursive`](https://gohugo.io/variables/page/#pages) also seems to indicate that nested sections/pages do not work as expected

LisaFC:
Thanks for this - before review, have you also asked about this on the Hugo forum? If Hugo is supposed to support deeply nested sections but they're not actually working it may be a broader Hugo issue/there may be broader fixes in the works.

chalin:
- The underlying Hugo behavior question was never settled and the draft predates the restructure. Thanks.

PR #1161, opened 2022-08-14
Ensure a Page always has a Title

metafeather:
Pages without Frontmatter for `title` or `linkTitle` result in empty entries in the sidebar and list pages.

This partial provides additional fallback Titles based on the Content H1 or File name, which can replace use in multiple templates where `$s.linkTitle` is hardcoded


LisaFC:
I like this idea! Do you have any examples of what a title pulled from the file name would look like?

We should also document this behaviour in the user guide, maybe with a small update to the (increasingly enormous) Navigation and Search page:
https://www.docsy.dev/docs/adding-content/navigation/#section-menu

chalin (commented):
Interesting. We started discussing this during the PSC meeting today. I have a few potential concerns which I'll share a little later. (Marking this as draft in the meantime to ensure it doesn't get inadvertently merged.)

chalin:
- The draft went inactive and predates the restructure. Thanks.

PR #1163, opened 2022-08-15
Avoid tabbed-pane.scss style leakage

chalin:
This PR applies minimal style changes in order to address the `highlight` style override reported in:

- #1154

---

Preview: https://deploy-preview-1157--docsydocs.netlify.app/docs/adding-content/shortcodes/#tabbed-panes

### Screenshots

Note the equal amount of spacing before and after the code block in the **After** screenshot.

| Before | After |
|--|--|
| image | image |

### DIFF
--- a/assets/scss/shortcodes/tabbed-pane.scss
+++ b/assets/scss/shortcodes/tabbed-pane.scss
@@ -1,4 +1,4 @@
-.td-content {
+.tab-content {
   .highlight {
     margin: 0rem 0 2rem 0;
   }


assets/scss/shortcodes/tabbed-pane.scss
```diff
@@ -1,4 +1,4 @@
-.td-content {
+.tab-content {
   .highlight {
     margin: 0rem 0 2rem 0;
   }
```
chalin, line 4: As I understand it, a code-block cannot appear in a `tab-content` outside of a `tab-pane`, but there is an override of `highlight` in `tab-pane` already, so I think that this style change is dead code and should be removed. WDYT @deining @LisaFC?

LisaFC: @deining may have a better take on it - I'll take a look, but my immediate thought is possibly we might want a style for `highlight` content that's in a tabbed pane but not inside a code block. Though if they're the same style settings....

LisaFC: Though hmm, we only apply `highlight` to code blocks, so that doesn't make sense.

chalin: This fixes the style "leakage". Can you approve it as is @LisaFC?
We can address the possible style "dead code" separately.

LisaFC: Sure

LisaFC (approved):

assets/scss/shortcodes/tabbed-pane.scss
chalin: We'll track this via:

- #1168

PR #1174, opened 2022-08-15
Added option to disable title in the navbar

### LINKED ISSUE #1025: Navigation Bar Logo and Title - Duplicate Site Name
It seems that if you add a logo that includes a name in addition to a graphic, as per https://www.docsy.dev/docs/adding-content/iconsimages/, you will get the name duplicated in the navigation bar.

So your logo would look like:  graphic, example.com, example.com

Although there is an option to hide the logo in the config file, there is no option to hide the text.
     navbar_logo = true

Maybe if something like this gets introduced:

     navbar_title = true

Then something like this can be introduced in navbar.html

<span class="navbar-title">{{ if .Site.Params.ui.navbar_title }}.........  can be used.

For the time-being, I have cloned navbar.html  and removed the following section from the end of line 4:
<span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>

gitautas:
Fixes #1025

### DIFF
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -1,7 +1,13 @@
 {{ $cover := and (.HasShortcode "blocks/cover") (not .Site.Params.ui.navbar_translucent_over_cover_disable) }}
 <nav class="js-navbar-scroll navbar navbar-expand navbar-dark {{ if $cover}} td-navbar-cover {{ end }}flex-column flex-md-row td-navbar">
         <a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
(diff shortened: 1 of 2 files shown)


chalin (changes requested):
Thanks, but I'd rather that we not introduce yet another config option for something that can be handled elegantly with custom styling. For details, see https://github.com/google/docsy/issues/1025#issuecomment-1219647469.

chalin:
#1025 has been fixed via #1190. For instructions on hiding the project name, see https://www.docsy.dev/docs/adding-content/lookandfeel/#styling-your-project-logo-and-name. Thanks.

PR #1176, opened 2022-08-16
Use SCSS variable for FA font name

### LINKED ISSUE #1170: Use SCSS variable for FA font name
We should define a variable for the FA font name and use it here and elsewhere. This can be done in this PR or a followup PR.

_Originally posted by @chalin in https://github.com/google/docsy/pull/1102#discussion_r945562824_

samiahmedsiddiqui:
- Closes #1170

### DIFF
--- a/assets/scss/_nav.scss
+++ b/assets/scss/_nav.scss
@@ -121,7 +121,7 @@
             font-variant: normal;
             text-rendering: auto;
             -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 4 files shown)


chalin:
No change to the generated site files other than the CSS map:

```console
$ (cd userguide/public && git diff) | grep ^diff
diff --git a/scss/main.css.map b/scss/main.css.map
```

I also double checked that there is but one hardcoded "Font Awesome 6 Free" string, namely the variable definition.

Issue #1182, opened 2022-08-18
Ensure that versions of dependencies remain in sync


chalin:
We need to ensure that versions of dependencies remain in sync across installation options, namely:

- Hugo modules
- NPM packages
- CDN

deining:
> We need to ensure that versions of dependencies remain in sync across installation options, namely:
> 
>     * Hugo modules
> 
>     * NPM packages
> 
>     * CDN

I agree.

Just created PR #1218 in order to achieve that for now.

PR #1187, opened 2022-08-18
Scripts: get bootstrap.js and popper.js at build time

chalin:
- Contributes to:
  - #1182
  - #605 by eliminating the CDN call to Bootstrap and making the CDN call to Popper at build time


chalin:
As is mentioned in [Components] from the Bootstrap docs:

> Tooltips and popovers [are used] for displaying and positioning (also requires [Popper](https://popper.js.org/))

Tooltips are still working (they're used in the footer):

> image

[Components]: https://getbootstrap.com/docs/4.6/getting-started/introduction/#components

LisaFC (approved):
Looks good!

chalin:
Thanks for the review. Merged so that I can more easily experiment with:

- #470

deining:
This commit broke docsy module installations. We need to add a mount so that bootstrap.js becomes available with module installations (see PR #1216).

PR #1037, opened 2022-06-03
Improve tabbed pane: allow right aligned tabs, allow tab groups

deining:
This PR improves and extends the shortcodes for tabbed panes:
* the shortcodes learned to show right aligned tabs (by giving `right=true` on the `tabpane`shortcode).
* the shortcodes now support spllitting up tabs in a left aligned and a right aligned tab group (by giving `right=true` on the dividing `tab` shortcode(s)). If wanted, even multiple tab groups can be rendered.

The PR also contains fixes as well as css and code improvements.
The PR is ready for review and subsequent merging.

PR #1202, opened 2022-08-29
Set language for tabs in URL query

titusfortner:
I've been using this in custom-hacked code for a while, but now that I'm upgrading to the most recent tabpane implementation (nice work on that!), I figured I should submit it to see if I can avoid adding my own hacks. :)

This would allow users to link directly to the language they want:

https://selenium.dev/documentation/webdriver/actions_api/mouse/?language=ruby#forward-click


geriom:
This is not working for me. It seems the persistent tab functionality is broken, but has nothing to do with your PR. I'm seeing an error on console:

image

I'm going to try and figure out what's going on.

[The page you linked to](https://selenium.dev/documentation/webdriver/actions_api/mouse/?language=ruby#forward-click) also shows an error, but a different one:

image

That one is likely caused by a recent update to the [tabpane shortcode](https://github.com/google/docsy/blob/4a41d9b2de37caecb9532df89c4f867feb0432b3/layouts/shortcodes/tabpane.html#L71) that changed `handleClcik` to `persistLang`. Updating to the latest docsy release should fix that.

titusfortner:
Looks like we're not actually using the latest version of docsy in our code right now either. Trying to update all the things right now. Let me put this in draft, and I'll ping you when I can verify it works. Thanks.

titusfortner:
Ok, I got the selenium.dev example working. We're using a version of docsy from a couple months ago, but the only commits that might change things are in #1037, and I don't see anything specific there.

My previous PR commit didn't set the local session properly, but that shouldn't result in the error you are seeing... I'm not a JS dev, though and don't know what `$('#' + element.id).tab('show')` is actually doing.

Also as a side note, is there a reason for the change from `code=true` to `text=false`? One option literally results in a `code` tag being generated and the other can be either html or markdown depending on the Hugo syntax used. I guess I think "plain text" when I see "text" (and also I don't want to update everything again). But also I know my project isn't the only one that uses git submodules instead of actual releases....

deining:
> This is not working for me. It seems the persistent tab functionality is broken, but has nothing to do with your PR. I'm seeing an error on console:
> 
> image

This doesn't work for me either (with tip of `main` branch, I get the same message on the error console). I started looking into that and found out that this is a regression introduced with 10fae88de482d836f0c81437afad5674418a888d. However, it's not fully clear to me yet what exactly is going wrong here. I hope this information can serve as starting point for further investigations.

geriom:
I tried this on you website and it works as expected, but for some reason it doesn't work when I try it with the docsy user guide. It's probably related to another issue and not your implementation. I gonna take another look today and get it merged.

titusfortner:
Well, We're not using the latest version of Docsy, so it's possible something else is getting in the way. :( 

I don't have time to dig into it, so we can put this into draft if you want and I can revisit it once @diemol updates us to the latest version of Docsy.

static/js/tabpane-persist.js
```diff
@@ -1,5 +1,15 @@
...
-    const activeLanguage = localStorage.getItem('active_language');
+    let activeLanguage = localStorage.getItem('active_language');
+
+    // Get active language from URL params if exists
+    const params = new Proxy(new URLSearchParams(window.location.search), {
+        get: (searchParams, prop) => searchParams.get(prop),
+    });
+    if (params.language !== null) {
+        activeLanguage = params.language;
+        localStorage.setItem('active_language', activeLanguage);
```
deining, line 10: I am uncertain if we should persist the active language given when given as URL parameter.
Currently, the langauge is always persisted, even if the user set `persistLang=false` for the tab pane. So maybe it is better to not persist the language at all when given as URL parameter.

deining (approved):
Overall, this PR looks good to me.

geriom:
> I can't confirm- For me, setting the language via URL parameter works as expected, both with a module site from scratch and with the user guide.
 
Must be something with my local environment, the Netlify preview also works fine.

> I am uncertain if we should persist the active language given when given as URL parameter. Currently, the language is always persisted, even if the user set `persistLang=false` for the tab pane. So maybe it is better to not persist the language at all when given as URL parameter.

I think it's OK to persist the language. I feel that having it in the URL explicitly makes a language selection that supersedes the `persistLang` flag.

titusfortner:
Maybe we can unclear it if they switch to another tab and persistLang isn't set? persistLang just makes much more sense when used in conjunction with #1227.

PR #1203, opened 2022-08-29
Add shortcode rendering in readfile shortcode

geriom:
This allows importing files that contain shortcodes themselves, and the shortcodes will be rendered correctly.

Requires Hugo v0.101.0+ to work correctly, for older versions this change has no effect on the output.

[The PR docs preview shows an example](https://deploy-preview-1203--docsydocs.netlify.app/docs/adding-content/shortcodes/#include-external-files)

### DIFF
--- a/userguide/content/en/docs/adding-content/shortcodes/index.md
+++ b/userguide/content/en/docs/adding-content/shortcodes/index.md
@@ -594,6 +594,10 @@ contents:
 ```go-html-template
 ## Installation
 
+{{%/* alert title="Note" color="primary" */%}}
+Check system compatibility before proceeding.
+{{%/* /alert */%}}
+
 1.  Download the installation files.
 
 1.  Run the installation script
--- a/userguide/content/en/docs/adding-content/shortcodes/index.md
+++ b/userguide/content/en/docs/adding-content/shortcodes/index.md
@@ -617,7 +617,9 @@ The following section explains how to install the database:
 
 ```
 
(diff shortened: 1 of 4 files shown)


userguide/content/en/docs/adding-content/shortcodes/index.md
```diff
@@ -594,6 +594,10 @@ contents:
 ```go-html-template
```
LisaFC, line 594: Should we add something in this section to the effect that you can only have shortcodes inside your includes from x version of Hugo, otherwise users who haven't updated might be disappointed to see shortcodes not rendering?

geriom: Good point. Added a note to the user guide.

Thanks!

LisaFC (approved):

polarweasel:
@LisaFC and @geriom We're running into a problem with this new readfile shortcode: it will not correctly render Docsy's tabpanes. Was there an issue associated with this PR that we can reopen, or do you want a new one?

geriom:
@polarweasel are you still having an issue with tabpanes?

Something like this should work:

```
{{< tabspane >}}

{{< tab "Tab1" >}}
Tab1 text
{{< /tab >}}

{{< tab "Tab2" >}}
Tab2 Text
{{< /tab >}}

{{< tab "Tab3" >}}
Tab3 Text
{{< /tab >}}

{{< /tabpane >
```

If it doesn't work feel free to open a new issue and @ me there.

PR #1204, opened 2022-08-29
feat: Expose cdn settings as params

bradbeam:
This moves the configuration of much of the cdn (js/css) into the configuration
so they can be more easily maintained.

Signed-off-by: Brad Beam <brad.beam@b-rad.info>

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -27,13 +27,13 @@
 {{ template "_internal/twitter_cards.html" . -}}
 {{ partialCached "head-css.html" . "asdf" -}}
 <script
-  src="https://code.jquery.com/jquery-3.6.0.min.js"
-  integrity="sha384-vtXRMe3mGCbOeY7l30aIg8H9p3GdeSe4IFlP6G8JMa7o7lXvnz3GFKzPxzJdPfGK"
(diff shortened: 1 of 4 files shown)


LisaFC:
Thanks for this! This is a potentially interesting approach (especially as we have some users who don't necessarily want to use our chosen CDNs), but is presumably going to break any site that doesn't update their config.toml/yaml

Thoughts @chalin @geriom @emckean?

fekete-robert:
Maybe we could use the [default](https://gohugo.io/functions/default/#readout) function of hugo, and use the config parameters if set, or use the original cdns if not.

LisaFC:
I'll have a look at how the scripts partial would look if we used defaults.

bradbeam:
Sorry for the delay, I just pushed up changes to include defaults inside the templates.

deining:
> This moves the configuration of much of the cdn (js/css) into the configuration so they can be more easily maintained.

I like the idea. This PR does not pull in the latest versions of the resources, however (e.g. it pulls in KaTeX 0.15.1, while latest version is 0.16.2.) Were great if this could be improved.

chalin:
- Superseded: integration versions are moving into site params (Mermaid today; KaTeX and markmap tracked in #2705 and #2706). Thanks.

PR #1206, opened 2022-08-30
 map keys must be unique in YAML

lavaicer:
Map keys must be unique YAML. After remove the second `imports`, it works


lavaicer:
Should I have another try?

PR #1208, opened 2022-08-30
map keys must be unique in YAML

lavaicer:
After removed the second imports, it works.


deining (approved):

deining:
Thanks for spotting and correcting this!

PR #1209, opened 2022-08-30
User guide, conversion to modules: fixing config.yaml

deining:
This PR fixes yaml syntax in config files in the user guide (continuation of #1208, which was incomplete).
It also updates to latest Hugo version 102.1.


chalin (approved):
👍

chalin:
Thanks for these fixes, and for your patience!

PR #1211, opened 2022-08-31
Avoiding 'Illegal shortcode state' (Hugo v0.102.2)

deining:
This PR avoids an Ìllegal shortcode state`, thrown by Hugo version 0.102.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,6 @@
     "bootstrap": "4.6.2"
   },
   "devDependencies": {
-    "hugo-extended": "0.101.0"
+    "hugo-extended": "0.102.2"
   }
 }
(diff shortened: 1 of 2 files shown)


deining:
Rendering of user guide failed due to a regression introduced in hugo v0.102.0, fixed with recently released version hugo v0.102.3.    
I would like to keep this PR open though, since the syntax cahnge of the shortcode make sit easier to read. I still think it is a good thing to merge this PR.

LisaFC:
Agree it's easier to read - I think we have some stuff in shortcodes (not sure about this one) that were added for compatibility with multiples of versions of Hugo ago.

LisaFC:
This version renders properly with the previous version of Hugo, right?

deining:
> This version renders properly with the previous version of Hugo, right?

Yes, it I tested it with hugo v0.101 and v0.102.x, and the result was always fine.

PR #1220, opened 2022-09-03
[conditional-text] Avoid linebreaks when using the shortcode inline

fekete-robert:
Until now, using the conditional-text shortcode inside a paragraph
caused linebreaks before and after the conditionally included text.
This patch fixes the issue.

### DIFF
--- a/layouts/shortcodes/conditional-text.html
+++ b/layouts/shortcodes/conditional-text.html
@@ -1,36 +1,36 @@
 <!-- Get the current buildcondition from the config and lowercase it -->
-{{ $condition := lower $.Site.Params.buildCondition }}
+{{- $condition := lower $.Site.Params.buildCondition -}}
 
-{{ if ne $condition "" }}
+{{- if ne $condition "" -}}
     
     <!-- Get the parameters from the shortcode invocation and lowercase them.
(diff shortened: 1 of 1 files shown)


fekete-robert:
I'd consider the linebreaks to be a bug, so in that sense it's not a
breaking change IMHO. Don't know if anyone else was actually using it.

LisaFC:
Good point! OK, I'm happy to merge this as agreed, the existing behaviour is not what I'd expect.

layouts/shortcodes/conditional-text.html
```diff
@@ -1,36 +1,36 @@
 <!-- Get the current buildcondition from the config and lowercase it -->
-{{ $condition := lower $.Site.Params.buildCondition }}
+{{- $condition := lower $.Site.Params.buildCondition -}}
```
chalin, line 2: Thanks for cleaning up the whitespace issues in this shortcode @fekete-robert.

For next time, it is best to only add the `-` marker where it is needed. As a general rule of thumb, I favor the end tag `-}}`. More rarely do we need `-` on an opening tag, and certainly not in the case where two directives follow each other like this:

```nocode
{{ if ... -}}
  {{ $var := val -}}
  {{ $var := val2 -}}
...
```

/cc @LisaFC

PR #1227, opened 2022-09-05
keep height of persistent language tabs constant

titusfortner:
Again, not a JS dev, but this seems to work as desired.

I haven't merged it into my project yet (https://github.com/SeleniumHQ/seleniumhq.github.io/pull/1157), but you can see an example by going here and clicking the "Kotlin" tab:

https://deploy-preview-1157--jovial-austin-42fe02.netlify.app/documentation/webdriver/actions_api/mouse/#drag-and-drop-by-offset

Not sure if you prefer an additional toggle added for this behavior & I know the shortcode docs section would need to be updated, either way.


LisaFC:
Geri, I know you were looking at some tabs stuff, can you take a look?

geriom:
Thank you @titusfortner, I'm looking into this later today.

geriom:
@deining Since you are the main contributor to this feature. Can you please take a look when you have a change?

titusfortner:
Well, something is wrong with this code after all. Clicking the tabs and measuring the heights isn't always giving accurate heights and I'm not sure why. If I throw a debug point and execute the code in the console it gives a different result from when it is executed in the js file. I have no idea if it's in a different context or what is going on. :(

AryanP45:
LGTM

titusfortner:
@AryanP45 you can see from this example: https://deploy-preview-1157--jovial-austin-42fe02.netlify.app/documentation/webdriver/browser/windows/?language=python#create-new-window-or-new-tab-and-switch

There's something wrong in the calculations, and I just couldn't figure out the issue. For some reason clicking the tab in the JS and calculating the height was giving me different value than when I ran it in the console. I don't know enough JS to know how to fix it.

Full conversation is here: https://github.com/SeleniumHQ/seleniumhq.github.io/issues/1167

Would be really great if someone could at least figure out *why the JS isn't determining the correct tab-content heights when executing. I put in a debug point to ensure it isn't a race condition. Just no idea why it's happening. Can't hope to do a workaround without that.

AryanP45 (approved):

chalin:
- @titusfortner - which version of Docsy are you using? AFAIR, there have been some updates to the tabpane/tab shortcodes. From looking at your site I get the impression that the version of Docsy that you're using doesn't have the latest fixes. Thanks for all the context you provided.
- @deining - gentle ping, in case you have something to add.

chalin:
I can't reproduce the problem reported in https://github.com/SeleniumHQ/seleniumhq.github.io/issues/1167 under macOS Chrome. @geriom, can you reproduce it? (Works fine under macOS Firefox too.)

geriom:
> I can't reproduce the problem reported in [SeleniumHQ/seleniumhq.github.io#1167](https://github.com/SeleniumHQ/seleniumhq.github.io/issues/1167) under macOS Chrome. @geriom, can you reproduce it? (Works fine under macOS Firefox too.)

Same, I can't reproduce either it under Linux Chrome and Firefox.

titusfortner:
Huh, when it happens for me it's been consistent, it's just not obvious to me why it happens in one place and not another. Seems to be the tabs with the most height?  We could set it not to overwrite height value if the max height is over a certain amount, but those are the tabs that end up causing the most problems. :)

titusfortner:
Erf, but for some reason the tab height does not work when I combine it with #1202 
So, this one might need more work again.

titusfortner:
I was just combining them incorrectly.

This example includes code from both of my Docsy PRs:
https://deploy-preview-1530--selenium-dev.netlify.app/documentation/webdriver/interactions/windows/?tab=python#execute-script

PR #1233, opened 2022-09-09
Fix: tab-pane, persistence of language: change load order of .js-scripts

### LINKED ISSUE #1232: Tab pane, persistence of langauge: js error
As mentioned and discused in #1202, a regression was introduced with 10fae88de482d836f0c81437afad5674418a888d:
When persisting the language of a tab pane, the following error is now thrown in the browser's javascript console:

```
Uncaught TypeError: $(...).tab is not a function
    <anonymous> http://localhost:1313/js/tabpane-persist.js:7
    <anonymous> http://localhost:1313/js/tabpane-persist.js:6
```

deining:
This PR fixes #1232. The problem mentioned there is caused by the fact that `js/tabpane-persist.js` is loaded prior to the `bootstrap` script. Simply changing the load order of the scripts cures the problem.


LisaFC (approved):

geriom (approved):
Thank you!

Issue #1198, opened 2022-08-25
docs navigation arrow on hover UX


Arhell:
example
![Screenshot_1](https://user-images.githubusercontent.com/26163841/186651177-3c8e586f-15f1-47a2-bf59-1ece9cfb48c3.png)

On hover, it should either be turned down or not, because on hover there is a color change and this shows that the item can be clicked.

And at the moment it rotates 30 degrees and looks like it is not behaving correctly.

I propose to remove this effect or change it to a slight offset rather than a rotation

LisaFC:
Oh yes, just enabled it myself in a site and it sort of wobbles on hover. That does look a bit odd. @raum51 @chalin thoughts?

PR #1237, opened 2022-09-15
update docs navigation arrow on hover

Arhell:
resolve https://github.com/google/docsy/issues/1198
I removed these turns, left the color change


LisaFC:
Thanks for this! Definitely less wobbly now.

PR #1245, opened 2022-09-28
Add click to copy button to code samples

geriom:
Adds a click to copy button to code samples that don't use the Prism highlighter.

This implementation requires the latest version of Popper. I ran a `grep` on the whole website and couldn't find any other Popper instances. So it seems updating the Popper library won't affect anything.

## Screenshots

| Normal view | Mouseover | Clicked |
| ----------------- | :-------------: | ---------- |
| Screenshot from 2022-09-28 17-11-19 | Screenshot from 2022-09-28 17-11-31 |  Screenshot from 2022-09-28 17-11-43 |

### DIFF
--- a/assets/js/click-to-copy.js
+++ b/assets/js/click-to-copy.js
@@ -1,0 +1,86 @@
+let codeListings = document.querySelectorAll('.highlight > pre');
+
+for (let index = 0; index < codeListings.length; index++)
+{
+  const codeSample = codeListings[index].querySelector('code');
+  const copyButton = document.createElement("button");
+  copyButton.setAttribute('type', 'button');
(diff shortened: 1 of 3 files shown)


geriom:
I'm going to merge this to include it in the release. I don't foresee any problems, but we can always roll it back in a revision release if there's a conflict.

chalin:
> This implementation requires the latest version of Popper. I ran a `grep` on the whole website and couldn't find any other Popper instances. So it seems updating the Popper library won't affect anything.

There is an issue with BS. For details, see #1264.

chalin:
If / once this makes it into a published release, maybe it would be good to mention this wonderful new feature in the changelog? WDYT @geriom?

PR #1248, opened 2022-09-29
Hide Breadcrumb navigation from particular pages not the entire website

### LINKED ISSUE #983: How hide breadcrumbs navigation on some docs pages  ?
Hi,
Is it possible to hide breadcrumb navigation for the root file of the doc (_index.md) ?

Example : 
I want to display Breadcrumb navigation if the doc (content A) have a child content (content B) : 
content A-> content B
But I want to hide if I visit the main content A page .

thank you,
C.

samiahmedsiddiqui:
- Fix #983


samiahmedsiddiqui:
@chalin @LisaFC Please review.

chalin:
Hi. Breadcrumbs can be hidden using CSS. We don't need another config option, IMHO. I'll comment more soon (not before next week).

chalin:
> @chalin Let's suppose I have 2,000 docs pages and I want to hide breadcrumb on 200 pages. How can I do that?

Hi @samiahmedsiddiqui. How would you do it within the scope of this PR? (I'm asking because it will better help me understand your use case, how you see it resolved, and whether there _might_ be an alternative solution that is already available.)

samiahmedsiddiqui:
@chalin I am doing this by defining another parameter in page. If you add `breadcrumbDisable: true` in the file then the breadcrumb will be disable for that particular page.

samiahmedsiddiqui:
@chalin Please review the PR.

fekete-robert (approved):
LGTM, but if it gets approved we should also mention that in the user guide.

samiahmedsiddiqui:
@chalin @fekete-robert Just wanted to nudge about the PR. Any chance we can get it merged soon?

huanlin:
Hey, I have tried the changes in my own docsy website, and after I see the result, I'm curious about when we need this feature.
Let me explain with an example:

Say I have a `_index.md` page with the following front matter:

```
---
title: How-tos
breadcrumbDisable: true
---
```

When I navigate to this "How-tos" page, the breadcrumb is not displayed. This part is OK.
Then I navigate to one of the child page of this "How-tos" page, say "Apply for your API key", and this child page will display the breadcrumb like below:

[My Docs](https://) / [How-tos](https://) / Apply for your API key

So, the user can still see "How-tos" in the breadcrumb. This makes me curious: then why would we hide "How-tos" from the breadcrumb? Afterall, users can still see it from the child pages.

Hopefully I explain it clearly.

samiahmedsiddiqui:
@huanlin This pull request aims to selectively remove the breadcrumb navigation element from specific pages within the website, without affecting its display on other pages or child pages. As a concrete example, this change will prevent the breadcrumb from appearing on the Homepage, while still allowing it to be visible on all other pages, including those that link back to the Homepage as a parent.

samiahmedsiddiqui:
@chalin When can I expect this PR to be merged? I want this feature to be released so that I no longer need to update the templates.

chalin:
For a proposed solution, see the following PR, which reuses the same config param name for both site and page/section-front-matter:

- #2161

PR #1251, opened 2022-10-04
CHANGELOG: mention FA font-glyph width changes

chalin:


chalin:
Going to merge now since I want to bring in more CL changes. I don't expect this edit to be too controversial, but we can always tweak the log file in a followup PR.

PR #1017, opened 2022-05-23
Apply table styling to all tables in td-content, not just direct chil…

NatasaKralj:
As stated in https://github.com/google/docsy/issues/883, tables were not being styled properly in `<details><summary>` blocks. I found the same issue to affect tables nested in list elements. 
Removing `>` from the table entry enables proper styling to be applied to any table element that's a child of `td-content`.

Issue #883, opened 2022-02-08
Markdown table rendering doesn't occur inside folding summary


chipzoller:
Probably not using the right terms, but if you nest a Markdown table inside a folding/summary block, the table doesn't get rendered. Searched and didn't see an issue (open or closed) which references this. Also on an older version of Docsy so maybe this has already been addressed. Here's what this looks like when you compare a rendered table to a non-rendered table inside one of these folding/summary blocks:

Good
![image](https://user-images.githubusercontent.com/29008123/153017751-5ef38e6d-d4f5-4a41-af83-54b41be486f4.png)

Bad
![image](https://user-images.githubusercontent.com/29008123/153017801-be8c401e-c493-4b33-b61d-28aaae3d6b98.png)

LisaFC:
Can you paste in the source and hopefully I can see what's going on? Markdown + CSS has many weirdnesses, it may just need some more specific styling provided.

chipzoller:
Hi there, @LisaFC . Sure. First table is just in a "clean" Markdown seaction:

```
| Value Type   | Input Type | JMESPath Type | Formatting |
|--------------|------------|---------------|------------|
| Number       | Integer    | Literal       | backticks  |
| Quantity     | String     | Raw           | quotes     |
| Duration     | String     | Raw           | quotes     |
| Labels (map) | Object     | Literal       | backticks  |
```

Second one is inside a folding block (I really don't know the correct terminology here):

```
### Add

<details><summary>Expand</summary>
<p>

The `add()` filter very simply adds two values and produces a sum. The official JMESPath library does not include most basic arithmetic operators such as add, subtract, multiply, and divide, the exception being `sum()` as documented [here](https://jmespath.org/specification.html#sum). While `sum()` is useful in that it accepts an array of integers as an input, `add()` is useful as a simplified filter when only two individual values need to be summed. Note that `add()` here is different from the `length()` [filter](https://jmespath.org/specification.html#length) which is used to obtain a _count_ of a certain number of items. Use `add()` instead when you have values of two fields you wish to add together.

`add()` is also value-aware (based on the formatting used for the inputs) and is capable of adding numbers, quantities, and durations without any form of unit conversion.

| Input 1            | Input 2            | Output   |
|--------------------|--------------------|----------|
| Number             | Number             | Number   |
| Quantity or Number | Quantity or Number | Quantity |
| Duration or Number | Duration or Number | Duration |
<br>

[...]

PR #1255, opened 2022-10-05
Apply styling to all elements in td-content

the-kraljica:
This PR is based off of [PR 1017](https://github.com/google/docsy/pull/1017), as I could not get clarity on signing the Google CLA from my corporate GH account. I hope the contribution is still useful, my apologies for the overly long wait time.

As stated in https://github.com/google/docsy/issues/883 (and the original PR), tables were not being styled properly in `<details><summary>` blocks. I found the same issue to affect tables nested in list elements.
Removing `>` from the table entry enables proper styling to be applied to any table element that's a child of td-content.

As per the request in the original PR comments, I also removed the `>` selector from `blockquote`, `ul li` and `ol li` elements.

### DIFF
--- a/assets/scss/_content.scss
+++ b/assets/scss/_content.scss
@@ -35,7 +35,7 @@
         @extend .img-fluid;
     }
 
-    > table {
+    table {
         @extend .table-striped;
 
         @extend .table-responsive;
@@ -43,14 +43,14 @@
         @extend .table;
     }
 
-    > blockquote {
+    blockquote {
         padding: 0 0 0 1rem;
         margin-bottom: $spacer;
         color: $gray-600;
         border-left: 6px solid $secondary;
     }
 
-    > ul li, > ol li {
+    ul li, ol li {
         margin-bottom: .25rem;
     }
 


chalin (approved):
LGTM, but would really want @geriom to give the final approval (and merge). Thanks all!

geriom (approved):
LGTM, Thank you!

anoadragon453:
Hello! This change caused `<table>` tags in our docsy site to have `display: block` instead of `display: table`, which broke some of our assumptions: https://github.com/matrix-org/matrix-spec/pull/1295/files#r1010523059.

I'm not sure if this was intended behaviour, but leaving a breadcrumb here if anyone would like to investigate on the docsy side :)

chalin:
@anoadragon453 - thank you for the in-depth analysis and feedback. Given that Docsy assumes that tables are responsive by default, yes this was intended. We're all looking forward to the Bootstrap upgrade. You may be right that after the upgrade your style change will be unnecessary. I'll add a comment to the Bootstrap-upgrade issue as a reminder for us to check.

PR #1266, opened 2022-10-08
Revert to Popper v1 so tooltips work

### LINKED ISSUE #1264: Version of Popper isn't compatible with Bootstrap 4, so revert?
In #1245, @geriom wrote:

> This implementation requires the latest version of Popper. I ran a `grep` on the whole website and couldn't find any other Popper instances. So it seems updating the Popper library won't affect anything.

Docsy is still using Bootstrap 4.x, which officially [requires Popper `^1.16.1`](https://getbootstrap.com/docs/4.6/getting-started/introduction/#separate), e.g., see:

> https://github.com/twbs/bootstrap/blob/v4.6.2/package.json#L96-L99
> ```json
>   "peerDependencies": {
>     "jquery": "1.9.1 - 3",
>     "popper.js": "^1.16.1"
>   },
> ```

I don't know if anything will break, but if the official BS docs require Popper v1, I'd stick with v1 for now --at least for the next release (0.5.0) of Docsy. 

Once we upgrade to Bootstrap 5 (#470), we'll be able to upgrade [Popper to v2](https://getbootstrap.com/docs/5.2/getting-started/introduction/) as well.  Thoughts?

chalin:
- Effectively disabling #1245 because it requires Popper v2, which is incompatible with the version of Bootstrap Docsy is using. I've chosen to leave in the code and styling.
- Closes #1264

/cc @geriom

### DIFF
--- a/layouts/partials/scripts.html
+++ b/layouts/partials/scripts.html
@@ -48,7 +48,7 @@ window.markmap = {
     {{ printf "onload='renderMathInElement(%s, %s);'" (( .Site.Params.katex.html_dom_element | default "document.body" ) | safeJS ) ( printf "%s" ( $.Site.Params.katex.options | jsonify )) | safeHTMLAttr }}></script>
 {{ end -}}
 
-{{ $jsPopper := resources.GetRemote "https://cdnjs.cloudflare.com/ajax/libs/popper.js/2.9.2/umd/popper.min.js" -}}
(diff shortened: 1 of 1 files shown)


chalin:
Tooltips are working once again:

> image

Merging now. Comments for followup changes are welcome, if any @geriom et al.

chalin:
- Followup issue: #1267

PR #1200, opened 2022-08-28
Fix different order on side-nav and blog list

LukeTimeWalker:
Fixes #1197

Issue #1197, opened 2022-08-24
Blog posts order differs in side-nav vs list


chalin:
To reproduce, create a new project:

```console
$ hugo new site myproject
$ cd myproject
$ npm install --save-dev google/docsy autoprefixer postcss-cli postcss
$ echo 'theme = "docsy"' >> config.toml
$ echo 'themesDir = "node_modules"' >> config.toml
```

Then create two posts with the same date, such as:

- **blog/post-1.md**:
   ```md
  ---
  title: Learn
  date: 2022-08-24
  ---
  ```
- **blog/post-2.md**:
   ```md
  ---
  title: Go
  date: 2022-08-24
  ---
  ```

Build and/or serve the site and visit `/blog`. This is what you'll see:

> <img width="638" alt="image" src="https://user-images.githubusercontent.com/4140793/186496294-bb53154c-28a9-4ced-bfd9-81758ed3440b.png">

Notice how the order of the posts differs.

LukeTimeWalker:
It seems that the entries in the side-navigation are sorted alphabetically, whereas the entries in the list are sorted in reverse alphabetical order. Maybe it's helpful...
![grafik](https://user-images.githubusercontent.com/45610865/187086964-d0fd60d6-8b26-40e3-b7ed-95cdbf935a15.png)

chalin:
Reopening because this didn't fix the issue for me. (I'll post more details as soon as I can.)

chalin:
This is what happens now: the main list of posts is in ascending order. It used to be in descending order (so that the most recent post was first).

> <img width="500" alt="image" src="https://user-images.githubusercontent.com/4140793/194723699-fc9749f7-18ef-42b6-ac56-fabbf4e25ea7.png">

So #1200 will need to be reverted.

chalin:
I think that the order of the posts in the blog page listing is correct. It'll be tricky to fix in the navbar (that code is quite complicated already). Maybe it's a "bug" in the Hugo sort function? That would require some careful testing, and I don't have time for that atm.

PR #1269, opened 2022-10-08
Revert #1200 to recover desc blog post order

chalin:
- Reverts #1200
- Related to #1197, which this PR leaves unsolved.


chalin:
@geriom - can we make sure that this gets merged before the next release?

chalin:
Merged, as agreed during PSC meeting.

PR #1277, opened 2022-10-12
Update version banner to point to current page

jlian:
This matches the version banner to behave the same as the version selector when `version_menu_pagelinks` is set to true (https://github.com/google/docsy/pull/1054)


fekete-robert:
This sounds great!

PR #1312, opened 2022-11-14
Add OpenTelemetry webpage to examples

pellared:
I think it is worth to add OpenTelemetry webpage as an example.
OpenTelemetry is the second most active CNCF project behind Kubernetes.

CC @chalin @svrnm @cartermp @austinlparker


chalin (approved):
❤️ 

I thought it was there, though maybe I'm recalling OTel appearing in a list from a Docsy quarterly report.

I'll let @LisaFC or @emckean merge this if they agree with the update :).

LisaFC:
No, it's not there already, this is great!

PR #1314, opened 2022-11-17
Add FontAwesome info to logos and images page

LisaFC:
Currently only mention FontAwesome in the context of menus, which has confused some users


LisaFC:
@logofile, does this help?

PR #1316, opened 2022-11-23
Fix: mermaid diagrams with parameters in site config

deining:
This PR fixes this two regressions introduced with 90843e360b:

- build breaks if mermaid diagram(s) are displayed on one or more pages **and** parameters for mermaid are given in the site config.
- parameters for mermaid given in the site config were picked up only if `mermaid.enable=true` was given. With auto-activation of mermaid diagrams in place, parameters should be considered event if `mermaid.enable` is missing in the site config.

This PR also bumps mermaid to the latest released version 9.2.2.


multun (approved):
I can confirm this patch fixes the regression :100:

PR #1320, opened 2022-11-25
prevent navbar text wrapping

multun:
When a navbar item has multiple words, and there are enough navbar item to cause the flexbox to be full at reduced window widths without transitioning through the lower media breakpoint, the navbar link name can wrap, which causes the alignment of all navbar items to break.

This change forbids navbar text wrapping even at bigger screen sizes.

### DIFF
--- a/assets/scss/_nav.scss
+++ b/assets/scss/_nav.scss
@@ -56,6 +56,10 @@
         font-weight: $font-weight-bold;
     }
 
+    .navbar-nav {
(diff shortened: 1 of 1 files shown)


geriom (approved):
Thank you!

geriom:
@multun It seems I can't push changes to your branch. Please rebase to include the latest changes from main so we can merge.

multun:
@geriom done

PR #1351, opened 2023-01-09
Math and chem code clocks, autoactivation: fix use with hugo version < 0.93

deining:
With #1341, autoactivation of `math` and `chem` code blocks was fixed. I now realized that as unwanted side effect, the build now breaks with hugo versions < 0.93. This PR corrects this. It also removes the `katex` from the config file of the use guide. The recommended approach is now to acticate math on a per page basis via `math=true` in the frontmatter of the page.


chalin (approved):
LGTM

deining:
With 2 approvals, I'm going to merge this PR so that it is included in another, related PR that I'm currently preparing.

PR #1357, opened 2023-01-16
Add support for relative links in the footer bar

chriseckhard:
In my use case, I have a developer information page in my Docsy site that is not in the normal, user-focused menu. I want this to be able to linked in the footer with the other developer links, but currently only links to external websites are allowed.

This change passes the site home context to the ```footer-links-block``` and makes it so that any relative links defined in the Site.Params.links are looked up and displayed. It also sets the target to ```_self``` for any reference links so that the page opens in the current window instead of another one.


chalin:
- No review interest accumulated since 2023 and the change predates the restructure. Thanks.

Issue #906, opened 2022-02-24
Use Hugo's native support for processing HTML & markdown, not file extension testing


chalin:
_Originally posted by @deining in https://github.com/google/docsy/issues/526#issuecomment-823577659_ :

> I just realized that shortcode `lead` treats its `.Inner` content either as markdown or as HTML, depending on the extension of the page that makes use of the the shortcode:

```
{{ if eq .Page.File.Ext "md" }}
    {{ .Inner | markdownify }}
{{ else }}
    {{ .Inner | htmlUnescape | safeHTML }}
{{ end }}
```

Does it make sense to implement this for shortcode `feature` in a similar way?

chalin:
Hey @trombonehero - this issue is a proposal to revert #378, which you submitted a little while back. Might you be willing to give some more details concerning your use case, possibly illustrating why docsy -- prior to #378 -- didn't seem to address your needs? Thanks!

PR #1376, opened 2023-01-25
Update to bootstrap 5: adapt shortcode 'cardpane'

### LINKED ISSUE #1158: card-code has excess margin at bottom
Visit https://www.docsy.dev/docs/adding-content/shortcodes/#shortcode-card-code-for-programming-code and you'll see:

> image

There's an excess 32px margin at the bottom of the **C** card-code example.

/cc @deining

### LINKED ISSUE #1375: [BSv5] Card groups are not rendered properly any more
After update to Bootstrap version 5, cards within card groups are displayed side by side any more. The shortcode(s) need to be adapted.

deining:
- Closes #1158, per https://github.com/google/docsy/pull/1376#issuecomment-1407464018
- Closes #1375

Preview: https://deploy-preview-1376--docsydocs.netlify.app/docs/adding-content/shortcodes/#card-panes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ For a list of issues targeted for the next release, see the [23Q1][] milestone.
   - `blocks/section`: **default** and accepted values of the `type` argument
     have changed! For details see [blocks/section] ([#1472]).
 
+- **[Adaptation of shortcode 'cardpane'][1376]. renamed CSS class `td-card-deck`
+  to `td-card-group`.
+
 **Other changes**:
 
 - Non-breaking changes that result from the Bootstrap v5 upgrade:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,13 @@ For a list of issues targeted for the next release, see the [23Q1][] milestone.
   - `blocks/section`: **default** and accepted values of the `type` argument
     have changed! For details see [blocks/section] ([#1472]).
 
-- **[Adaptation of shortcode 'cardpane'][1376]. renamed CSS class `td-card-deck`
-  to `td-card-group`.
+- **[Adaptation of shortcodes for diplay of cards (#1376)][1376]:
+  - shortcode `cardpane`: renamed CSS class `td-card-deck` to `td-card-group`.
+  - shortcode `card`, `card-code`: markup of inner content (html/markdown)
+    now depends on the syntax of the calling shortcode, not on extension
+    of page file any more [#906][906].
+  - shortcode `card-code` is now deprecated, use shortcode `card` with named
+    parameter `code=true` instead.
 
 **Other changes**:
 
--- a/layouts/shortcodes/card-code.html
+++ b/layouts/shortcodes/card-code.html
@@ -1,4 +1,4 @@
-<div class="td-card card">
+<div class="td-card card border me-4">
   {{ $lang := "" }}
   {{ with $.Get "lang" }}
     {{ $lang = $.Get "lang" }}
@@ -12,7 +12,7 @@
       {{- $.Get "header" | markdownify -}}
     </div>
   {{end}}
-  <div class="card-body code p-l0 m-0">
+  <div class="card-body code p-0 m-0">
       {{ with $.Inner }}
         {{- highlight $.Inner $lang $highlight -}}
       {{ end }}
--- a/layouts/shortcodes/card-code.html
+++ b/layouts/shortcodes/card-code.html
@@ -1,20 +1,16 @@
+{{ warnf `shortcode 'card-code' is deprecated: use shortcode 'card' with named parameter 'code=true' instead.` -}}
+{{ $lang := default "" ($.Get "lang") -}}
+{{ $highlight := default "" ($.Get "highlight") -}}
+
 <div class="td-card card border me-4">
-  {{ $lang := "" }}
-  {{ with $.Get "lang" }}
-    {{ $lang = $.Get "lang" }}
-  {{ end }}
-  {{ $highlight := "" }}
-  {{ with $.Get "highlight" }}
-    {{ $highlight = $.Get "highlight" }}
-  {{ end }}
-  {{- with $.Get "header" -}}
+  {{ with $.Get "header" -}}
     <div class="card-header bg-white">
-      {{- $.Get "header" | markdownify -}}
+      {{ . | markdownify -}}
     </div>
-  {{end}}
+  {{ end -}}
   <div class="card-body code p-0 m-0">
-      {{ with $.Inner }}
-        {{- highlight $.Inner $lang $highlight -}}
-      {{ end }}
+    {{ with $.Inner -}}
+      {{ highlight . $lang $highlight -}}
+    {{ end -}}
   </div>
 </div>
--- a/layouts/shortcodes/card.html
+++ b/layouts/shortcodes/card.html
@@ -1,4 +1,4 @@
-<div class="td-card card mb-4">
+<div class="td-card card border me-4">
   {{ with $.Get "header" }}
     <div class="card-header">
       {{ if eq $.Page.File.Ext "md" }}
@@ -19,7 +19,7 @@
       </h5>
     {{ end }}
     {{ with $.Get "subtitle" }}
-      <h6 class="card-title mb-2 text-muted">
+      <h6 class="card-title ms-2 text-muted">
         {{ if eq $.Page.File.Ext "md" }}
           {{ $.Get "subtitle" | markdownify }}
         {{ else }}
--- a/layouts/shortcodes/card.html
+++ b/layouts/shortcodes/card.html
@@ -1,49 +1,47 @@
+{{/* Check parameter type */ -}}
+{{ $code := false -}}
+{{ with .Get "code" -}}
+  {{ $type := printf "%T" . -}}
+  {{ if ne $type "bool" -}}
+    {{ errorf `shortcode 'card': boolean value expected for parameter 'code', but got %s` $type -}}
+  {{ else }}
+    {{ $code = . -}}
+  {{ end -}}
+{{ end -}}
+
+{{ $lang := default "" (.Get "lang") -}}
+{{ $highlight := default "" (.Get "highlight") -}}
+
 <div class="td-card card border me-4">
-  {{ with $.Get "header" }}
-    <div class="card-header">
-      {{ if eq $.Page.File.Ext "md" }}
-        {{ $.Get "header" | markdownify }}
-      {{ else }}
-        {{ $.Get "header" | htmlUnescape | safeHTML }}
-      {{ end }}
+{{ with $.Get "header" -}}
+    <div class="card-header{{- cond $code " bg-white" "" -}}">
+      {{ . | markdownify }}
     </div>
-  {{ end }}
-  <div class="card-body">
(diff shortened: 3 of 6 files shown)


chalin:
Preview: https://deploy-preview-1376--docsydocs.netlify.app/docs/adding-content/shortcodes/#card-panes

chalin (commented):
Could you add an entry to the CHANGELOG, under breaking-changes so that users are aware of the class name change?

chalin:
This PR improves the rendering of card panes, but note that this change still doesn't bring the concerned page in visual agreement with what the paged looked like before.

Before (https://deploy-preview-1356--docsydocs.netlify.app/docs/adding-content/shortcodes/#card-panes):

> image

With this PR (https://deploy-preview-1376--docsydocs.netlify.app/docs/adding-content/shortcodes/#card-panes):

> image

deining:
> Could you add an entry to the CHANGELOG, under breaking-changes so that users are aware of the class name change?

Done.

deining:
> This PR improves the rendering of card panes, but note that this change still doesn't bring the concerned page in visual agreement with what the paged looked like before.

I was aware of this when submitting the PR.

From the [bootstrap migration guide](ttps://getbootstrap.com/docs/5.0/migration/#card)

```
BREAKING: Dropped .card-deck in favor of our grid.
Wrap your cards in column classes and add a parent .row-cols-* container
to recreate card decks (but with more control over responsive alignment).
```

Using of the bootstrap grid in the `cardpane` shortcode would mean a major rework of the shortcode. I fiddled around with adding a margin/padding to the individual cards, with no satisfactory results. Yes, the page looks different, but it doesn't look bad IMHO. I would leave this as is for now.

deining:
> This PR improves the rendering of card panes, but note that this change still doesn't bring the concerned page in visual agreement with what the paged looked like before.
 
On a second attempt, I now was able to bring the concerned page in visual agreement with what the page looked before.
The PR now also closes #1158.

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ For a list of issues targeted for the next release, see the [23Q1][] milestone.
     - Dropped `$primary-light`.
     - Dropped `color-diff()`.
 
+- **[Adaptation of shortcode 'cardpane'][1376]. With the update to Bootstrap 5,
+  the class name used in the `cardpane` shortcode and the associated css file
+  had to be changed from `card-deck` to `card-group`.
```
chalin, line 26:
```suggestion
- **[Adaptation of shortcode 'cardpane'][1376]: renamed CSS class `td-card-deck` to `td-card-group`.
```

chalin (commented):
See inline comment.

Also pls rebase, then I'll do a final review.

CHANGELOG.md
deining: Changed as proposed.

deining:
> Also pls rebase,

Done.

> then I'll do a final review.

Thanks.

deining:
Following up on #906, I reworked the `card` shortcode by adding a second commit to this PR. This commit introduces the following changes:

-  shortcode `card`, `card-code`: markup of inner content (html/markdown) now depends on the syntax of the calling shortcode, not on extension of page file any more.
- deprecated shortcode `card-code`, you can now use shortcode `card` with named parameter `code=true` instead.
- Simplified and improved code of shortcodes.

chalin (commented):
Is `layouts/shortcodes/card-code.html` actually still used? It seems to me that this PR was getting rid of it, but the file is still present.

layouts/shortcodes/card-code.html
```diff
@@ -1,20 +1,15 @@
...
-  {{ end }}
-  {{ $highlight := "" }}
-  {{ with $.Get "highlight" }}
-    {{ $highlight = $.Get "highlight" }}
-  {{ end }}
-  {{- with $.Get "header" -}}
+{{ $lang := default "" ($.Get "lang") -}}
+{{ $highlight := default "" ($.Get "highlight") -}}
+
+<div class="td-card card border me-4">
```
chalin, line 4: I thought that this shortcode was being dropped?

If not, then I'd suggest adding any custom styling through `cards-pace.scss`; avoid the `border` class here, and even consider moving the `me-4` into, say, `.td-card.card`.

layouts/shortcodes/card.html
```diff
@@ -1,49 +1,44 @@
...
+  {{ if ne ( printf "%T" . ) "bool" -}}
+    {{ errorf "shortcode card: parameter 'code' must be either true or false" -}}
+  {{ end -}}
+{{ end -}}
+
+{{ $code := default false (.Get "code") -}}
+{{ $lang := default "" (.Get "lang") -}}
+{{ $highlight := default "" (.Get "highlight") -}}
+
+<div class="td-card card border me-4">
```
chalin, line 12: As I mentioned before, I'd suggest adding custom styling through `cards-pace.scss`; avoid the `border` class here, and even consider moving the `me-4` into, say, `.td-card.card`.

layouts/shortcodes/card.html
```diff
@@ -1,49 +1,44 @@
...
-        {{ $.Get "header" | markdownify }}
-      {{ else }}
-        {{ $.Get "header" | htmlUnescape | safeHTML }}
-      {{ end }}
+{{/* Check parameter type */ -}}
+{{ with .Get "code" -}}
+  {{ if ne ( printf "%T" . ) "bool" -}}
+    {{ errorf "shortcode card: parameter 'code' must be either true or false" -}}
+  {{ end -}}
+{{ end -}}
```
chalin, line 6: Is the type check really necessary? Why not interpret the value simply as truthy / falsy?

layouts/shortcodes/card.html
```diff
@@ -1,49 +1,44 @@
...
+
+{{ $code := default false (.Get "code") -}}
+{{ $lang := default "" (.Get "lang") -}}
+{{ $highlight := default "" (.Get "highlight") -}}
+
+<div class="td-card card border me-4">
+{{ with $.Get "header" -}}
+    <div class="card-header{{- cond $code " bg-white" "" -}}">
```
chalin, line 14: Here's another place where we could avoid hard-coding custom styling, but for now I'm ok if we leave this as is, and address it later.

chalin (commented):
Other than the question about `card-code` and border, LGTM. Thanks!

layouts/shortcodes/card.html
deining: This type check is here for a good reason: the correct syntax is `<card code=true>`. If you write `<card code="true">`, an error ist thrown. What happens if you omit the check? See shortcode `readfile` which has no type check (yet!?):

https://github.com/google/docsy/blob/3eb2cd0cfb26521d94cc8503e0092cc4a5c40cc2/layouts/shortcodes/readfile.html#L31

Users are expected to write `<readfile code="true">`, if they write `<readfile code=true>`, nothing happens. This will be surprising for users and I doubt that all users immediately see what they did wrong (giving parameter value `true` instead of `"true"`).

Does that make clear why I added the type check? Does this answer your question and resolve the issue?

**Additional note**: IMHO shortcode `readfile` does things wrong here: it treats parameter `code` as string and tests for string equality while the parameter `code` has boolean character and should be treated like that in the shortcode.

deining:
> Is `layouts/shortcodes/card-code.html` actually still used? 

No, in the userguide, we don't use it any more.

> It seems to me that this PR was getting rid of it, but the file is still present.

CHANGELOG:

```
- shortcode `card-code` is now deprecated, use shortcode `card` with named
    parameter `code=true` instead.
```

I wrote `deprecated`, which means the file is still present and works, but it is not documented any more and users are discouraged to further use this shortcode.

> but the file is still present.

Yes, this is intentional: if I remove the shortcode, this will break all existing pages that make use of this shortcode and we will force users to adapt their pages (by replacing shortcode `card-code` with `card`).

Does that clarify things? Would you like to see the file deleted?

layouts/shortcodes/card-code.html
deining: > I thought that this shortcode was being dropped?

See my answer in this [commen](https://github.com/google/docsy/pull/1376#issuecomment-1429333570)

layouts/shortcodes/card.html
deining: > Here's another place where we could avoid hard-coding custom styling,

I agree.

> but for now I'm ok if we leave this as is, and address it later.

I will open a new issue that addresses this issue shortly.

layouts/shortcodes/card.html
deining: > As I mentioned before, I'd suggest adding custom styling through `cards-pace.scss`; avoid the `border` class here, and even consider moving the `me-4` into, say, `.td-card.card`.

I agree. I will open a new issue that addresses this issue shortly.

layouts/shortcodes/card-code.html
deining: > If not, then I'd suggest adding any custom styling through `cards-pace.scss`; avoid the `border` class here, and even consider moving the `me-4` into, say, `.td-card.card`.

I agree. I will open a new issue that addresses this issue shortly.

deining:
> Other than the question about `card-code` and border, LGTM. Thanks!

I rebased the PR to get it ready for merging. Let me know if you want me to delete the `card-code` shortcode. Styling issues will be addressed in a separate issue, to be opened shortly. Thanks for your thorough review, I appreciate your comments.

layouts/shortcodes/card.html
chalin: Yes, that makes sense. In that case, the error message could be clearer, stating that a boolean value is expected, but another type was given. E.g.,

> {{ errorf "shortcode card: boolean value expected for parameter 'code', but got type %s" ( printf "%T" . ) -}}

You can even extract the value of `( printf "%T" . )` earlier and reuse it in both places.

layouts/shortcodes/card-code.html
```diff
@@ -1,20 +1,15 @@
...
-  {{ $lang := "" }}
-  {{ with $.Get "lang" }}
-    {{ $lang = $.Get "lang" }}
-  {{ end }}
-  {{ $highlight := "" }}
-  {{ with $.Get "highlight" }}
-    {{ $highlight = $.Get "highlight" }}
-  {{ end }}
-  {{- with $.Get "header" -}}
+{{ $lang := default "" ($.Get "lang") -}}
```
chalin, line 1: Consider adding at least a comment to the effect that this is deprecated.

I had missed the deprecation note int the changelog, so others might miss it too. I propose that you add a warning stating that it is deprecated and will be removed in a future release. This way projects using the shortcode will get notified when they build their site. WDYT?

chalin (commented):
Thanks for your clarifications. See inline comments.

layouts/shortcodes/card-code.html
deining: > Consider adding at least a comment to the effect that this is deprecated.

Great idea, I will do that shortly.

deining: Done.

layouts/shortcodes/card.html
deining: Improved error message as requested.

chalin (approved):
I've rebased and resolved the minor conflicts. Thanks for the progressive improvements to this PR. 

LGTM ✨

PR #1384, opened 2023-01-31
[BSv5] replace lighten by tint-color, and var/func clean up

chalin:
- Contributes to #470
- Replaces custom coloring of Draw.io edit button by outline-primary:
  - Before: image
  - After: see below
  - Reference: https://www.docsy.dev/docs/adding-content/diagrams-and-formulae/#diagrams-with-diagramsnet
- Drops (now) unused, elements (breaking changes). I've scanned all of the Docsy-based projects that I work on, and none use these elements:
  - `$primary-light`
  - `color-diff()`

### Draw.io edit-button screenshot

> Screen Shot 2023-01-31 at 10 12 48


LisaFC:
I think the button change is fine, I can't imagine anyone's look and feel being absolutely wedded to having a red edit button, and the new version will match the surrounding site.

Have we checked any of our contributed examples for those elements?

chalin:
> Have we checked any of our contributed examples for those elements?

I've sampled the projects listed under [Examples](https://www.docsy.dev/docs/examples/) and none use these two SCSS elements. Of course, any project that might happen to use these elements can define their own after this PR is merged -- e.g., by copying the definitions that Docsy gave. The point that we're making here in dropping these elements is to recognize that they aren't a necessary part of Docsy (they weren't even mentioned in the User Guide).

LisaFC (approved):

chalin:
Possible followup:

- #1385

Issue #1387, opened 2023-02-01
Simplify how code color is defined


chalin:
This is a followup to

- #470 

Here's how `<code>` element color is defined in Docsy currently:

1. **[assets/scss/_variables.scss](https://github.com/google/docsy/blob/main/assets/scss/_variables.scss) overrides the Bootstrap `$code-color` variable**. The BS `$code-color` default is pink (unlikely to be a useful color), but Docy's override of using the secondary color as a base isn't much more helpful (at least not in projects I've worked on).
2. Docsy has context-specific overrides setting code color to `inherit`, which usually means it'll fall back to `$body-color` (a reasonable default IMHO):
  - `.td-content p code`
  - `.td-content li > code`
  -  `.td-content table code`
3. But some contexts fall back to `$code-color`,  such as the `<dd>` in [Shortcode blocks](http://localhost:1313/docs/adding-content/shortcodes/#shortcode-blocks).

On almost all projects that I've worked on, a single code color is used throughout the project, and often it matches the base "body color". So, IMHO, we should set `$code-color` to be equivalent to `$body-color`[^1] and use this in all contexts (i.e., avoid contextual overrides).

Thoughts?

[^1]: We'll probably need to set it to `var(--bs-body-color)` since BS's `$body-color` won't be defined by the time we choose to set the value of `$code-color`.

PR #1388, opened 2023-02-01
[BSv5] Fix SCSS functions import issue, and replace darken()

chalin:
⚠️ See below for an **IMPORTANT DESIGN NOTE**!

This PR:

- Contributes to #470
- Imports Bootstrap's `functions` first! For details, see the design note below.
- Replaces `darken()` by other functions, getting some guidance from:
  - [Bootstrap did it][bs-color-sys] ([#30622])
  - [Sass `darken()` documentation][sass-darken] , which describes suitable replacements

Followup:

- #1387

---

### Is this fix really necessary?

Without the import fix included in this PR, use of Bootstrap `functions` can result in bogus errors like the following:

> argument `$color` of `<some-bs-function>(...)` must be a color

when in fact the argument _is_ a color.

### Learn more

For details concerning the need to import `functions.scss` first, see the Bootstrap migration guide under [New _maps.scss][bs-new-maps]

### Design choice: double import vs. inlining many imports

In this PR, I'm proposing to `@import "functions"` before our variable overrides, and then import the full Bootstrap suite of SCSS. This will result in the "functions" file being loaded twice, but according to the [Sass `@import` documentation](https://sass-lang.com/documentation/at-rules/import):

> If the same stylesheet is imported more than once, it will be evaluated again each time. If it just defines functions and mixins, this usually isn’t a big deal, but if it contains style rules they’ll be compiled to CSS more than once.

The [_functions.scss](https://github.com/twbs/bootstrap/blob/main/scss/_functions.scss) file doesn't contain any style rules, so we should be ok. The alternative would be to inline all of the 40+ imports in https://github.com/twbs/bootstrap/blob/main/scss/bootstrap.scss, which would be a significant maintenance overhead, one that I think that we can avoid (until proven otherwise).

[#30622]: https://github.com/twbs/bootstrap/pull/30622
[bs-color-sys]: https://getbootstrap.com/docs/5.3/migration/#color-system
[bs-new-maps]: https://getbootstrap.com/docs/5.3/migration/#new-_mapsscss
[sass-darken]: https://sass-lang.com/documentation/modules/color#darken

### DIFF
--- a/assets/scss/_variables.scss
+++ b/assets/scss/_variables.scss
@@ -38,7 +38,7 @@ $gray-800: #333 !default;
 $gray-900: #222 !default;
 $black: #000 !default;
 
-$code-color: darken($secondary, 20%) !default;
+$code-color: shade-color($secondary, 40%) !default;
 
 // UI element colors
 
@@ -52,9 +52,9 @@ $td-sidebar-border-color: $border-color !default;
 // These colors are all part of the theme palette, but the mix is fairly random to create variation. This can be overridden by the project if needed.
 $td-box-colors: $dark, $primary, $secondary, $info, $gray-600, $success, $warning, $dark, $danger, $primary, $secondary, $info !default;
 
-$link-color: darken($blue, 15%) !default;
+$link-color: adjust-color($blue, $lightness: -15%) !default;
 $link-decoration: none !default;
-$link-hover-color: darken($link-color, 15%) !default;
+$link-hover-color: shade-color($link-color, 30%) !default;
 $link-hover-decoration: none !default;
 
 // Fonts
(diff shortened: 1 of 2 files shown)


assets/scss/main.scss
```diff
@@ -1,3 +1,5 @@
+@import "../vendor/bootstrap/scss/functions";
```
chalin, line 1: See the opening comment of this PR for a design rationale.

assets/scss/_variables.scss
```diff
@@ -52,9 +52,9 @@ $td-sidebar-border-color: $border-color !default;
 // These colors are all part of the theme palette, but the mix is fairly random to create variation. This can be overridden by the project if needed.
 $td-box-colors: $dark, $primary, $secondary, $info, $gray-600, $success, $warning, $dark, $danger, $primary, $secondary, $info !default;
 
-$link-color: darken($blue, 15%) !default;
+$link-color: adjust-color($blue, $lightness: -15%) !default;
```
chalin, line 55: While I followed BS's upgrade path and generally replaced `darken()` by `shade-color()`, in this case the resulting color didn't look as good, so I opted for a call to `adjust-color()` --matching the previous color exactly.

LisaFC:
Makes sense to me! I don't see any obvious (or easy to maintain) alternative to this approach.

LisaFC (approved):

Issue #939, opened 2022-03-21
Shortcodes: drop support for pre-Hugo-0.54.x behavior of `{{% %}}`


chalin:
Here is some context:

- #87 

While the change introduced by that PR was relevant at the time (during a transition period from Hugo 0.54.x to 0.55+), we now only officially support the latest version of Hugo. So, all shortcodes making use of

```
{{ $_hugo_config := `{ "version": 1 }` }}
```

should have that line removed, and the shortcode adjusted, if necessary.

<details><summary>File list</summary>

```console
$ find . -type f -not -path "*/node_modules/*" -exec grep -le '_hugo_config' {} \;
./layouts/shortcodes/alert.html
./layouts/shortcodes/blocks/cover.html
./layouts/shortcodes/blocks/section.html
./layouts/shortcodes/blocks/lead.html
./layouts/shortcodes/pageinfo.html
```

</details> 

### Tasks

- [x] `layouts/shortcodes/alert.html` - #941
- [x] `layouts/shortcodes/blocks/cover.html` - #2480
- [x] `layouts/shortcodes/blocks/section.html` - #1472
- [x] `layouts/shortcodes/blocks/lead.html` - #1398
- [x] `layouts/shortcodes/pageinfo.html` - #2480

PR #1400, opened 2023-02-10
shortcode 'blocks/cover': improvements

deining:
This PR is part of #906. It contains improvements for the shortcode `blocks/cover`:

* the shortcode now uses Hugo's native support for processing HTML & markdown (#906). This was done by adding three comment lines in oder to fulfill the `end` condition of the html block containing the shortcode body (as defined in the CommonMark [spec](https://spec.commonmark.org/0.30/#html-blocks), number 7).
* the shortcode now throws an error now if parameter `height` contains an invalid value
* support for pre-Hugo-0.54.x behavior of {{% %}} was dropped (#939)

The PR also brings a few improvements to the section of the user guide explaining the use of the shortcode.


chalin:
Thanks for the PR, but I decided to go with a simpler solution for now:

- #2480

PR #1404, opened 2023-02-11
Rename config.yaml to hugo.yaml

chalin:
For details, see https://github.com/gohugoio/hugo/releases/tag/v0.110.0.


deining:
Two comments/thoughts on this:

- with that change in place, we should follow up and rename `dependencies/config.toml` to `dependencies/hugo.toml` as well.
- [hugo userdoc]() states:

```
We will still look for config.toml etc., but we recommend you eventually rename it
(but you need to wait if you want to support older Hugo versions)
```

In the current form of the repo, e.g. userguide runs with latest `hugo` version 0.110.0 only. Shouldn't we add a copy of `hugo.toml` as `config.toml`, thus making this a backwards compatible change?

@chalin: WDYT?

chalin:
Ah, right. _Officially_ Docsy only supports the latest version of Hugo, but I'm certainly willing to bring this up during Monday's PSC meeting. Thanks for the feedback!

chalin:
Also, I think that folks using an older version of Hugo will be able to use Hugo's `--config` flag to get around this (i.e., to pick up Docsy's `hugo.*` config).

Can you test this @deining?

deining:
> Also, I think that folks using an older version of Hugo will be able to use Hugo's `--config` flag to get around this (i.e., to pick up Docsy's `hugo.*` config).
> 
> Can you test this @deining?

I had this idea, too, and no, I wasn't able to make older version run just by using the `--config` flag. The only way to serve the userguide with hugo versions <0.110 (with current HEAD of repo work) was:

- _copy_ `userguide/hugo.yaml` to `userguide/config.yaml`
- _copy_ `hugo.yaml` to `config.yaml`

For whatever reason, sym-linking these files wasn't suffiient, so I had to copy **both** files.

PR #1415, opened 2023-02-13
[BS5] Adjust HTML of dropdowns for version and lang

chalin:
- Contributes to #470
- Updates the HTML generated for the nav versions and languages dropdown. The updates were made according to ttps://getbootstrap.com/docs/5.3/components/dropdowns/

### Screenshot (with a versions menu added)

> image


chalin:
Merged, but as always, feedback is welcome @deining et al.

PR #1429, opened 2023-02-16
[BSv5] Adjusted offline search popover

tekezo:
- Closes #1425
- Stop using `$targetSearchInput.popover`.
- Move popover options to `new bootstrap.Popover` constructor from `$searchInput.data`.
- Use `customClass` option to add `td-offline-search-results ` class instead of using template option.
- Fixed an issue that  `shown.bs.popover` is called more times than required when popover is shown multiple times.

Screenshot 2023-02-17 at 0 42 22

PR #1434, opened 2023-02-22
Search resource cleanup and style tweak

chalin:
- Followup to #1429
- Runs Prettier over `js/*search.js` assets
  - Diff best viewed by ignoring whitespace changes
- Makes search input field 100% wide -- we need as wide as we can get it to be. 

### Before

> image

### After

> image

> image

### DIFF
--- a/assets/scss/_search.scss
+++ b/assets/scss/_search.scss
@@ -3,7 +3,7 @@
 .td-search {
     background: transparent;
     position: relative;
-    width: 90%;
+    width: 100%;
 
     // Search icon
     &__icon {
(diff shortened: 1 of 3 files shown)


assets/scss/_search.scss
```diff
@@ -3,7 +3,7 @@
 .td-search {
     background: transparent;
     position: relative;
-    width: 90%;
```
chalin, line 6: FYI: I think that this 90% was an artefact of the many iterations we did while attempting to redesign the search input box a little while back.

PR #1439, opened 2023-02-24
Pin Docsy version in other installation options

chalin:
- Updates instructions for "other installation options" so that the version of Docsy is pinned to the latest official release (as we've done for the main installation option).
- This is in response to #1421

**Preview**: e.g., see https://deploy-preview-1439--docsydocs.netlify.app/docs/get-started/other-options/#option-3-docsy-as-an-npm-package

/cc @NatasaKralj

### DIFF
--- a/userguide/content/en/docs/get-started/other-options.md
+++ b/userguide/content/en/docs/get-started/other-options.md
@@ -161,6 +161,15 @@ your project's root directory:
 1.  Install Docsy as a Git submodule:
 
     ```sh
+    git submodule add https://github.com/google/docsy.git themes/docsy
+    cd themes/docsy
+    git checkout v{{% param version %}}
+    ```
+
+    To work from the development version of Docsy (not recommended),
+    run the following command instead:
+
+    ```sh
     git submodule add --depth 1 https://github.com/google/docsy.git themes/docsy
     ```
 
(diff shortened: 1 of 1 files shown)


userguide/content/en/docs/get-started/other-options.md
```diff
@@ -220,7 +229,7 @@ from your project's root directory:
 
 ```sh
 cd themes
-git clone https://github.com/google/docsy
+git clone -b v{{% param version %}}  https://github.com/google/docsy
```
deining, line 232: In generea, I like the idea. However, if a user follows the instruction given, he ends up with a detached state of his repo:

```
$ git clone -b v0.6.0  https://github.com/google/docsy
Cloning into 'docsy'...
remote: Enumerating objects: 9500, done.
remote: Counting objects: 100% (92/92), done.
remote: Compressing objects: 100% (51/51), done.
remote: Total 9500 (delta 42), reused 77 (delta 35), pack-reused 9408
Receiving objects: 100% (9500/9500), 8.70 MiB | 8.27 MiB/s, done.
Resolving deltas: 100% (5940/5940), done.
Note: switching to '5597d435dc74ce68240e0c3871addf24567493b0'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false
```

Detached state adds extra complexity to the process and I'm not sure if all users can handle that. If we would like to keep this command in the current form, we should add least add this to the the checkout sequence given:

```
$ git switch -c my-new-branch
Switched to a new branch 'my-new-branch'
```

chalin: Thanks for the feedback. IMHO, cloning is for Docy "power users". It feels inadvisable to me to instruct _all_ users to create a branch -- when they might not need one.

I've slightly adjusted the text to explain how to clone from `main@HEAD`.

PR #1440, opened 2023-02-24
NPM package config: set Node version constraint

chalin:
This formally encode the Docsy project constraint that we officially support only the active LTS version of Node: https://github.com/google/docsy/blob/6f796f91dfabfbc26e3cfc1587a501a43503c686/.nvmrc#L1

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,5 +28,8 @@
   },
   "devDependencies": {
     "hugo-extended": "0.110.0"
+  },
+  "engines": {
+    "node": ">=18"
   }
 }


deining (approved):
LGTM.

In this context, I would propose to also add (and pin) the node version in `netlify.toml`, thus getting more reproducible builds:

```
[build.environment]
NODE_VERSION = "18.14.2"
GO_VERSION = "1.19.5"
HUGO_THEME = "repo"
```

chalin:
It is already pinned via https://github.com/google/docsy/blob/main/.nvmrc, which is better than adding a constraint in the Netlify config, as pointed out in https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript:

> image

Let's keep things as dry as we can. The `engines.node` constraint is a lower bound -- if I could, I'd put the equivalent of `lts/*` but that doesn't seem possible. As you know we've had problems in the past with folks reporting issues that arose because they were using a version of Node that was far too old. This will help improve on that scenario.

chalin:
Thanks for the feedback @deining. I'll merge this as is for now.

PR #1172, opened 2022-08-15
Check links in the user guide

chalin:
- Closes #1089
- Will check links after preview and production builds
- To check links locally: `npm run check-links`

This isn't my ideal setup because I'd like to avoid the makefile entirely, but it is good enough -- and this uses one of the fastest link checkers I know!

PR #1441, opened 2023-02-24
Adding GitHub action for link check of user guide

deining:
With #1172, link checking in the user guide was introduced, thanks for that @chalin. A [comment](https://github.com/google/docsy/pull/1172#issue-1339362872) given in this PR states:

```
This isn't my ideal setup because I'd like to avoid the makefile entirely
```

This PR follows up on this comment and proposes an approach that allows to eleminate `userguide/Makefile` introduced with #1172.
This is done by adding [GitHub action] that makes use of the [htmltest] link checker to check the links of the user guide.
In its current form, this link check action will be performed on each push to the `main` branch of the `docsy` repo (this can be adapted, of course). To see the results of the last workflow run(s), select the `htmltest` workflow in the `Actions` [section](../actions/workflows/htmlcheck.yml) of the GitHub web UI. On the same page, you can also disable the workflow if needed. To do so, press the button `…` right beneath the search field that allows you to filter workflow runs.

This PR added two URLs to the `IgnoredURLs` setting in the config file of the link checker. With this adapation, the link checker currently reports back green status for the link check:

grafik

I hope you like this PR. If opens up several possibilities to streamline the workflow e.g. for pull requests.

**Note:**: In order to allow the Github action workflow to cache npm dependencies, `packages.json.lock` was added to the repo.

[GitHub action]: https://docs.github.com/en/actions
[htmltest]: https://github.com/wjdp/htmltest

### DIFF
--- a/.github/workflows/htmltest.yml
+++ b/.github/workflows/htmltest.yml
@@ -1,0 +1,59 @@
+# .github/workflows/htmltest.yml
+# Run htmltest link checker on generated HTML output from user guide
+
+name: Link check with htmltest
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+    - main
+  pull_request:
+
+jobs:
+  htmltest:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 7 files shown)


chalin:
- CI link checking has since been reworked (Lychee-based, monorepo CI); an htmltest action no longer applies. Thanks for the contribution.

Issue #1442, opened 2023-02-25
Reintroduce RTL support using BSv5 RTLCSS


chalin:
In our move to Bootstrap v5 (BSv5), we are adopting the Bootstrap approach to RTL, as described in https://getbootstrap.com/docs/5.3/getting-started/rtl/. The good news is that we don't have to rename all properties that have [logical][] counterparts (at least for now), since [RTLCSS] handles their [translation appropriately].

### 24H2 work

- [x] Starting over with the strategy of using a single "main" CSS file per page, using an adapted RTL CSS for RTL pages:
    #2093
- [x] Review of 24H1 PRs that might need to be totally or partially reverted:
  - [x] #1443 - nothing to revert
  - [x] #2002 - ok except for `layouts/partials/head-css.html`, which is being reworked.
    Reverted via #2093
  - [x] #2020 & #2021 - nothing to revert
  - [x] #2018 - reverted via #2093
  - [x] #2025 - reverted via #2096
- [x] Create an `rtl` branch to be used during development
    **Branch preview**: https://rtl--docsydocs.netlify.app/
- [x] On `rtl` branch, set the `en` language as `rtl` for testing purposes - #2095
- [x] Old 24H1 issues that go away with #2093:
  - [x] #2003
  - [x] #2005
- [x] Test it out on the Docsy example
- [x] Cleanup - #2097
- [x] Update the UG - #2098

### 24H1 work - some of it reverted in H2

Entries processed (either ok as is, or reverted in whole or part), are marked with ⤴.

- [x] ⤴ Drop import of custom Docsy RTL styling, which was designed for Bootstrap v4 - #1443 
- [x] ⤴ #2002, fixing
    #2004
- [x] ⤴ #2020
    #2021
- [x] ⤴ #2018
  - [x] #829
  - [x] #2011
- [x] ⤴ #2025

---

Related:

- https://github.com/kubernetes/website/issues/45077
- https://github.com/kubernetes/website/issues/22730

Notes:

- Context: #470
- Would a switch to Bootstrap's RTL solution help address the following as well:
  https://github.com/matrix-org/docsy/issues/1? 
  Related(?): #605

Other:

- https://github.com/twbs/bootstrap/blob/cf9454caa00872899215603e5e036d9a824b1b11/package.json#L46

[logical]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties
[RTLCSS]: https://rtlcss.com
[...]

PR #1443, opened 2023-02-25
[BSv5] Drop import of rtl/main

chalin:
- As is mentioned in the changelog, this drops the BSv5-incompatible RTL support we had before. We drop the `rtl/main` import from the main SCSS file, but keep the `rtl` style files for now.
- Contributes to #1442, which we'll use to track progress of RTL-support reintroduction
- View diff by ignoring whitespace changes


geriom (approved):
Thanks!

PR #1451, opened 2023-03-01
Use <pre> tag for mermaid code blocks

### LINKED ISSUE #1450: Mermaid diagrams fail to render
Since the following change has been applied, some previously working Mermaid diagrams report "syntax error":

- #990

But nothing has changed in the diagrams and the live Mermaid editor are fine with the Mermaid code.

I believe that the problem stems from the decision to render the code block as a `<div>` rather than `<pre>`: https://github.com/google/docsy/blob/26d4d117a863fb55be71b624256cfd37f501c27f/layouts/_default/_markup/render-codeblock-mermaid.html#L3-L5

Ok, after some digging, I've found that `<pre>` is _required, see [Requirements for the Mermaid API](https://mermaid.js.org/intro/n00b-gettingStarted.html#requirements-for-the-mermaid-api):

> image

References:

- https://github.com/google/docsy/pull/990/files#r997232082
- Screenshot when rendering fails for page https://opentelemetry.io/docs/demo/architecture/:
  > image

/cc @stephanlachnit @deining

deining:
Right now a mermaid code block is enclosed in `<div>` tags. With this PR applied we now use `<pre>` tags instead, as suggested and discussed in #1450. 

**Note**: We are using `<pre>`tag for mermaid code blocks already here, so this PR makes behaviour consistent in all places.

https://github.com/google/docsy/blob/26d4d117a863fb55be71b624256cfd37f501c27f/assets/js/mermaid.js#L9-L12

This PR closes #1450.

---

**Preview**: https://deploy-preview-1451--docsydocs.netlify.app/docs/adding-content/diagrams-and-formulae/#diagrams-with-mermaid

### DIFF
--- a/layouts/_default/_markup/render-codeblock-mermaid.html
+++ b/layouts/_default/_markup/render-codeblock-mermaid.html
@@ -1,5 +1,5 @@
 {{ .Page.Store.Set "hasmermaid" true -}}
 
-<div class="mermaid">
-  {{- .Inner | safeHTML }}
-</div>
+<pre class="mermaid">
+  {{- .Inner }}
+</pre>


chalin (commented):
Thanks for the PR. As noted in #1450, according to my tests the pipe into `safeHTML` needs to be dropped as well.

I'm ok accepting this change as is, but it should be marked as contributing to #1450, not closing it.

deining:
> As noted in #1450, according to my tests the pipe into `safeHTML` needs to be dropped as well.

Oops I missed that. `safeHTML`  is now removed.

chalin (approved):
Thanks again for this fix!

chalin:
Some style adjustments will be necessary following this change. I'll file a separate issue or submit a PR.

chalin:
Followup:

- #1453

PR #1452, opened 2023-03-02
[BSv5] Drop redundant setting of BS `$enable-rounded` variable

chalin:
- Contributes to #470 (badges)
- Drops the redundant setting of `$enable-rounded` since we're using the BS default value, see https://getbootstrap.com/docs/5.3/customize/options/.

### DIFF
--- a/assets/scss/_variables.scss
+++ b/assets/scss/_variables.scss
@@ -1,17 +1,7 @@
-/*
+// Bootstrap options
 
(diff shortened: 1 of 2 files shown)


chalin:
@LisaFC - will merge now. Post-merge feedback is always welcome 😄.

LisaFC:
Sorry, just got to this now - looks fine to me!

PR #1454, opened 2023-03-02
Reformat SCSS sources using Prettier

chalin:
- Contributes to #470
- This will make reviewing SCSS file diffs easier. I should have done this from the start of the BSv5 effort -- better now than later 😄 
- Runs Prettier over all SCSS sources


chalin:
Sampled pages. It's all looking good.

PR #1455, opened 2023-03-02
[BSv5] Drop `bg-gradient-variant()` mixin

### LINKED ISSUE #1369: [BSv5] Drop `bg-gradient-variant()` mixin
Bootstrap 5 has dropped the `bg-gradient-variant()` mixin -- for details, see the [corresponding entry in the migration guide](https://getbootstrap.com/docs/5.0/migration/#sass).

To avoid this change, we've kept a custom version of the mixin over time. For example, see:

- #328

We should follow BSv5's lead and drop this mixin in favor of the the approach recommended by BS5 in their migration guide.

I'm marking this as a BSv5-migration followup so that progress can be made to address breaking changes.

chalin:
- Closes #1369
- Contributes to #470
- Eliminates use of `td-box--gradient` from layouts: projects can customize the corresponding `td-box-*` styles instead.
- Adds entry to change log
- Cleans up affected layouts by removing unnecessary whitespace

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,23 +14,24 @@ For a list of issues targeted for the next release, see the [23Q1][] milestone.
 
 **Breaking changes**:
 
-- **[Upgraded Bootstrap (#470)][470]** to v5. For a list of Bootstrap breaking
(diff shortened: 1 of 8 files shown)


LisaFC (approved):
Looks great!

PR #1456, opened 2023-03-02
 Adjust Mermaid block styling and block content

### LINKED ISSUE #1453: Adjust Mermaid block styling and block content
- Context: #1451
- As a `<div>` the Mermaid code block has no extra styling.
  Now as a `<pre>` it does -- extra padding all around, in particular -- since it applies to all td pre elements.
- We should also avoid introducing extra whitespace the `<pre>` element: use `{{- .Inner -}}`

deining:
This PR addresses and closes #1453.

### DIFF
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -81,5 +81,6 @@
   pre.mermaid {
     background-color: inherit;
     font-size: 0;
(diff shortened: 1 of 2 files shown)


chalin (approved):
LGTM, Thanks!

PR #1462, opened 2023-03-07
[BSv5] Docsy variables cleanup and BS version test

chalin:
- Checks that BSv5 is being used by a project, otherwise an error is reported by the SASS compiler.
- Cleans out `_variables.scss`:
  - Factors out **forward declarations** of BS SCSS variables into their own file, so that it is clear that they _are_ forward declarations: `_variables_forward.scss`. Eventually, the content of this file will be auto-generated to avoid definition mismatches.
  - In the process of the refactoring previously mentioned, I dropped unnecessary redeclarations of BS SCSS variables (of the same value).
- Switches to using the "BSv5-way" of customizing, e.g.,:
  - Now making use of `$link-shade-percentage`, rather than overriding `$link-hover-color`
- Switches to an official BSv5 `$blue-*` color variable for the link color rather than use a custom adjusted color.
- Marks some Docsy variable declarations as candidates for namespacing and/or moving into the UG -- I'll open an issue to follow up on this.

### DIFF
--- a/assets/scss/_variables.scss
+++ b/assets/scss/_variables.scss
@@ -5,28 +5,17 @@ $enable-shadows: true !default;
 
 // Theme colors
 
+// TODO: consider moving all of these theme-color definitions into the User
+// Guide SCSS, or declare theme here, but namespaced:
+
 $primary: #30638e !default;
 $secondary: #ffa630 !default;
 $success: #3772ff !default;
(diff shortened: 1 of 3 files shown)


chalin:
For the record, this PR introduced an error, that was fixed via:

- #1631

PR #1465, opened 2023-03-08
Avoid warning 'calling IsSet with unsupported type "invalid" (<nil>)'

### LINKED ISSUE #1464: Confusing warning 'calling IsSet with unsupported type "invalid" (<nil>)'
**How to reproduce:**

Remove these lines from `userguide/hugo.yaml`:

https://github.com/google/docsy/blob/706dbf9af3d61338be9335e261d2a4538125800c/userguide/hugo.yaml#L67-L84

Start preview of the user guide:

```
hugo server --themesDir ../..
```

With configuration parameter `params.ui` not being present in `hugo.yaml` any more, a warning message is now emitted:

```
$ hugo server --themesDir ../..
Start building sites …
hugo v0.111.2-4164f8fef9d71f50ef3962897e319ab6219a1dad+extended windows/amd64 BuildDate=2023-03-05T12:32:20Z VendorInfo=gohugoio
```

As pointed out in #607 already, such a warning is too noisy and should be avoided.

deining:
This PR addresses #1464.

As pointed out in #608 already, the root cause of the warning is the wrong use of `isset`.
This PR replaces `isset` and uses `with` clauses instead, thus avoiding the confusing warn messages.

This PR closes #1464.

### DIFF
--- a/layouts/partials/sidebar-tree.html
+++ b/layouts/partials/sidebar-tree.html
@@ -1,8 +1,5 @@
 {{/* We cache this partial for bigger sites and set the active class client side. */ -}}
-{{ $sidebarCacheLimit := 2000 -}}
-{{ with .Site.Params.ui.sidebar_cache_limit -}}
-    {{ $sidebarCacheLimit = . -}}
-{{ end -}}
+{{ $sidebarCacheLimit := .Site.Params.ui.sidebar_cache_limit | default 2000 -}}
 {{ $shouldDelayActive := ge (len .Site.Pages) $sidebarCacheLimit -}}
 <div id="td-sidebar-menu" class="td-sidebar__inner{{ if $shouldDelayActive }} d-none{{ end }}">
   {{ if not .Site.Params.ui.sidebar_search_disable -}}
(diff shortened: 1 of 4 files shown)


chalin (changes requested):
Thanks for looking into this.

There's a simpler way to resolve such expressions, for example:

```
{{ $sidebarCacheLimit := and .Site.Params.ui .Site.Params.ui.sidebar_cache_limit | default 2000 -}}
```

It's _might_ be a good idea to also include a test of `.Site.Params` since that section is absent when, for example, hugo creates a new site.

```
{{ $sidebarCacheLimit := and .Site.Params .Site.Params.ui .Site.Params.ui.sidebar_cache_limit | default 2000 -}}
```

Could you change all assignment expressions to be of this form?

deining:
> Thanks for looking into this.

Thanks for reviewing this promptly.

> There's a simpler way to resolve such expressions, for example:
> 
> ```
> {{ $sidebarCacheLimit := and .Site.Params.ui .Site.Params.ui.sidebar_cache_limit | default 2000 -}}
> ```

Great idea. I don't see, however, why we additionally should evaluate `.Site.Params.ui` in this context. `.Site.Params.ui` is a map, so if `.Site.Params.ui.sidebar_cache_limit` is set, `.Site.Params.ui` is set, too. I think we can safely reduce this to:

```
{{ $sidebarCacheLimit := .Site.Params.ui.sidebar_cache_limit | default 2000 -}}
```

This is also what the [official documentation](https://gohugo.io/functions/default/) suggests.

> Could you change all assignment expressions to be of this form?

Done.

deining:
> Have you tested your solution with the `params.ui` section of the Hugo config commented out?

Yes, I did, an it returned the default value, as expected.

chalin (approved):
Strange. It is as if Hugo's `.` is acting as a safe selector (like `.?` in some other languages). I don't think it was always like that, but I'm happy that it works now. Thanks for fixing this.

chalin:
It might be worth fixing the other occurrences in the use of "isset"?

Issue #1466, opened 2023-03-08
[BSv5] Row/col formatting breaks Docsy components


chalin:
Formatting of rows and columns has changed in BSv5, in particular:

- There seems to be an (not-explicitly documented) assumption that a `row`'s child is meant to be a `col`. And a row's child is now set at `width` and `max-width` of 100%. This breaks the layout of several components in Docsy, like `blocks/*` and page footer.
- https://github.com/twbs/bootstrap/discussions/37951

Tasks:

- [ ] Foot adjustments:
  - [x] #1373
  We _might_ want to rework the footer, dropping the flex-shrink fix and adding a proper row/col hierarchy.
  ~Nope, can't get it to work that way 🤷‍♂️~ Yes, got it to work, see next item.
  - [ ] #1523
- [x] #1467
- [x] #1469
- [x] #1471
- [x] #1472

chalin:
I've searched for `.row` and ensured that the immediate children elements are columns. I think that all cases have been covered. The only exception is the footer, which I can't get to work other than how it was setup via #1373. Closing for now.

PR #1469, opened 2023-03-08
[BSv5] blocks/lead row-col hierarchy fix

chalin:
- Contributes to #1466
- Fixes the width/max-width of `blocks/lead`
- Makes the child of `section.row` a column (`col-12`)

**Preview**: https://deploy-preview-1469--docsydocs.netlify.app


chalin:
Note that the margins for the content of blocks/lead aren't exactly the same as they were for all widths pre-BSv5, but that's because of breakpoint changes in BSv5.

deining (approved):
LGTM

PR #1471, opened 2023-03-08
[BSv5] community/list row-col hierarchy fix

chalin:
- Contributes to #1466
- Fixes the width/max-width of `community/list.html`
- Makes the child of `section.row` a column (`col-12`)
- Makes padding style more precise in `layouts/shortcodes/blocks/cover.html`

**Preview**: https://deploy-preview-1471--docsydocs.netlify.app/community/


chalin:
Feedback is still welcome.

PR #1472, opened 2023-03-08
[BSv5] blocks/section row-col hierarchy fix

chalin:
- Contributes to #1466
- Contributes to #906
- Fixes the width/max-width of `blocks/section`s
- Section types are now switched! The `type` argument can be either `container` (the **default**) or `row`.
  This is a **BREAKING CHANGE**. I've updated the docs and the CHANGELOG.
- Whitespace cleanup
- Fixes a long-standing bug of a blank paragraph at the end of the About page, since before BSv5 changes

**Preview**: https://deploy-preview-1472--docsydocs.netlify.app/about

---

#### Before

Notice the following problems (all of which are resolved by this PR):

- White space between the end of the last block and the footer
- Misformatting of bulleted lists
- Excessive width

> image

#### After

> image

### DIFF
--- a/layouts/shortcodes/blocks/section.html
+++ b/layouts/shortcodes/blocks/section.html
@@ -1,16 +1,14 @@
-{{ $_hugo_config := `{ "version": 1 }` -}}
 {{ $col_id := .Get "color" | default .Ordinal -}}
 {{ $height := .Get "height" | default "auto"  -}}
-{{ $type   := .Get "type" | default "" -}}
-<a id="td-block-{{ .Ordinal }}" class="td-offset-anchor"></a>
+{{ $type   := .Get "type" | default "container" -}}
+
+<div><a id="td-block-{{ .Ordinal }}" class="td-offset-anchor"></a></div>
 <section class="row td-box td-box--{{ $col_id }} td-box--height-{{ $height }}">
-	<div class="col">
-		<div class="row {{ $type }}">
-			{{ if eq .Page.File.Ext "md" -}}
-				{{ .Inner | markdownify -}}
-			{{ else -}}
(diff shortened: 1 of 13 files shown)


layouts/shortcodes/blocks/section.html
```diff
@@ -1,16 +1,14 @@
-{{ $_hugo_config := `{ "version": 1 }` -}}
```
deining, line 1: by removing this line, this PR also contributes to #939. Can you add this in the issue description, please?

deining (approved):
LGTM.

layouts/shortcodes/blocks/section.html
chalin: Ah yes! Thanks for the reminder, ... and the review! 😄

LisaFC:
Looks great, thanks Patrice!

userguide/content/en/about/index.md
```diff
@@ -2,7 +2,6 @@
 title: About Docsy
 linkTitle: About
 menu: {main: {weight: 10}}
-layout: docs
```
chalin, line 5: @LisaFC et al.: prior to deleting line 5 above, there was a `_print` version of the about page, which pretty much looked like the non-print version. With line 5 gone, we've lot the `_print` version of the About page. Is that ok with you? Given that the two pages are identical AFAICT.

chalin: For the record, this is what the `_print` page looked like (excerpt), _before_ this PR was merged:

> image

LisaFC: I don't think that's a problem, given that the page was exactly the same.

PR #1475, opened 2023-03-10
Drop mycplus.com iframe from Shortcode page

### LINKED ISSUE #1468: Consider removing/changing the iframe to mycplus.com
The iframe to https://www.mycplus.com/featured-articles/hello-world-programs-in-300-programming-languages often causes delays in loading the page, brings in the Google Tag Manager etc. Consider dropping it or loading something less problematic.

chalin:
- Closes #1468
- Drop the mycplus.com iframe from Shortcode page so now a **page refresh**, on my machines takes **< 400 ms** as compared to 5+ sec. The section immediately after contains an example of a `{{< tab ... right=true text=true >}}`, so nothing is lost.
- Whitespace cleanup in various code blocks

### DIFF
--- a/userguide/content/en/docs/adding-content/shortcodes/index.md
+++ b/userguide/content/en/docs/adding-content/shortcodes/index.md
@@ -120,11 +120,9 @@ For announcement of latest features etc.
 ### blocks/feature
 
 ```go-html-template
-
 {{%/* blocks/feature icon="fa-brands fa-github" title="Contributions welcome!" url="https://github.com/gohugoio/hugo" */%}}
 We do a [Pull Request](https://github.com/gohugoio/hugo/pulls) contributions workflow on **GitHub**. New users are always welcome!
 {{%/* /blocks/feature */%}}
-
 ```
 
 | Parameter        | Default    | Description  |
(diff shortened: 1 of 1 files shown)


userguide/content/en/docs/adding-content/shortcodes/index.md
```diff
@@ -388,9 +383,6 @@ object HelloWorld extends App {
   println("Hello world!")
 }
 {{< /tab >}}
-{{< tab header="Reference" right=true text=true >}}
```
deining, line 391: I agree that `www.mycplus.com` needs to be replaced or removed.

What about replacing it with `http://helloworldcollection.de`.

```
{{< /tab >}}
{{< tab header="Reference" right=true text=true >}}
{{< iframe src="http://helloworldcollection.de" >}}
{{< /tab >}}
```

chalin:
This was discussed during the March PSC meeting, and we agreed to drop the `iframe`, especially since it is secondary to that section (which is for tabbed panes). As remarked in the notes:

> If we do choose to reintroduce an iframe, especially hidden under a tab, then it should be fully lazily loaded. (But I don't want to have to deal with that until BSv5 release of Docsy is out).

I'm going to merge this now since I keep hitting the iframe-loading delays during my BSv5 migration work, and I'm trying to wrap this up the BSv5 work ASAP. As mentioned above, we can look into re-introducing an iframe (if we feel that is important enough), provided it gets lazily loaded.

PR #1478, opened 2023-03-10
Tabpane shortcodes: improved reporting of warnings and errors

deining:
This PR follows up on https://github.com/google/docsy/pull/1448: Tabpane warning: make it easier to find source page.
Only now I realized that we can do even better here and report not only the source page, but also the offending position (line number) inside the page.
This PR improves various error messsages, they all report the position of the offending code now. Also the wording was improved for several messages, making them easier to understand.


chalin:
After rebasing, the build is failing. Can you take a look @deining? 

> 1:25:54 PM: ERROR 2023/05/04 17:25:54 shortcode "tab" must be enclosed inside a "tabpane" block, but no parent block was found. Error position: "/opt/build/repo/userguide/content/en/docs/adding-content/shortcodes/index.md:431:309"

chalin:
The good think about the error, is that it tested the new error reporting code. It's great to be able to click and have VS Code deep link to the point of error!

chalin (approved):
I didn't do a detailed review, but looks like a great improvement overall!

PR #1486, opened 2023-03-16
Fix page-meta for single language sites

### LINKED ISSUE #138: Broken logic for building GitHub URL for "Edit this page" link
Hallo to the Docsy theme team :)

The `page-meta-links` partial layout (at [line 8](https://github.com/google/docsy/blob/master/layouts/partials/page-meta-links.html#L8)) contains the following code, which doesn't look right:

```
{{ if and ($gh_subdir) (.Site.Language.Lang) }}
```

I tried removing the `and`, and I also tried moving the `and` to between the two bracketed clauses, but neither of those updates works as expected. So I've decided to hand this problem over to people who have more background into the recent updates to that file.

I found the above code while investigating a problem when upgrading Kubeflow to the latest Docsy. Kubeflow does not use a language directory, and therefore doesn't need the `/en` inserted into the GitHub path. On testing the upgrade,  we found that the latest version of Docsy insists on including `/en` in the file path. (Thank you to @joeliedtke for pointing out that the "Edit this page" link was broken!)

For Kubeflow, we'll to continue using our [custom layout](https://github.com/kubeflow/website/blob/master/layouts/partials/page-meta-links.html#L6), which checks `.Site.IsMultiLingual` to determine whether to add the `/en` or not. We need the custom layout anyway at this stage, because it contains the code to inject the page URL into the issue body when creating an issue. See feature request https://github.com/google/docsy/issues/91. At some stage, though, it'll be good to remove our custom layout and use the standard Docsy one, when both these issues are fixed.

### LINKED ISSUE #981: page meta links should respect `.Site.IsMultiLingual`
Currently, the page meta links will always include "en" even if the site is not multi-lingual and has content directly in the "contents" directory (vs "contents/en").

lyind:
Fixes #981
Fixes #138

## Summary

The "edit", "view source" and similar links in the page-meta area (on the right) are broken for single-language Hugo sites (HTTP 404).

Those sites *may* store their source documents without language specific path segment depending on the `.Site.IsMultiLingual` property.
 
This PR changes the page-meta partial template to take that property into account.

### DIFF
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -10,13 +10,9 @@
   {{ warnf "Warning: use of `github_url` is deprecated. For details see https://www.docsy.dev/docs/adding-content/repository-links/#github_url-optional" -}}
   <a href="{{ $gh_url }}" target="_blank"><i class="fa-solid fa-pen-to-square fa-fw"></i> {{ T "post_edit_this" }}</a>
 {{ else if $gh_repo -}}
-  {{ $gh_repo_path := printf "%s/content/%s" $gh_branch $pathFormatted -}}
-  {{ if and (.Site.IsMultiLingual) ($gh_subdir) (.Site.Language.Lang) -}}
-    {{ $gh_repo_path = printf "%s/%s/content/%s/%s" $gh_branch $gh_subdir ($.Site.Language.Lang) $pathFormatted -}}
-  {{ else if and (.Site.IsMultiLingual) (.Site.Language.Lang) -}}
-    {{ $gh_repo_path = printf "%s/content/%s/%s" $gh_branch ($.Site.Language.Lang) $pathFormatted -}}
-  {{ else if $gh_subdir -}}
-    {{ $gh_repo_path = printf "%s/%s/content/%s" $gh_branch $gh_subdir $pathFormatted -}}
+  {{ $gh_repo_path := printf "%s/content/%s" $gh_branch ( relLangURL $pathFormatted ) -}}
+  {{ if $gh_subdir -}}
+    {{ $gh_repo_path = printf "%s/%s/content%s" $gh_branch $gh_subdir ( relLangURL $pathFormatted ) -}}
   {{ end -}}
 
   {{/* Adjust $gh_repo_path based on path_base_for_github_subdir */ -}}
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -11,9 +11,9 @@
   <a href="{{ $gh_url }}" target="_blank"><i class="fa-solid fa-pen-to-square fa-fw"></i> {{ T "post_edit_this" }}</a>
 {{ else if $gh_repo -}}
   {{ $gh_repo_path := printf "%s/content/%s" $gh_branch $pathFormatted -}}
-  {{ if and ($gh_subdir) (.Site.Language.Lang) -}}
+  {{ if and (.Site.IsMultiLingual) ($gh_subdir) (.Site.Language.Lang) -}}
     {{ $gh_repo_path = printf "%s/%s/content/%s/%s" $gh_branch $gh_subdir ($.Site.Language.Lang) $pathFormatted -}}
-  {{ else if .Site.Language.Lang -}}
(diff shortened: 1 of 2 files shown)


LisaFC:
Thanks for this, and sorry for the belated review!

I am going to look at this properly but from back in b/138 we found checking IsMultiLingual didn't work for all sites - that may have changed since then, of course!

Going to also try this new tip from the Hugo forum: https://discourse.gohugo.io/t/check-if-default-language-content-is-in-a-language-subdir/21425/6

LisaFC:
So this seems to work even if there aren't any language subdirs (which the older version of IsMultiLang didn't seem to work for), hoorah!

Do you want me to try the new template functions and see if we can make it even simpler? I think we can get rid of some of the conditionals...

lyind:
I quickly tested [relLangURL](https://gohugo.io/functions/rellangurl/) in a `.Site.IsMultiLingual == false` environment: it works.

So one could indeed re-factor the code a bit.

lyind:
@LisaFC I took a shot at the refactoring, could you review, please?

LisaFC:
Just taking a look now! At first glance in a multilang site it seems to be working (give or take an extra slash) for everything except the default language, for some reason. Let me see if I can fix it.

LisaFC:
Right, got it working! Few review comments/edits incoming.

layouts/partials/page-meta-links.html
```diff
@@ -10,13 +10,9 @@
...
   <a href="{{ $gh_url }}" target="_blank"><i class="fa-solid fa-pen-to-square fa-fw"></i> {{ T "post_edit_this" }}</a>
 {{ else if $gh_repo -}}
-  {{ $gh_repo_path := printf "%s/content/%s" $gh_branch $pathFormatted -}}
-  {{ if and ($gh_subdir) (.Site.Language.Lang) -}}
-    {{ $gh_repo_path = printf "%s/%s/content/%s/%s" $gh_branch $gh_subdir ($.Site.Language.Lang) $pathFormatted -}}
-  {{ else if .Site.Language.Lang -}}
-    {{ $gh_repo_path = printf "%s/content/%s/%s" $gh_branch ($.Site.Language.Lang) $pathFormatted -}}
-  {{ else if $gh_subdir -}}
-    {{ $gh_repo_path = printf "%s/%s/content/%s" $gh_branch $gh_subdir $pathFormatted -}}
+  {{ $gh_repo_path := printf "%s/content/%s" $gh_branch ( relLangURL $pathFormatted ) -}}
```
LisaFC, line 13: So (for some reason the GH web UI won't let me comment on any lines outside your change) you need to do the following:

Change line 2 above from
 {{ $pathFormatted := replace .File.Path "\\" "/" -}}
to
{{ $pathFormatted := strings.TrimPrefix hugo.WorkingDir $.File.Filename -}}

Then change these lines to:
  {{ $gh_repo_path := printf "%s%s" $gh_branch $pathFormatted -}}
  {{ if $gh_subdir -}}
    {{ $gh_repo_path = printf "%s/%s%s" $gh_branch $gh_subdir $pathFormatted -}}

That works for both sites with language subdirectories and ones without!

LisaFC:
Here's my working version for comparison:
https://github.com/LisaFC/multilang/blob/main/layouts/partials/page-meta-links.html

lyind:
Which Hugo versions are we supposed to support (Hugo minimum version)?

LisaFC:
We specify a minimum of 0.110.0, but I think we may have to bump it to 0.112.0 for our next release if we use this as that's when that WorkingDir function was introduced.

(Hugo is currently on 0.119.0)

LisaFC:
Did that work?

LisaFC:
Bump! I'd like to get this in as it will fix one of our very very oldest unresolved issues.

lyind:
@LisaFC Will try it later.

lyind:
@LisaFC Tested your variant at https://github.com/LisaFC/multilang/blob/main/layouts/partials/page-meta-links.html with a single-language/no language site: **OK** ❤️👍

lyind:
Should we close this PR in favor of your change?

LisaFC:
OK! I'll add you to my PR and you can close yours then.

PR #1493, opened 2023-04-04
allow fetching js from same origin

datamuc:
There already is $td-enable-google-fonts which disables google-fonts. But the same GDPR issues problem arise with these javascript.

This patch allows to self host them without maintaining the complete head.html locally.


datamuc:
duplicate of #1204

PR #1501, opened 2023-04-21
[BSv5] fix main container height

yann-soubeyrand:
Setting a minimum height instead of a fixed height fixes footer overflow issues.

Before:
Capture d’écran du 2023-04-21 19-28-30

After:
Capture d’écran du 2023-04-21 19-28-52


chalin (approved):
Merci @yann-soubeyrand.

PR #1502, opened 2023-04-21
[BSv5] fix horizontal scrollbar issues

yann-soubeyrand:
IIUC, Bootstrap row has negative left and right margins to counteract col padding (https://getbootstrap.com/docs/5.3/layout/grid/#how-it-works). Setting col padding to 0 can lead to issues where a pointless horizontal scrollbar is added to the browser window.

Before:
Capture d’écran du 2023-04-21 19-39-00

After:
Capture d’écran du 2023-04-21 19-39-13

### DIFF
--- a/layouts/community/list.html
+++ b/layouts/community/list.html
@@ -4,7 +4,7 @@
 <section
   class="row td-box td-box--primary position-relative td-box--height-auto"
 >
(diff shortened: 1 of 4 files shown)


chalin (approved):
Thanks @yann-soubeyrand!

PR #1503, opened 2023-04-25
Blocks/feature fix

fekete-robert:
This PR:
- Documents the so far undocumented url_text parameter
- Changes the shortcode so it doesn't automatically add an ellipsis if url_text is set

### DIFF
--- a/layouts/shortcodes/blocks/feature.html
+++ b/layouts/shortcodes/blocks/feature.html
@@ -1,5 +1,4 @@
 {{ $icon := .Get "icon" | default "fa-lightbulb" -}}
-{{ $url_text := .Get "url_text" -}}
 <div class="col-lg-4 mb-5 mb-lg-0 text-center">
 <div class="mb-4 h1">
   <i class="{{ if not (or (hasPrefix $icon "fas ") (hasPrefix $icon "fab ")) }}fas {{ end }}{{ $icon }}"></i>
@@ -10,5 +9,5 @@
 <div class="mb-0">
 {{ .Inner }}
 </div>
-{{ with .Get "url" }}<p><a href="{{ . }}">{{ with $url_text }}{{ $url_text }}{{ else }}{{ T "ui_read_more" }} …{{ end }}</a></p>{{ end }}
+{{ with .Get "url" }}<p><a href="{{ . }}">{{ with .Get "url_text" }}{{ . }}{{ else }}{{ T "ui_read_more" }} …{{ end }}</a></p>{{ end }}
 </div>
--- a/layouts/shortcodes/blocks/feature.html
+++ b/layouts/shortcodes/blocks/feature.html
@@ -9,5 +9,5 @@
 <div class="mb-0">
 {{ .Inner }}
 </div>
-{{ with .Get "url" }}<p><a href="{{ . }}">{{ with .Get "url_text" }}{{ . }}{{ else }}{{ T "ui_read_more" }} …{{ end }}</a></p>{{ end }}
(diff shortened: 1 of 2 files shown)


layouts/shortcodes/blocks/feature.html
```diff
@@ -10,5 +10,5 @@ <h4 class="h3">
 <div class="mb-0">
 {{ .Inner }}
 </div>
-{{ with .Get "url" }}<p><a href="{{ . }}">{{ with $url_text }}{{ $url_text }}{{ else }}{{ T "ui_read_more" }}{{ end }} …</a></p>{{ end }}
+{{ with .Get "url" }}<p><a href="{{ . }}">{{ with $url_text }}{{ $url_text }}{{ else }}{{ T "ui_read_more" }} …{{ end }}</a></p>{{ end }}
```
deining, line 13: Minor nit: for the sake of brevity:

```
{{ with $url_text }}{{ . }} ...
```

instead of

```
{{ with $url_text }}{{ $url_text }} ...
```
Another inconsistency: we use `{{ with .Get "url" }}`, I think we could use `{{ with .Get "url_text" }}` here, too (variable `$url_text` is used only once!).

And yes, I know that these refer to existing code, not to changes proposed from you. But why not take the chance while we are at it?

deining (approved):
LGTM.

layouts/shortcodes/blocks/feature.html
fekete-robert: Sure, good idea :) Thanks a lot for the comments!

deining: Checks are failing now 😢
I guess we need `$.Get "url_text"` here!?

fekete-robert: Ah, right. I shouldn't try changing code so late. I might be wrong, but my guess is that the problem is that  `{{ with .Get "url_text" }}` is within the context of `{{ with .Get "url" }}`. Don't know if there is an easy way around it.

fekete-robert: I think `$.Get` will do the trick

userguide/content/en/docs/adding-content/shortcodes/index.md
```diff
@@ -129,6 +129,7 @@ We do a [Pull Request](https://github.com/gohugoio/hugo/pulls) contributions wor
 | ---------------- |------------| ------------|
 | title | | The title to use.
 | url | | The URL to link to.
+| url_text | The value of the `ui_read_more` parameter in the language file. For English, that's "Read more" | The link text to use.
```
chalin, line 132:
```suggestion
| url_text | The [language parameter](/docs/language/#internationalization-bundles) value of [`ui_read_more`](https://github.com/google/docsy/blob/main/i18n/en.toml)  followed by " …" | The link text to use.
```

chalin (commented):
That shortcode could use some cleanup, but it'll do for now. Copyedit suggestion inline. Thanks.

fekete-robert:
@chalin No, there wasn't any issue IIRC, I just needed the same feature and realized it was already there just not documented.

fekete-robert:
And I added the part about the ellipsis because it didn't make sense with the text we used, and figured it might be the case for others as well.

chalin:
Thanks for the clarifications @fekete-robert. My suggestion would be to reduce the scope of this PR to the doc-page update only. Once we settle on a solution to #1820, another PR can be submitted. I'm looking forward to discussing this with @emckean.

chalin (approved):
Per PSC meeting, I've dropped the ellipsis \o/
And tweak the table Default-description prose.
LGTM now. Thanks @fekete-robert.

PR #1509, opened 2023-05-04
module: ignore Bootstrap config

razonyang:
This PR ignores the config https://github.com/twbs/bootstrap/blob/main/config.yml from Bootstrap, since it will be merged into user's configuration and override some Hugo module's configuration which cause build fails.


deining (approved):
LGTM.

chalin (approved):
If @deining is happy, then so am I :)

chalin:
Thanks!

PR #1529, opened 2023-06-01
Bump dependency Font-Awesome to latest version 6.4.0

deining:
This PR bumps dependency `Font-Awesome` to latest released version 6.4.0.

PR #1418, opened 2023-02-15
Make docsy point to latest version of dependencies

deining:
This This PR is related to #1409. It ensures that we use and test docsy (module setup) with latest version of dependencies.

PR #1533, opened 2023-06-02
Make docsy theme point to latest version of dependencies

deining:
- This PR is related to #1529. It ensures that we use and test docsy (module setup) with latest version of dependencies.
- Supersedes #1418


chalin (approved):
RSLGTM

PR #1542, opened 2023-06-06
Make it possible to add custom attributes to the body tag

fekete-robert:
Make it possible to add custom attributes to the body tag, from the page frontmatter, similarly to body_class.


fekete-robert:
To be honest, it seems not many users missed the body_class feature so far, and I suspect even less will need the attributes feature - it's just something that will come handy for the pagefind integration.
While I agree that combining the two features would make sense and result in a nicer config, I think that the separate parameters are easier to configure, and combining them might be an overkill.

But if you think it's needed for the PR to be accepted I'll do it when I have some free time.

chalin:
- Demand never materialized (as anticipated in-thread). Thanks.

PR #1567, opened 2023-06-13
 UG, start page: improve Download CTA button (#1389)

### LINKED ISSUE #1389: Homepage: drop or replace Download CTA button
The homepage has a **Download** call-to-action (CTA) button. That doesn't make sense given that "downloading" Docsy is the least recommended of the installation options.

I'd like to suggest that we replace **Download** with [**Get started**](https://www.docsy.dev/docs/get-started/). Thoughts @LisaFC et al.?

deining:
This PR closes #1389.


chalin (approved):
LGTM

(I might follow up with some style tweaks.)

PR #1569, opened 2023-06-15
Update _variables.scss

lorenzoraffle:
just changed the font


chalin:
You can change the font locally for your project. We'll stick with the current font as the Docsy default for now.
/cc @emckean @LisaFC

PR #1572, opened 2023-06-15
Blog list layout: no need for row/col-12 grid styling of content

### LINKED ISSUE #1571: [BSv5] Blog posts page has layout issues
The symptom is that the RSS button takes up the equivalent space of a full column, as can be seen here:

> image

The problem was not present before the BSv5 migration, so I'll consider this as part of the BSv5 migration work (#470).

chalin:
- Contributes to #470
- Closes #1571
- Drops the unnecessary use of row/col-12 combos. The main content doesn't need any grid styling.

---

It is a bit hard to see, but there are no content changes in the layout other than:

- Replacing `<div class="row"><div class="col-12">{{ if .Pages -}}...</div></div>` by
  `<div class="td-blog-posts">{{ if .Pages -}}...</div>`
- And similarly replacing the `div`s for the pagination
- Change from `<h2>` to `<div class="h2">` of the post-list header

Note that the class names `td-blog-posts*` currently have not styles associated with them.

In the before and after images, note the following fixes:

- The RSS button is flush-right and no longer takes up a column's worth of space
- ~Spacing above the **View page source** page-meta buttons~

### Before

> image

### After

> image

Showing the pagination buttons:

> image

### DIFF
--- a/layouts/blog/list.html
+++ b/layouts/blog/list.html
@@ -5,40 +5,36 @@
   {{$.Scratch.Set "blog-pages" .Pages -}}
 {{ end -}}
 
-<div class="row">
-	<div class="col-12">
-		{{ if .Pages -}}
-		  {{ $pag := .Paginate (( $.Scratch.Get "blog-pages").GroupByDate "2006" ) -}}
-			{{ range $pag.PageGroups }}
-			<h2>{{ T "post_posts_in" }} {{ .Key }}</h2>
-			<ul class="td-blog-posts-list">
-				{{ range .Pages }}
-				<li class="td-blog-posts-list__item">
-					<div class="td-blog-posts-list__body">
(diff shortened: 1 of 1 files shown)


deining:
As seen in the [preview](https://deploy-preview-1572--docsydocs.netlify.app/blog/), the spacing above the `View page source` buttons is **not** fixed by this PR. I tried locally too, both with chrome and firefox and I'm always getting  the same, deficient result.

chalin:
Thanks for noticing that. I had only looked at the build of the example site -- it is strange that the page-meta spacing would be fixed there. I've stuck out that item from the opening comment -- it wasn't the original issue targeted by this PR. I can track the page-meta issue separately:

- #1573

Any other comments @deining?

deining (approved):
With the spacing issue moved out of #1573, this PR looks good to me now.

PR #1577, opened 2023-06-16
User guide, installation from example site: clone latest tagged version

deining:
This PR contributes to the release of [version 0.7.0](https://github.com/google/docsy-example/issues/222) of the docsy example site. It extends the `clone` command given ion the installation instructions of the docsy uer guide so that now the latest tagged/released version of the example site is cloned.


chalin (approved):
👍

PR #1576, opened 2023-06-16
Add TOML syntax highlighting to code snippet basic-configuration.md

coliff:

PR #1578, opened 2023-06-16
User guide: reflect updates in example site

deining:
Recent changes in the doscy example site:

- default branch was renamed from `master` to `main`.
- the config file was renamed from `config.toml` to `config.yaml`.

With this PR applied these changes are reflected in the docsy user guide, too.

This PR also includes the change proposed with PR #1576.


chalin (approved):
Thanks for these updates.

PR #1589, opened 2023-06-22
User guide: replace config.[toml|yaml] with hugo.[toml|yaml]

deining:
As of hugo version `0.113.0`, the command `hugo new site` now creates `hugo.toml` inside the newly created site directory, while with older versions, `config.toml` was used. This PR reflects this change throughout the whole user guide.


chalin (approved):
Thanks. LGTM

PR #1596, opened 2023-06-27
NPM package version: add suffix -dev.0-unreleased

chalin:
We might not have a 0.7.2 release, and instead go directly to 0.8.0, but we at least want all work until then to be identified as "unreleased".


deining (approved):
LGTM

Issue #1605, opened 2023-07-05
Tabpane shortcode fixes and enhancements (July)


chalin:
- [x] #1602
- [x] #1607
  - [x] Rename `persistLang()` and/or replace it -- https://github.com/google/docsy/pull/1602/files#r1252787856 
- [x] #1608
  - [x] Address warnings now issued for the user guide
- [x] #1609
- [x] #1611

chalin:
@deining - there might be an issue with the new `tabpane` code, but I'll investigate further once #1607 and #1608 are merged (without #1608 I can't test the new functionality in the User Guide).

chalin:
Main features added. Moving remaining items to:

- #1641

PR #1607, opened 2023-07-05
tabpane: use `data-` attribute instead of class name

chalin:
- Contributes to #1605
- Switches to attaching  the persistence key to tab HTML elements as a `data-` attribute rather than a CSS class name
- Preserves converting the persistence key to lowercase, though now there's no longer a new to replace any characters
- Other improvements to the JS to make it more robust


deining (approved):
LGTM

PR #1611, opened 2023-07-06
Tabpane: support persistence of all active tabs

chalin:
- Updates option value to disable persistence: now use `persist=disabled` (instead of `off`)
- Enhances tabpane to support the persistence of the active tab selection for **all tabpanes** (excluding those with `persist=disabled`),  rather than just the last activated tab
- Restores all persisted activated tabs on page load
- Persists in localStorage a count of the tab-activated events
- Persists each activated tab in localStorage using the tab's persistence key. The value associated with the key is the tab-activated-event count plus one (also saved to localStorage). This supports typical use cases, in addition to the one explained below.
- Drops `button.onclick` in favor of `click` event handlers.

### Handling of non-uniform shared tabs

Assume that a page contains three tabpanes with the following distribution of tabs A, B, C, D -- where these letters represent the tab-persistence keys:

1. A, B
2. B, D
3. A, B, C

If you activate tab B from any tabpane, it will become active in all three tabs and the selection will be persisted. If you then select C, then tab C will become active in tabpane 3, but tab panes 1 and 2 will remain active at B -- and these settings are persisted. If you then select A, then: tabpanes 1  and 3 become active at A, but tabpane 2 remains unchanged.

IMHO, this applies the principle of least surprise.

### Preview and tests

- https://deploy-preview-1611--docsydocs.netlify.app/docs/adding-content/content/#custom-sections -- select `yaml` and refresh the page. You'll see that the choice of language persists in this page and is active in other pages with the same tab-persistence key
- https://deploy-preview-1611--docsydocs.netlify.app/docs/adding-content/shortcodes/#tabbed-panes -- select, say, German, and Scala, then refresh to see the persistence in action. Return to the previous page to see that the `yaml` persistence is still active.

### Other

- FYI, I've successfully tested this on the OTel website.
- Updates to the changelog and docs will be made in a followup PR.

### DIFF
--- a/layouts/shortcodes/tabpane.html
+++ b/layouts/shortcodes/tabpane.html
@@ -106,7 +106,6 @@
         {{ if $disabled }} disabled{{ end -}}"
           id="{{ $tabid }}" data-bs-toggle="tab" data-bs-target="#{{ $entryid }}" role="tab"
           {{ if and $persistTab $persistKey -}}
-            onclick="tdPersistActiveTab({{ $persistKey }});" {{/* */ -}}
             {{ printf "%s=%q " $tpPersistAttrName $persistKey | safeHTMLAttr -}}
           {{ end -}}
           aria-controls="{{- $entryid -}}" aria-selected="{{- cond ( and ( not $activeSet ) ( not $disabled ) ) "true" "false" -}}">
--- a/layouts/shortcodes/tabpane.html
+++ b/layouts/shortcodes/tabpane.html
@@ -25,9 +25,9 @@
 
 {{ $_persist := .Get "persist" -}}
 {{ with $_persist -}}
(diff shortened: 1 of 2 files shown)


deining (approved):
Nice improvement! Thanks for your hard work!
LGTM, looking forward seeing this merged!
Once we manage to eventually review and incorporate long-standing PR #1202 and #1227, docsy theme will offer a really great comfort-solution for displaying ad handling tabs!

LisaFC:
This looks really good!

LisaFC (approved):

chalin:
Thanks for the reviews. As I mention in the associated issue, I'll update the tab-activation algorithm soonish, but this first version should be good enough for most sites. (I'll be fielding it soon too.)

PR #179, opened 2020-01-22
add option to translate feedback question and options

tobiaskohlbau:
Fixes #175 

Would like to gather more languages. Any chance there are people around who speak the missing languages?

PR #1615, opened 2023-07-07
Feedback questions: add localization

### LINKED ISSUE #175: i18n of page feedback
Currently the `feedback.html` partial is not internationalized as the values are hardcoded into English. 
https://github.com/google/docsy/blob/1b46d71fe3c1894cc17e0b2be066dc5345cbb616/layouts/partials/feedback.html#L17-L19

I've changed the files in my local repo to resolve this issue. If you the maintainers are happy to accept a PR for this I would like to submit it. If you wanna have a look at the current state see here: https://github.com/tobiaskohlbau/docsy/commit/ed57f05f89fe38837a69418e2cb5b33c54d1ee26

deining:
This PR adds localization to docsy's feedback feature.

---

- Closes #175
  - Previous implementation: #179
- **Preview**, e.g.: https://deploy-preview-1615--docsydocs.netlify.app/docs/

/cc @tobiaskohlbau

### DIFF
--- a/i18n/de.toml
+++ b/i18n/de.toml
@@ -79,3 +79,13 @@ other = "Wenn Du aktiv beteiligen und einen Beitrag zu {{ .Site.Title }} leisten
 other = "Wie Du selbst zu {{ .Site.Title }} beitragen kannst, kannst Du nachlesen in unseren"
 [community_guideline]
 other = "Richtlinien für Beiträge"
+
+# Feedback
+[feedback_title]
+other = "Rückmeldung"
+[feedback_question]
+other = "War diese Seite hilfreich?"
+[feedback_positive]
+other = "Ja"
+[feedback_negative]
+other = "Nein"
--- a/layouts/partials/feedback.html
+++ b/layouts/partials/feedback.html
@@ -14,10 +14,10 @@
   }
 </style>
 <div class="d-print-none">
-<h2 class="feedback--title">Feedback</h2>
-<p class="feedback--question">Was this page helpful?</p>
-<button class="btn btn-primary mb-4 feedback--answer feedback--answer-yes">Yes</button>
-<button class="btn btn-primary mb-4 feedback--answer feedback--answer-no">No</button>
+<h2 class="feedback--title">{{- i18n "feedback_title" -}}</h2>
+<p class="feedback--question">{{- i18n "feedback_question" -}}</p>
+<button class="btn btn-primary mb-4 feedback--answer feedback--answer-yes">{{- i18n "feedback_positive" -}}</button>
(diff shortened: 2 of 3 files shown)


layouts/partials/feedback.html
```diff
@@ -15,9 +15,9 @@
 </style>
 <div class="d-print-none">
 <h2 class="feedback--title">Feedback</h2>
-<p class="feedback--question">Was this page helpful?</p>
-<button class="btn btn-primary mb-4 feedback--answer feedback--answer-yes">Yes</button>
-<button class="btn btn-primary mb-4 feedback--answer feedback--answer-no">No</button>
+<p class="feedback--question">{{- T "feedback_question" -}}</p>
+<button class="btn btn-primary mb-4 feedback--answer feedback--answer-yes">{{- T "feedback_positive" -}}</button>
+<button class="btn btn-primary mb-4 feedback--answer feedback--answer-no">{{- T "feedback_negative" -}}</button>
```
chalin, line 20: Hmm, maybe we should start using `i18n` instead of its alias `T`? Ok to use the alias for now, since it what is used elsewhere in Docsy.

Ref: https://gohugo.io/functions/i18n/

layouts/partials/feedback.html
```diff
@@ -15,9 +15,9 @@
 </style>
 <div class="d-print-none">
 <h2 class="feedback--title">Feedback</h2>
```
chalin, line 17: Also add i18n for the feedback_title.

i18n/de.toml
```diff
@@ -79,3 +79,11 @@ other = "Wenn Du aktiv beteiligen und einen Beitrag zu {{ .Site.Title }} leisten
...
 [community_guideline]
 other = "Richtlinien für Beiträge"
+
+# Feedback
+[feedback_question]
+other = "War diese Seite hilfreich?"
+[feedback_positive]
+other = "Ja"
+[feedback_negative]
+other = "Nein"
```
chalin, line 89: Nit: end with a newline.

chalin (commented):
Great, thanks. See inline comments.

i18n/de.toml
deining: Newline added as requested.

layouts/partials/feedback.html
deining: > Also add i18n for the feedback_title.

Done. Thanks for spotting this!

layouts/partials/feedback.html
deining: > Hmm, maybe we should start using `i18n` instead of its alias `T`?

I like the idea and changed all occurrences of `T` to `i18n`.

chalin (approved):
Thank you ✨

PR #1633, opened 2023-07-29
Fix warning: '.File.UniqueID on zero object' (#1007)

### LINKED ISSUE #1007: hugo build: warning '.File.UniqueID on zero object'
**How to reproduce**

* Create a new site from scratch as described [here](https://www.docsy.dev/docs/get-started/docsy-as-module/start-from-scratch/).
* Create a new page inside the `docs` folder: `hugo new content/docs/_index.md`
* Run `hugo -D` and you will get this warning:

```
hugo v0.98.0-165d299cde259c8b801abadc6d3405a229e449f6+extended windows/amd64 BuildDate=2022-04-28T10:23:30Z VendorInfo=gohugoio
WARN 2022/05/09 09:57:25 .File.UniqueID on zero object. Wrap it in if or with: {{ with .File }}{{ .UniqueID }}{{ end }}
```

Using `hugo --panicOnWarning`, you can track down where the error originates:

```
Error: Error building site: failed to render pages: render of "section" failed:
execute of template failed: template: docs/list.html:12:11:
executing "main" at <partial "section-index.html" .>
: error calling partial: "...\layouts\partials\section-index.html:7:17":
execute of template failed: template: partials/section-index.html:7:17
: executing "partials/section-index.html"
at <where $pages "Parent.File.UniqueID" "==" $parent.File.UniqueID>:
error calling where:
Warning trapped. Remove the --panicOnWarning flag to continue.
```

So the offending code line is here:

https://github.com/google/docsy/blob/6fc622c3874b2af8d1d72a9212bf64acb41d9823/layouts/partials/section-index.html#L7

deining:
This PR closes #1007 (see my analysis there). The fix is quite easy, all we have to do here is to put a `{{ with }}` statement around the offending code line 8:

```
7  {{ with .Parent.File -}}
8     {{ $pages = (where $pages "Parent.File.UniqueID" "==" $parent.File.UniqueID) -}}
9  {{ end -}}
```

 Also I did some code cleanup in order to prevent superfluous whitespace and lines.
I did a `diff` of the published user guide, and there are no changes caused by this PR other than white space cleanup.

### DIFF
--- a/layouts/partials/section-index.html
+++ b/layouts/partials/section-index.html
@@ -1,31 +1,33 @@
 <div class="section-index">
-    {{ $parent := .Page }}
-    {{ $pages := (where .Site.Pages "Section" .Section).ByWeight }}
+    {{ $parent := .Page -}}
+    {{ $pages := (where .Site.Pages "Section" .Section).ByWeight -}}
     {{ $pages = (where $pages "Type" "!=" "search") }}
-    {{ $pages = (where $pages ".Params.hide_summary" "!=" true) }}
-    {{ $pages = (where $pages ".Parent" "!=" nil) }}
-    {{ $pages = (where $pages "Parent.File.UniqueID" "==" $parent.File.UniqueID) }}
-    {{ if or $parent.Params.no_list (eq (len $pages) 0) }}
+    {{ $pages = (where $pages ".Params.hide_summary" "!=" true) -}}
+    {{ $pages = (where $pages ".Parent" "!=" nil) -}}
+    {{ if .Parent.File -}}
+        {{ $pages = (where $pages "Parent.File.UniqueID" "==" $parent.File.UniqueID) -}}
+    {{ end -}}
+    {{ if or $parent.Params.no_list (eq (len $pages) 0) -}}
     {{/* If no_list is true or we don't have subpages we don't show a list of subpages */}}
-    {{ else if $parent.Params.simple_list }}
+    {{ else if $parent.Params.simple_list -}}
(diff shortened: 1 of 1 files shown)


layouts/partials/section-index.html
```diff
@@ -1,31 +1,33 @@
...
+    {{ $parent := .Page -}}
+    {{ $pages := (where .Site.Pages "Section" .Section).ByWeight -}}
     {{ $pages = (where $pages "Type" "!=" "search") }}
-    {{ $pages = (where $pages ".Params.hide_summary" "!=" true) }}
-    {{ $pages = (where $pages ".Parent" "!=" nil) }}
-    {{ $pages = (where $pages "Parent.File.UniqueID" "==" $parent.File.UniqueID) }}
-    {{ if or $parent.Params.no_list (eq (len $pages) 0) }}
+    {{ $pages = (where $pages ".Params.hide_summary" "!=" true) -}}
+    {{ $pages = (where $pages ".Parent" "!=" nil) -}}
+    {{ with .Parent.File -}}
```
chalin, line 7: Since we're not using the value, I'd opt for `if`:

```suggestion
    {{ if .Parent.File -}}
```

chalin (approved):
Wonderful, thank you. See inline comment for one nit.

chalin:
The test failure is due to a flaky external link check (which should be fixed via #1790).

layouts/partials/section-index.html
deining: Done.

chalin (approved):

PR #1648, opened 2023-08-05
Template for new files at GitHub: add info on customization

deining:
The template file `new-page-template.md` specifies the frontmatter and content used for new pages at GitHub (via link `Create new child page` in the right sidebar). This PR adds documentation on how to override the frontmatter and the content defined in this file via a project specific file in your project.

### DIFF
--- a/assets/stubs/new-page-template.md
+++ b/assets/stubs/new-page-template.md
@@ -6,6 +6,13 @@ description: >-
      Page description for heading and indexes.
 ---
 
+<!--
+In order to change the default frontmatter and the content for new pages,
+add a new template file `assets/stubs/new-page-template` to your project.
+This file will override the default settings of the theme. Adapt this file
+according to your needs.
+-->
+
 ## Heading
 
 Edit this template to create your new page.


chalin (commented):
I think that we should document this feature in the User Guide instead of adding an HTML comment to the template file, since the information is more pertinent to website maintainers than to contributors adding a new page.

My 2 cents. WDYT @emckean @LisaFC @geriom?

deining:
> I think that we should document this feature in the User Guide

This **is** [documented](https://www.docsy.dev/docs/adding-content/repository-links/) in the user guide already:

```
* Create child page: Brings the user to a create new file form in their fork of your docs repo.
The new file will be located as a child of the page they clicked the link on.
The form will be pre-populated with a template the user can edit to create their page.
You can change this by adding assets/stubs/new-page-template.md to your own project.
```
> I think that we should document this feature in the User Guide instead of adding an HTML comment to the template file

The HTML comment is an **additional** reminder, thought as backing of the docu in the userguide. We had support questions in the past regarding the use of this file, so think adding the comment makes sense.

chalin:
- The customization is already documented, as noted in review; stalled since 2023. Thanks.

Issue #1250, opened 2022-09-30
Docsy + Algolia v3


jmichelgarcia:
## Info
Source code: https://github.com/jmichelgarcia/docsy-algolia
Url: https://sensational-chebakia-db2125.netlify.app/

## Details
When using Algolia, the search result displays as a popup, as opposed to what is mentioned in the docs. Is this the expected behavior?

> When you’ve completed these steps, Algolia search should be enabled on your site. **Search results are displayed as a drop-down under the search box**, so you don’t need to add any search results page.

<img width="1478" alt="Screenshot 2022-09-30 at 15 23 52" src="https://user-images.githubusercontent.com/2745523/193291303-e4e09df5-0211-4111-b870-d4567e9938cf.png">

Secondly, once we have a narrow design, there is no search available. Is this expected?

![Screenshot 2022-09-30 at 15 25 14](https://user-images.githubusercontent.com/2745523/193291590-354fc983-324c-4fa7-8b26-a5035b3a988f.png)

LisaFC:
1. Yes, the new Algolia default search results box is a pop-up rather than a drop-down, we should update the docs. I'm not even sure if you *can* easily configure it as a drop-down but I'll look into it.
2. Let me take a look at that, I believe we didn't support sidebar/narrow site search for Algolia because it was a drop-down and didn't really work with the navbar, but if it's a pop-up we *should* be able to support it for sidebar search as well. It may look a bit strange on mobile though. I will experiment!

LisaFC:
1. Proposed update in #1270 
2. It should be possible to get it working for mobile/smaller screens - I've tried and it's easy to put extra searchboxes on a page so the left nav searchbox would work for Algolia. One snag is that the Algolia code requires each searchbox div to have its own id - at the moment we reuse the searchbox code exactly between the two locations so that won't work without a bit of tweaking, you just end up with two divs with the same id and only get the search box for one of them. The other interesting thing is that the new Algolia styling has a "mobile-width" version that's just a magnifying glass (see image) so we could possibly do some nice styling with it.

![image](https://user-images.githubusercontent.com/10744406/194881631-ba17084b-c535-445e-ae3e-f97c6eaa169c.png)

[...]

PR #1651, opened 2023-08-09
Algolia search, fix: show search box in the sidebar

### LINKED ISSUE #1433: Algolia - Mobile navigation/search
Hey everyone,

I've read the documentation regarding Algolia not supporting the sidebar, however our team recently re-enabled Algolia on our doc site for testing and we do not see the search box appearing in the mobile navigation dropdown. When I inspect the mobile navigation, I see the element `<div id="docsearch"></div>` but none of the content is displaying from it and the div doesn't actually appear. It's as if the Javascript from Algolia isn't being injected properly on mobile.

image

image

Is there any additional configuration settings or modifications that need to be made to the search-input partial to enable this? I see the search box on desktop just fine.

deining:
When enabling algolia search, only the search box in the navbar appears, the search box in the sidebar doesn't show up. As correctly analysed [here](https://github.com/google/docsy/issues/1250#issuecomment-1273359341), this is due to the fact that currently both search boxes use the same id in their corresponding div element (which violates the principle that ids should be unique(!)).

This PR solves this issue, both search boxes show up with this patch applied. This PR also updates the documentation on algolia search in the user guide: we now have to provide **two** `docsearch` functions in the docsearch javascript code.

---

- Closes #1433

### DIFF
--- a/layouts/partials/search-input.html
+++ b/layouts/partials/search-input.html
@@ -4,7 +4,8 @@
   <input type="search" class="td-search__input form-control td-search-input" placeholder="{{ T "ui_search" }}" aria-label="{{ T "ui_search" }}" autocomplete="off">
 </div>
 {{ else if .Site.Params.algolia_docsearch -}}
-<div id="docsearch"></div>
+{{ $.Scratch.Add "docsearch-id-num" 1 -}}
+<div id="docsearch-{{ mod (.Scratch.Get "docsearch-id-num") 2 }}"></div>
 {{ else if .Site.Params.offlineSearch -}}
 {{ $offlineSearchIndex := resources.Get "json/offline-search-index.json" | resources.ExecuteAsTemplate "offline-search-index.json" . -}}
 {{ if hugo.IsProduction -}}
--- a/userguide/content/en/docs/adding-content/search.md
+++ b/userguide/content/en/docs/adding-content/search.md
@@ -141,20 +141,25 @@ algolia_docsearch: true
    * In `head-end.html` add the DocSearch CSS:
 
       ```html
-      <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" />
+      <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" integrity="sha512-O6ywtjtiV4QGgC6cD75C2Tf04SlTn9vvka8oV/dYpDL1JWM4lVP0QoZbdZt5RLtOWDpaP+MObzUrwl43ssqfvg==" crossorigin="anonymous" />
       ```
 
-
-   * In `body-end.html` add the DocSearch script, replacing the `docsearch` details with the snippet you get from Algolia (the example below is Algolia's own site index!). You must provide `#docsearch` as your `container` value as that's the ID of the `div` we provide in Docsy's layout:
+   * In `body-end.html` add the DocSearch script. In the code below, replace `docsearch` argument values for `appId`, `apiKey` and `indexName` with those provided by Algolia. Use the `container` values given below, they correspond to the IDs found in Docsy's layout:
 
       ```html
-      <script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3"></script>
-      <script type="text/javascript">docsearch({
-        container: '#docsearch',
-        appId: 'R2IYF7ETH7',
-        apiKey: '599cec31baffa4868cae4e79f180729b',
-        indexName: 'docsearch',
-        });</script>
+      <script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3" integrity="sha512-sQFwlNnBlEyWdI4EwJ9d2wjViu0ZPmIMjPP8kCmBoOjqcljGndf2gb4mldT4ZHxKCQcrdJ0+rxnMFKHuGWB7ag==" crossorigin="anonymous" ></script>
+      <script type="text/javascript">
+        for (let i = 0; i < 2; i++) {
+          docsearch({
+            container: `#docsearch-${i}`,
+            // Replace the values below with those for your site.
+            // These are Algolia's test values used for illustrative purposes:
+            appId: 'R2IYF7ETH7',
+            apiKey: '599cec31baffa4868cae4e79f180729b',
+            indexName: 'docsearch',
+          });
+        }
+      </script>
      ```
 
 You can find out more about how to configure DocSearch in the Algolia DocSearch V3 [Getting started](https://docsearch.algolia.com/docs/DocSearch-v3) guide.


deining:
> It completely makes sense that the IDs should be unique!
> 
> I'd like to propose that this be solved more simply by using, say, an ordinal (`.Ordinal`) as a suffix to the ID. Such an approach will confine changes to `layouts/partials/search-input.html`. It will also make it easier to keep things dry during the initializing calls to `docsearch()`: we'll be able to iterate from 0 to the number_of_input_searches - 1 in our calls to that function.
> 
> WDYT?

While authoring my PR, I had exactly the same idea of using an `.Ordinal`. Then I decided to another route and ended up with a solution that works both when publishing (via `hugo`) and previewing (via `hugo server`).

I tried to implement the PR following your recommendations given above and ended up with this code:

```
{{ with .Scratch.Get "ordinal" }}
  {{ $.Scratch.Add "ordinal" 1 -}}
  {{ else }}
  {{ $.Scratch.Set "ordinal" 1 -}}
{{ end }}
```

and for the output of the search box `<div>`:

```
<div id="docsearch-{{- sub ($.Scratch.Get "ordinal") 1 -}}"></div>
```

First I liked this minimal solution. It works perfectly fine when publishing (via command `hugo`), here I'm getting two `div` elements with IDs `docsearch-0` and `docsearch-1`.
Soon I realized that is does **not** work when running `hugo server`, here, I'm getting two `div` elements with IDs `docsearch-3` and `docsearch-4` (for whatever reason). Any ideas why this happens?

How to proceed here?

> (Also, I can't recall if we have an easy way to test this?)

Good point. Fortunately, I do have a algolia enabled site at hand, but I don't have a solution how to test this in an easy manner in this repo.

chalin:
Thanks for trying all of that out.

To be clear: so both the use of `.Ordinal`, and the solution you outline above, result in different ordinal values depending on the hugo command used (build or serve)?

That's strange.

---

Regardless, consider using `mod`, since we need only two distinct IDs:
```mustache
<div id="docsearch-{{ mod .Ordinal 2 }}"></div>
```

Or use `($.Scratch.Get "docsearch-id-num")` if `.Ordinal` doesn't work.

deining:
> Thanks for trying all of that out.

You are welcome.

> Or use `($.Scratch.Get "docsearch-id-num")` if `.Ordinal` doesn't work.

`.Ordinal` is availbale in shortcodes only, I didn't manage to get it work in partials.

> To be clear: so both the use of `.Ordinal`, and the solution you outline above, result in different ordinal values depending on the hugo command used (build or serve)?

Yes, the solution outline above results in different ordinal values depending on the hugo command used.

> That's strange.

Indeed.

> Regardless, consider using `mod`, since we need only two distinct IDs:
> 
> ```
> <div id="docsearch-{{ mod .Ordinal 2 }}"></div>
> ```

Great idea. I did that and now it works like a charm!

Thanks for your input, looking forward seeing this merged soon!

userguide/content/en/docs/adding-content/search.md
```diff
@@ -141,20 +141,27 @@ algolia_docsearch: true
    * In `head-end.html` add the DocSearch CSS:
 
       ```html
-      <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" />
+      <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" integrity="sha512-O6ywtjtiV4QGgC6cD75C2Tf04SlTn9vvka8oV/dYpDL1JWM4lVP0QoZbdZt5RLtOWDpaP+MObzUrwl43ssqfvg==" crossorigin="anonymous" />
       ```
 
-
-   * In `body-end.html` add the DocSearch script, replacing the `docsearch` details with the snippet you get from Algolia (the example below is Algolia's own site index!). You must provide `#docsearch` as your `container` value as that's the ID of the `div` we provide in Docsy's layout:
+   * In `body-end.html` add the DocSearch script, replacing the `docsearch` details with the snippet you get from Algolia (the example below is Algolia's own site index!). You must provide two `docsearch` functions, with their `container` values matching the IDs of the `div` elements we provide in Docsy's layout:
```
chalin, line 147:
```suggestion
   * In `body-end.html`, add the DocSearch script. In the code below, replace `docsearch()` argument values for `appId`, `apiKey`, and `indexName` with those provided by Algolia. Use the `container` values shown, since they correspond to the IDs found in Docsy's layout:
```

userguide/content/en/docs/adding-content/search.md
```diff
@@ -141,20 +141,27 @@ algolia_docsearch: true
...
+          appId: 'R2IYF7ETH7',
+          apiKey: '599cec31baffa4868cae4e79f180729b',
+          indexName: 'docsearch',
+        });
+        docsearch({
+          container: '#docsearch-1',
+          appId: 'R2IYF7ETH7',
+          apiKey: '599cec31baffa4868cae4e79f180729b',
+          indexName: 'docsearch',
+        });
```
chalin, line 163: Let's help projects keep things DRY:

```suggestion
      for (let i = 0; i < 2; i++) {
        docsearch({
          container: `#docsearch-${i}`,
          // Replace the values below with those for your site.
          // These are Algolia's test values used for illustrative purposes:
          appId: 'R2IYF7ETH7',
          apiKey: '599cec31baffa4868cae4e79f180729b',
          indexName: 'docsearch',
        });
      }
```

chalin (commented):
Much cleaner, as you point out.

A few more comments, see inline.

layouts/partials/search-input.html
```diff
@@ -1,10 +1,15 @@
...
+  {{ $.Scratch.Set "docsearch-id-num" 1 -}}
+{{ end -}}
 {{ if .Site.Params.gcs_engine_id -}}
 <div class="td-search">
   <div class="td-search__icon"></div>
   <input type="search" class="td-search__input form-control td-search-input" placeholder="{{ T "ui_search" }}" aria-label="{{ T "ui_search" }}" autocomplete="off">
 </div>
 {{ else if .Site.Params.algolia_docsearch -}}
-<div id="docsearch"></div>
+<div id="docsearch-{{ mod (.Scratch.Get "docsearch-id-num") 2 }}"></div>
```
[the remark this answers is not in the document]
deining: Good point, simplified as requested.

layouts/partials/search-input.html
```diff
@@ -1,10 +1,15 @@
+{{ with .Scratch.Get "docsearch-id-num" -}}
+  {{ $.Scratch.Add "docsearch-id-num" 1 -}}
+{{ else -}}
+  {{ $.Scratch.Set "docsearch-id-num" 1 -}}
+{{ end -}}
```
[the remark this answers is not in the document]
deining: Moved block into `if else` clause.

userguide/content/en/docs/adding-content/search.md
deining: Good suggestion, implemented as-is.

userguide/content/en/docs/adding-content/search.md
deining: Changed wording as proposed.

chalin (approved):
Wonderful, thanks!

PR #1653, opened 2023-08-10
CHANGELOG copyedits

chalin:


chalin:
@LisaFC gave a thumbs up during the last PSC. Merging so that I can make followup changes.

PR #1678, opened 2023-09-12
Algolia search, fix: point js + css file to specific version

deining:
Currently js + css file for algolia docsearch are pulled in via these URLs:

* https://cdn.jsdelivr.net/npm/@docsearch/css@3
* https://cdn.jsdelivr.net/npm/@docsearch/js@3

When pulling in these files, they are checked via a SRI hash. In both cases, SRI hashes are given for version 3.5.1. However, meanwhile @docsearch npm is at version 3.5.2. This means, that given SRI hashes are invalid now, algolia docsearch does not work and more.

This PR fixes this by pulling in specific version `3.5.2`.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -58,8 +58,8 @@
 {{ define "algolia/head" -}}
 
 {{ if and .Site.Params.search (isset .Site.Params.search "algolia") -}}
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3"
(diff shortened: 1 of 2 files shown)


emckean (approved):
LGTM, thanks so much for fixing this!

froboy:
Confirming I'm seeing similar behavior. Docsearch seems to be broken in HEAD without this.

chalin (approved):
Thanks!

Issue #1655, opened 2023-08-10
Docsy/User-guide "skin" styles should be factored out and/or dropped in some cases


chalin:
This issue is a part of:

- #1431

Docsy has **essential** (including flow-related and spacing) styles, and other style changes that we might consider a Docsy/User-guide "**skin**" -- such as choices for colors, preferred table formating, etc: 
https://github.com/google/docsy/blob/d30b1c9468294c5d5327b63124490f3bdc758ceb/assets/scss/_variables.scss#L8-L18

Another example is this button styling: https://github.com/google/docsy/blob/d30b1c9468294c5d5327b63124490f3bdc758ceb/assets/scss/_buttons.scss#L3-L15

Right now, it is a challenge for projects to opt out of these "skin" styles, since they need to address each style change individually.

PR #1684, opened 2023-10-02
Make links underlined by default

froboy:
As per the change upstream in Bootstrap 5 https://github.com/twbs/bootstrap/pull/30389

Removing the underline from links was in vogue for a while, but WebAIM [recommends not doing that](https://webaim.org/techniques/hypertext/link_text#underlining).

> Browsers underline hypertext links by default. It is possible to remove the underline using Cascading Style Sheets (CSS), but this is a bad idea most of the time.

### DIFF
--- a/assets/scss/_sidebar-toc.scss
+++ b/assets/scss/_sidebar-toc.scss
@@ -50,12 +50,24 @@
 
       &:hover {
         color: $blue;
-        text-decoration: none;
       }
     }
   }
 
   ul {
     padding-left: 0;
+  }
+}
+
+.td-page-meta,
+.td-sidebar-toc,
+.td-toc {
+  a {
+    text-decoration: none;
+
+    &:hover,
+    &:focus {
+      text-decoration: underline;
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


froboy:
Cursor_and_Welcome_to_Docsy___Docsy
Maybe we just do this for the body? Or we leave the default and override for `nav, aside`

froboy:
Ok, sorry for the mess of changes here, and for not opening an issue first, but this is ready for review now. Thanks for your consideration and all of your work. 

We're using this in production for the YMCA Website Services documentation at https://ds-docs.y.org/ and it's gotten fantastic feedback.

chalin:
Thanks for this PR!

Nice website, you might consider submitting a PR to added it to https://www.docsy.dev/docs/examples/?

I agree that having underlines in the body only (as you've done on your website) is probably best. I'll take a closer look soon.

chalin:
Here's the issue to track this:

- #1814

Also, I think I have an alternative to propose that will contribute to #1655 at the same time.

chalin:
Thanks again for this PR, it helped kick this off. A fix has landed, see:

- #1815

froboy:
Thank you! I appreciate your consideration of the issue and I'll get our site submitted to the list of examples. Thanks for your work on Docsy!

PR #1688, opened 2023-10-11
update mermaid to 10.5.0

dogmatic69:


deining:
This PR is outdated already, the latest version of mermaid is `10.6.1`  right now,
In order to better cope with mermaid's very short release intervals, I authored #1410. With this PR in place, the latest version of mermaid is pulled in automatically by default. However users can pin mermaid to any custom version they prefer by simply specifiying the desired version in their site config `hugo.yaml`.

dogmatic69:
And that PR is near a year now. 😮‍💨

Issue #1337, opened 2022-12-16
Breadcrumb links should be encoded as local path not external URLs


chalin:
Currently, links in breadcrumbs are encoded as external URLs, even if they are obviously site-local paths. This causes the following problems:

- When working on docsy sites served from Cloud editors like those offered by GitPod. The "external" links encoded in the breadcrumbs don't actually correspond to the URL the site is served from -- which can be of the form https://1313-foo-bee-doo.ws-us79.gitpod.io.
- External link checking is unnecessarily burdened (which leads to timeouts etc) for external links that are actually both site-local and unlikely to be invalid.

I thought that we had an issue open over this before, but can't find any. Having previously investigated this, there doesn't seem to be a hard requirement (contrary to what is state in some articles), that breadcrumb links be full URLs.

chalin:
Another option is to use link-checker ignore attributes. E.g., see:

- #1432

PR #1699, opened 2023-10-29
Change breadcrumb and print links to RelPermalink

fekete-robert:
The breadcrumb links and the print metalink used Permalink instead of RelPermalink, causing them to always link to the full domain referenced in the baseurl of the build. Using relpermalink makes these links work without referencing the baseurl, so they work even in dev environments (for example, when the site is deployed in a local docker container).

---

**Preview**, e.g.: https://deploy-preview-1699--docsydocs.netlify.app/docs/get-started/docsy-as-module/

### DIFF
--- a/layouts/partials/breadcrumb.html
+++ b/layouts/partials/breadcrumb.html
@@ -23,7 +23,7 @@
     {{ if $isActive -}}
       {{ .p1.LinkTitle -}}
     {{ else -}}
-      <a href="{{ .p1.Permalink }}">{{ .p1.LinkTitle }}</a>
+      <a href="{{ .p1.RelPermalink }}">{{ .p1.LinkTitle }}</a>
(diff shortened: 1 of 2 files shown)


chalin:
Nice! Something I've been wanting to fix for quite a while!

Fixes:

- #1337

Were you able to confirm that breadcrumb links encoded as paths rather than full URLs won't break SEO etc.?

fekete-robert:
TBH I haven't thought that this would impact SEO, just noticed that these
are the only ones with absolute links.

chalin (approved):
I'm so glad to see this land, I've been wanting this fixed for a while now. Thanks @fekete-robert!

fekete-robert:
Glad that I could help :)
I hope to take a look at the PR you mentioned me in by tomorrow.

PR #1700, opened 2023-11-02
add post_view_this to i18n ja

hanapedia:
## What's changed
- Added missing post_view_this for i18n ja translation


chalin (approved):
Thanks!

Issue #1667, opened 2023-08-17
Release 0.8.0 preparation


chalin:
### Issues/PRs in scope (actual)

- [x] #1120
  - [x] #1743
- [x] #1311
- [x] Google Analytics 4 (GA4) support:
  - [x] #1096
  - [x] #1728
  - [x] #1302
  - [x] #1742
- [x] #1337
  - [x] #1699
- [x] #1385
- [x] #1641

Other / support / loose ends:

- [x] Document that:
  - [x] New page feedback feature emits a custom `page_helpful` event - #1742
  - [x] New single-Hugo module impacts non-Hugo-module projects by creating a sibling folder - #1743
- [x] https://github.com/google/docsy-example/pull/255

### Issues/PRs in scope (_candidates_)

See [milestone 23Q4](https://github.com/google/docsy/milestone/9).

- #1558
- #1100
- #941
- #1654 
- #1528

### Release-prep checklist

For the complete instruction set, see [Publishing a release](https://github.com/google/docsy/blob/main/CONTRIBUTING.md#publishing-a-release):

- [x] Update Changelog: for inspiration, go over changes since last release https://github.com/google/docsy/compare/v0.7.2...main
- [x] Update the Docsy version in `package.json` and [userguide/hugo.yaml](https://github.com/google/docsy/blob/main/userguide/hugo.yaml)
- [x] Test pre-release from branch: 👍🏻 
- ~Pre-release patches: ...~
- [x] Create and push new version tag
- [x] Create and publish the new release 🎉

Docsy example followup (tracked separately):

- https://github.com/google/docsy-example/pull/257
- https://github.com/google/docsy-example/issues/258

### Related

- Previous: #1666
- Next: #1759
- https://github.com/google/docsy/compare/v0.7.2...main

Issue #1302, opened 2022-11-01
Page feedback: support Google Analytics 4 (GA4)


chalin:
Universal Analytics (UA) is deprecated. For details, see #1096. Here's some of the code associated with the page feedback feature: https://github.com/google/docsy/blob/d9bf98b876d74772b850ef910c6f59b898f61ccf/layouts/partials/feedback.html#L37-L53

This still uses the `analytics.js` library's `ga()` function. Note that this code is **incompatible with GA4** because of the library that is used and the event name, as is explained below.

### Tasks

- [x] [Migrate from analytics.js to gtag.js](https://developers.google.com/analytics/devguides/migration/ua/analyticsjs-to-gtagjs) - #1177
- [x] #1727
- [x] Fix event emitting code that calls `gtag` - #1726
  - [x] Confirm that `page_helpful` events are being emitted, with a default value of 100 for "yes" and 0 for "no" responses:
    - [x] For docsy.dev
    - [x] For example.docsy.dev
- [x] Add entry to the CHANGELOG
- [x] Update the docs, [Access the feedback data](https://www.docsy.dev/docs/adding-content/feedback/#access-the-feedback-data) in particular - #1726
- [x] Add a comment to #1426

---

Notes:

- Interestingly, Kubernetes docs had switched to using `gtag.js` in their override of [layouts/partials/feedback.html](https://github.com/kubernetes/website/blob/main/layouts/partials/feedback.html) already 4 years ago
- Indirectly related:
  - #175

/cc @caniszczyk @nate-double-u

chalin:
Note: the use of `click` as a custom event name _might not prevent_ projects from accessing page-feedback event data (I'm still exploring how to do this in GA4), but it does seem to make it unnecessarily harder when the same event name is also used by GA4 for another purpose.

chalin:
In the interim, it seems that projects can define an event-modification rule through the analytics console interface, to change the event name of page-feedback events to `page_helpful` from `click`. For extra details, see https://github.com/kubernetes/website/issues/37657#issuecomment-1301548361.

Thanks for the pointer to the relevant GA4 doc pages @emckean! 🧡

jmichelgarcia:
_Thank you for the pointers here, I just spent one hour trying to figure out this_

chalin:
All done!

PR #1726, opened 2023-11-11
[GA4] Emit `page_helpful` events for page feedback

chalin:
- Contributes to #1667
- Contributes to #1302
- Switches to emitting a `page_helpful` rather than a `click` event when the page feedback button is clicked. This makes it easier to view page feedback in the GA4 analytics console. FYI, the code in this PR has been used and vetted as part of the falco.org website, see https://github.com/falcosecurity/falco-website/issues/974.
- Adds support for a configurable `max_value` emitted for "yes" events, whose default value is now 100
- Updates the docs to mention the new configuration parameter.
- Adds changelog entry

**Preview**, e.g.: https://deploy-preview-1726--docsydocs.netlify.app/docs/adding-content/feedback/#search-engine-optimization-meta-tags -- navigate to the end of the page to see the feedback form

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,11 @@ For the full list of changes, see the [release][0.8.0] notes.
 
 **Breaking changes**:
 
-- **Page feedback**, or [User feedback] ([#1726]):
-  - Feedback-event attribute changes:
+- **Page feedback**, or [User feedback]:
+  - In support of projects configuring analytics outside of Docsy, feedback
+    functionality is enabled regardless of whether
+    `site.Config.Services.GoogleAnalytics.ID` is set ([#1727]).
+  - Feedback-event attribute changes ([#1726]):
     - Event `name` is `page_helpful`rather than`click`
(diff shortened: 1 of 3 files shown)


fekete-robert:
I don't have a setup where I could actually test it, but LGTM

chalin:
Thanks for the feedback @fekete-robert. Right, we'll have to iterate, if necessary, given that preview builds don't include analytics. This code was in use in another one of our projects, so there's a good chance it'll work here too.

geriom:
I want to test this, it may take a couple of more days. If you need this merged sooner I can approve it.

geriom (approved):
LGTM

chalin:
LOL, I forgot that we don't have analytics setup yet for the User Guide!

- #1097

chalin:
I'll test as soon as I can and report here. (Since the UG _isn't_ using this new feature, then there's no rush.)

chalin:
Ok, feature tested! `page_helpful` events are being emitted, with a value of 100 for "yes" responses. (Test results shared with PSC out-of-band.)

PR #1734, opened 2023-11-13
Fall back to Hugo's default behavior for disableHugoGeneratorInject

### LINKED ISSUE #1707: Docsy does not respect configuration parameter "disableHugoGeneratorInject"
According to https://gohugo.io/getting-started/configuration/ the behavior should be like this:

Hugo will, by default, inject a generator meta tag in the HTML head on the home page only. You can turn it off by setting **disableHugoGeneratorInject = true**

Docsy seems to inject the generator meta tag **in each HTML head unconditionally**.

chalin:
- Fixes #1707 
  That is, without this PR we get the `generator` meta tag in **every page**:
  ```console
  $ find userguide/public/* -type f -name "*.html" -exec grep -nH -e 'name="generator"' {} +
  userguide/public/404.html:6:<meta name="generator" content="Hugo 0.120.4">
  userguide/public/about/index.html:6:<meta name="generator" content="Hugo 0.120.4">
  userguide/public/blog/2022/hello/index.html:6:<meta name="generator" content="Hugo 0.120.4">
  ...
  ```
- With this fix, with or without disableHugoGeneratorInject set to true, we get the Hugo-default behavior of at most adding the meta tag to the homepage:
  ```console
  $ npm run build 
  $ find userguide/public/* -type f -name "*.html" -exec grep -nH -e 'name="generator"' {} +
  userguide/public/index.html:4:  <meta name="generator" content="Hugo 0.120.4">
  $ HUGO_DISABLEHUGOGENERATORINJECT=true npm run build                                                
  $ find userguide/public/* -type f -name "*.html" -exec grep -nH -e 'name="generator"' {} +
  $ 
  ```

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -1,6 +1,5 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-{{ hugo.Generator }}
 {{ range .AlternativeOutputFormats -}}
 <link rel="{{ .Rel }}" type="{{ .MediaType.Type }}" href="{{ .Permalink | safeURL }}">
 {{ end -}}


fekete-robert (approved):
Works as expected

LisaFC (approved):

PR #1737, opened 2023-11-13
Properly set baseURL for the User Guide

chalin:
- Fixes the User-guide Hugo-config param `baseURL`, setting it to the full site URL, as recommended by Hugo folks; see https://discourse.gohugo.io/t/permalink-not-permalinking/42613.
- In preparation for https://github.com/google/docsy/issues/1337#issuecomment-1808180217

### DIFF
--- a/userguide/hugo.yaml
+++ b/userguide/hugo.yaml
@@ -1,4 +1,4 @@
-baseURL: /
+baseURL: https://www.docsy.dev/
 title: &title Docsy
 description: &desc Docsy does docs
 enableRobotsTXT: true


fekete-robert:
We usually use different hugo environments, and set the baseurl to / in the default one, which works for local builds, and set it to the full url in the production environment, which is used for the prod deployment. (And also set the GA id and similar only in prod.)

emckean:
Oh, maybe we can add that as a best-practice tip in the docs?

chalin:
For the projects I work on, including the setup for the User Guide: the base config for production. We pass in a different baseURL for preview builds: https://github.com/google/docsy/blob/eaba498955e01525e5c6a7a6c70ea3b067e759d1/userguide/package.json#L9

Dev builds are handled like this: https://github.com/google/docsy/blob/eaba498955e01525e5c6a7a6c70ea3b067e759d1/userguide/package.json#L7

PR #1743, opened 2023-11-16
CHANGELOG + docs: mention single Hugo-module support

### LINKED ISSUE #1120: Offer Docsy as a single Hugo Module
The intent would be to **streamline use of Docsy for Hugo Module users**, something which (IMHO) makes sense to do now that [Git submodules]() were removed in [0.4.0](https://github.com/google/docsy/blob/main/CHANGELOG.md#040), and NPM-package support is available (#949).

@deining - I can't find the thread, but at some point we discussed the possibility of eliminating the `dependencies` module, which was introduced as a means of addressing:

- #848

This issue is to record that intent, and offer a place for discussing it.

### Tasks

- [x] Module rework: #1217
  - [x] #173 
- [x] #1730
- [ ] #1731
- [ ] User guide updates - #1743
- [ ] Changelog entry marking this as a breaking change for Hugo-module users of Docsy - #1743

### References

- https://github.com/golang/go/issues/37397

chalin:
- Fixes #1120


deining (approved):
Made some further adaptions. LGTM now.

chalin:
Great! Thanks for the review and the added changes.

PR #1744, opened 2023-11-16
Fix links for single language sites

### LINKED ISSUE #138: Broken logic for building GitHub URL for "Edit this page" link

### LINKED ISSUE #981: page meta links should respect `.Site.IsMultiLingual`

LisaFC:
- [x] Fixes #138
- [ ] Fixes #981

Requires Hugo 0.112.0+

### DIFF
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -1,5 +1,5 @@
 {{ if .File }}
-{{ $pathFormatted := replace .File.Path "\\" "/" -}}
+{{ $pathFormatted := strings.TrimPrefix hugo.WorkingDir $.File.Filename -}}
 {{ $gh_repo := ($.Param "github_repo") -}}
 {{ $gh_url := ($.Param "github_url") -}}
 {{ $gh_subdir := ($.Param "github_subdir") -}}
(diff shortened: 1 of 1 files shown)


LisaFC:
See discussion with @lyind in #1486

chalin (changes requested):
This breaks the page-meta links for some pages on the OTel website, for example the [mission](https://opentelemetry.io/community/mission/) page. I'll investigate further as soon as I can.

chalin:
@LisaFC et al. - do you have a reference to a public site that exhibits the broken page-meta links? Or do you have a simple test case to reproduce the problem?

chalin:
Ah, nm, found one :)

layouts/partials/page-meta-links.html
```diff
@@ -1,5 +1,5 @@
 {{ if .File }}
-{{ $pathFormatted := replace .File.Path "\\" "/" -}}
+{{ $pathFormatted := strings.TrimPrefix hugo.WorkingDir $.File.Filename -}}
```
chalin, line 2: The line your are changing was added in support of builds under Windows:

- #163

Removing the line will introduce a regression IMHO.

LisaFC:
Ok, let me take a further look

LisaFC:
If you remove the overridden partial in this project you should see the problem: https://github.com/LisaFC/monolang

chalin:
I got the monolang repo working with Docsy at HEAD (using simple front matter) but, regardless, I'm feeling that your fix is the right way to go! 🎉  I have a tweaked version locally, and will share it + more feedback when I'm back.

layouts/partials/page-meta-links.html
chalin: Actually, our tests are passing under windows, so I'll assume that this is still ok. If not, we can address the regression later.

chalin (approved):
Great work! This definitely seems like the right way to fix this. There is some cleanup in the code that I'd like to do, but I'll address that in a followup PR, and add a statement about this being a breaking change to the CHANGELOG.

Issue #1753, opened 2023-11-25
NPM dependencies: Update font-awesome version which includes new Twitter logo


venkyvajrala:
Currently docsy doesn't support new Twitter logo X as font-awesome introduced it in 6.4.2 version.
Updating package version gives users to use new Twitter logo in their applications.

PR #1754, opened 2023-11-25
NPM dependencies: Update font-awesome package to include new twitter logo

venkyvajrala:
Fixes: #1753

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,6 @@ go 1.12
 // AUTO-GENERATED using `npm run get:hugo-modules`
 
 require (
-	github.com/FortAwesome/Font-Awesome v0.0.0-20230327165841-0698449d50f2 // indirect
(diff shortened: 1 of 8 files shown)


deining:
Thanks for your PR!
This PR takes care of the npm dependency. We have font-awesome as module dependency too, which also needs to be updated.
@venkyvajrala: can you add this missing part to the PR as well, please?

deining (changes requested):
Thanks for your PR!
This PR takes care of the npm dependency. We have font-awesome as module dependency too, which also needs to be updated.
@venkyvajrala: can you add this missing part to the PR as well, please?

venkyvajrala:
@deining Updated module dependency for font-awesome library.

chalin (approved):
LGTM, thanks 👍🏻 

FYI: In the past we've had patch updates to FA that affecting spacing, which is why I preferred waiting for 0.8.0 to be released first.

Issue #1781, opened 2024-01-13
[CI] Publish CHANGELOG and other top-level `*.md` files on the Docsy site and drop `markdown-link-check`


chalin:
I think that it's useful to have at least the CHANGELOG published to the Docsy website. If we publish all top-level `*.md` files we'll be able to eliminate the need for the extra link checking via `markdown-link-check`. We'll then be able to rely on `htmltest`'s reference caching capabilities.

I'm marking this is CI/infrastructure related since this is in part motivated replacing `markdown-link-check` by use of `htmltest`.

chalin:
And here's another reason: CI runs can fail because external link checking is inherently unstable. E.g., https://github.com/google/docsy/actions/runs/7513143585/job/20454576805:

```
  22 links checked.
  [✖] https://badges.netlify.com/api/docsydocs.svg?branch=main → Status: 0
Error: Process completed with exit code 123.
```

PR #1790, opened 2024-01-15
[CI] Fix markdown-link-check ignore pattern & NPM link vers update

chalin:
- Contributes to #1781 in that it fixes the `markdown-link-check` ignore pattern to remove the most flaky external link
- Updates links into NPM docs to access most recent version


chalin:
Merging. Comments are still welcome @emckean @deining @fekete-robert et al.

PR #1797, opened 2024-01-21
npm install: get rid of platform dependent cp command

deining:
This PR follows up on the discussion we have had in #1793.
It alters the mount points for npm boostrap dependency. After this change, the platform dependent `cp` command which does not run on linux, can be discarded, thus solving #1793.


chalin (commented):
Thanks for this PR! I'll take a look as soon as I can when back at the office the week of the 29th. In the meantime, could you ensure that all of the **GH actions pass**?

deining:
> Thanks for this PR! I'll take a look as soon as I can when back at the office the week of the 29th. In the meantime, could you ensure that all of the **GH actions pass**?

All GH actions are passing now :smile:. Looking forward to your comments!

chalin (commented):
Oh, I love this! ❤️✨ It is a much better solution, thanks for finding and proposing it @deining 🙏 

We need to drop `assets/_vendor`. I'll push that in and then merge if the tests still pass.

chalin:
Unless I misunderstood, this PR is meant to offer a replacement solution for the files under `assets/_vendor`, right? As you can see, with these files removed (via https://github.com/google/docsy/pull/1797/commits/2951dae1f78d285bb0ee386aeb6e6ef3e0115250), the smoke tests are failing: https://github.com/google/docsy/actions/runs/7740838481/job/21106641865?pr=1797. Can you look into it? Thanks!

deining:
> Unless I misunderstood, this PR is meant to offer a replacement solution for the files under `assets/_vendor`, right?

Yes, correct. In an ideal world, we could drop `assets/_vendor` indeed.

> As you can see, with these files removed (via [2951dae](https://github.com/google/docsy/commit/2951dae1f78d285bb0ee386aeb6e6ef3e0115250)), the smoke tests are failing: https://github.com/google/docsy/actions/runs/7740838481/job/21106641865?pr=1797. Can you look into it? Thanks!

But world is not always ideal and (go) software has bugs. The bug that affects us in this case is mentioned [here](https://raw.githubusercontent.com/google/docsy/main/assets/_vendor/README.md). And the corresponding mount instruction is here: 

https://github.com/google/docsy/blob/91efe359985017cfac3cc0cddf45dac79b3086ab/hugo.yaml#L45-L48

To make it short, unless we move [`_rfs.scss`](https://github.com/google/docsy/blob/main/assets/_vendor/bootstrap/scss/_rfs.scss) to some other place in the repo, we have to keep `assets/_vendor`. Moving `_rfs.scss` to another location in the repo seems possible, though, then we have to adjust the mount instruction, too. Afterwards we should be able to delete folder `assets/_vendor`.

chalin:
The whole point of the `_cp:bs-rfs` NPM script is to ensure that the file is automatically kept fresh. If we can't drop the local copy of `_rfs.scss`, then the mount points added in this PR are of no use: the UG already builds _without_ those extra mount points even if the `_cp:bs-rfs` script is removed.

Hmm, maybe #1806 offers a win-win solution. PTAL

Issue #1796, opened 2024-01-20
Smoke tests: feedback


deining:
Thanks for providing the smoke tests! They proved to be useful already!
Here comes some feedback:

1. Synopsis, retrieved via `./make-site.sh -h`

    ```
    Creates a Docsy-themed site under SITE_NAME using the Hugo new command.
     Docsy is fetched as an NPM package from google/docsy in GitHub,
     unless the -l or -d HUGO_MOD flags are used.
    ```

    I don't see a `-d` flag, however. I guess this should be `-s`?
    This is now addressed in #1800.

2. Small nit:

    ```
    ./make-site.sh -s foo
    ERROR: invalid argument to -s flag: FOO`
    ```
    This should be `foo`, not `FOO`. Can be fixed easily.
    This is now addressed in #1800.
3. After the hugo build, there is this line on STDOUT:
    ```
    + [[ -n 1 ]]
    ```
    What does that mean?
4. After the npm install, there is this line on STDOUT:
    ```
    + [[ NPM == HUGO* ]]
    ```
    What is that meant for?
4. Running this command fails
    ```
    ./make-site.sh -fv v0.5.1 -s HUGO
    ```
   This was to be expected, since module depencendies were altered in v0.8.0 (#1120). Can we check the version given here and raise and error of below v0.8.0 with module installation? An alternative would be to mention this limitation in the synopsis.

5. Inside the smoke tests there is a matrix defined so that tests can be reused and executed on Linux and Windows machines and for NPM and hugo module build. This seems logical, it comes with one disadvantage, though: as soon as a smoke test for Linux fails, the corresponding smoke test on Windows is cancelled. For me this turned out to be unfortunate, so I decided to create individual action files for the different environments (see this [commit](https://github.com/deining/docsy/commit/9184e12440970983a49be220711573f03f115555) in my docsy fork). Separate workflow also do have the advantage that individual smoke tests now can be easily enabled/disabled without touching the code.

chalin:
Thanks for the detailed feedback, I'll get to it as soon as I'm back at the office the week of the 29th.

chalin:
(1) & (2) Thanks for the fixes you made in #1800. They LGTM once the `name` field changes are reverted.

(3) & (4) As for the output starting with `+`, this is subcommand trace output that comes from `set -x`. This helps diagnose issues when tests fail. If you'd rather not see that, use the `-q` flag. I'd rather keep logging verbose by default in CI jobs though.

[...]

PR #1800, opened 2024-01-21
Smoke test: improvements

deining:
This PR brings a few improvements to the newly added smoke tests. Also see #1796 for an explanation.


chalin (commented):
I appreciate the fixes, thanks.

As for the `name` changes, I'd prefer that you revert them, otherwise the job lists become difficult to figure out at a glance, as you can see here:

> image

vs

> image

Ideally, there would be a `description` field, but this is only a proposal at the moment, see https://github.com/orgs/community/discussions/12213.

deining:
`name`changes are reverted, as requested.

chalin (approved):
LGTM, thx!

PR #1802, opened 2024-01-28
Bump dependencies to their latest versions

deining:
This PR bumps dependencies to their latest versions.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -6,7 +6,7 @@ publish = "userguide/public"
 command = "npm run docs-install && npm run build:preview"
 
 [build.environment]
-GO_VERSION = "1.21.4"
+GO_VERSION = "1.21.6"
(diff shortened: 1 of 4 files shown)


chalin (approved):
Thank you for the updates.

I can confirm that the UG generates as expected (only one surprise additionally generated `</p>` closing tag, but that seems valid):

```console
$ (cd public && git diff -bw --ignore-blank-lines -I 'Hugo 0.12' -I 'meta name="twitter:|meta property="og:image"|meta itemprop="image"') | grep ^diff | head | grep -Ev '_print|\.xml$'
diff --git a/docs/adding-content/feedback/index.html b/docs/adding-content/feedback/index.html
diff --git a/docs/get-started/docsy-as-module/installation-prerequisites/index.html b/docs/get-started/docsy-as-module/installation-prerequisites/index.html
```

PR #1806, opened 2024-02-01
[CI] Replace cp command by fs.copyFile + add test-all script

chalin:
- Proposed solution to #1793
- Proposed alternative to #1797

/cc @marshalc @deining


chalin:
Closing this in favor of another solution.

PR #1815, opened 2024-02-02
Link styles: follow recommended practice and underline by default

### LINKED ISSUE #1814: Use Bootstrap's default link style, which is recommended
- From #1684:
  > As per the change upstream in Bootstrap 5 [twbs/bootstrap#30389](https://github.com/twbs/bootstrap/pull/30389)
  > 
  > Removing the underline from links was in vogue for a while, but WebAIM [recommends not doing that](https://webaim.org/techniques/hypertext/link_text#underlining).
  > 
  > > Browsers underline hypertext links by default. It is possible to remove the underline using Cascading Style Sheets (CSS), but this is a bad idea most of the time.
- I agree with @froboy that Docsy should stick to the Bootstrap & recommended default.
- This is in support of:
  - #1655

chalin:
- Fixes #1814
- Closes #1684 by superseding it

**Preview**, e.g.: https://deploy-preview-1815--docsydocs.netlify.app/docs/

---

Projects that wish to preserve Docsy's old behavior can add the following to their `assets/scss/_variables_project.scss`, among other ways:

```scss
$link-decoration: none !default;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,12 +38,19 @@ For the full list of changes, see the [0.9.0] release notes.
   [path_base_for_github_subdir]. Projects will need to adjust the value of
   [path_base_for_github_subdir] to be relative to the file's physical location.
 
-**New**:
+- Docsy statically generates anchor-links after headings using Hugo's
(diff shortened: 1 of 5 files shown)


deining (approved):
This change is fine for me.

PR #1827, opened 2024-02-03
[CI] Smoke tests: fix for repo forks, but also disable on push

### LINKED ISSUE #1826: [CI] Smoke tests failing in repo forks
For example, see https://github.com/chalin/docsy/actions/runs/7766711140.

chalin:
- Fixes #1826
- Fixes `tools/make-site.sh` so that it can be run on repo forks, but ...
- Restricts the execution of smoke jobs on push to the main repo -- we don't need tests to run again on forks.


chalin:
I'm not 100% sure that this approach will work given the use of the script's fail-fast flag, but let's give it a spin.

PR #1834, opened 2024-02-04
Issue template 'enhancement': Fix error

deining:
This is a follow up on merged PR #1824.
As you can see [here](https://github.com/google/docsy/blob/main/.github/ISSUE_TEMPLATE/enhancement.md), there is an error with the issue template `enhancement.md`:

```
There is a problem with this template

About can't be blank. 
```

This PR fixes this error.


chalin (approved):
👍🏻

Issue #1196, opened 2022-08-23
Blog pages should contain `page-meta-lastmod.html` too


chalin:
@LisaFC et all: are you aware of any reason why blog pages don't contain "last mod" info like regular doc pages do?

Some context:

- `_default`: https://github.com/google/docsy/blob/519930cf0d642c37941b74a25361e743165f94bc/layouts/_default/content.html#L19
- The blog [`content.html`](https://github.com/google/docsy/blob/main/layouts/blog/content.html) override omits `page-meta-lastmod.html`

If there's no reason that you're aware of to omit that info at the bottom of the page, I suggest that we add it. Thoughts?

chalin:
I've quickly experimented with this and it looks best if we end the `<div>` in https://github.com/google/docsy/blob/main/layouts/blog/content.html with `page-meta-lastmod.html` last:

```
{{ partial "pager.html" . }}
{{ partial "page-meta-lastmod.html" . }}
```

For example:

> <img width="782" alt="image" src="https://user-images.githubusercontent.com/4140793/186265458-332ba00c-1958-445f-87dc-6848ec173316.png">

LisaFC:
Happy to add it if you think it's useful - I think the main reason we didn't was that blog posts tend not to be updated (unless someone discovers an error post-publication) and have a publication date already. If they're being regularly updated they probably have content that should be in regular user docs. :)

Can't remember the issue/PR I commented on but I also think we should probably have a different "page meta links" for blog posts as most projects probably don't want to be inviting other users to edit guest blog posts. Thoughts?

PR #1857, opened 2024-02-09
Smoke tests: don't cancel Windows build after an error in Linux build

deining:
This PR follows up on #1196, issue 6 raised:

> Inside the smoke tests there is a matrix defined so that tests can be reused and executed on Linux and Windows machines and for NPM and hugo module build. This seems logical, it comes with one disadvantage, though: as soon as a smoke test for Linux fails, the corresponding smoke test on Windows is cancelled.

This PR overcomes this limitation while still following the DRY principle.


chalin (approved):
Works for me, thanks!

Issue #1812, opened 2024-02-02
Release 0.9.1 or 0.10.0 preparation


chalin:
### Issues/PRs in scope

- [ ] ...

### Candidate issues/PRs in scope

- [ ] TBC

Also see:

- [milestone 24Q1](https://github.com/google/docsy/milestone/10)

### Release-prep checklist

For the complete instruction set, see [Publishing a release](https://github.com/google/docsy/blob/main/CONTRIBUTING.md#publishing-a-release):

- [ ] Update Changelog: for inspiration, go over changes since last release https://github.com/google/docsy/compare/v0.8.0...main
- [ ] Update the Docsy version in `package.json` and [userguide/hugo.yaml](https://github.com/google/docsy/blob/main/userguide/hugo.yaml)
- [ ] Test pre-release from branch: 👍🏻 
- [ ] Pre-release patches?
- [ ] Create and push new version tag
- [ ] Create and publish the new release 🎉

Docsy example followup (tracked separately):

- `https://github.com/google/docsy-example/pull/xxx`

### Related

- #1759
- https://github.com/google/docsy/compare/v0.9.0...main

PR #1867, opened 2024-02-16
Release v0.9.1 preparation

chalin:
I wouldn't normally put out a patch release so quickly, but I'm OOO next week, so here it is!

Related:

- #1812
- https://github.com/google/docsy/compare/v0.9.0...main


deining (approved):
LGTM

PR #1885, opened 2024-03-06
Mermaid diagram with diagram specific settings: fix 'uncaught TypeError'

### LINKED ISSUE #1882: Mermaid graph with no parameters defined in hugo config results in 'Uncaught TypeError'
https://github.com/google/docsy/blob/a72c57c9398b9795daaeac6f6944369cb76d5adb/layouts/partials/scripts/mermaid.html#L21

If nothing is defined in hugo.toml for [params.mermaid] then the referenced code results in `var params = null` being generated.

https://github.com/google/docsy/blob/a72c57c9398b9795daaeac6f6944369cb76d5adb/layouts/partials/scripts/mermaid.html#L29

Then when attempting the if statement a call is made against a null value, resulting in the browser console error.

`Uncaught TypeError: Cannot read properties of null (reading 'hasOwnProperty')`

deining:
This PR fixes the issue raised in #1882.
This PR closes #1882.

### DIFF
--- a/layouts/partials/scripts/mermaid.html
+++ b/layouts/partials/scripts/mermaid.html
@@ -18,7 +18,7 @@ import mermaid from "{{ $cdnurl }}";
         return;
     }
 
-    var params = {{ .Site.Params.mermaid | jsonify | safeJS }};
+    var params = {{ with .Site.Params.mermaid }}{{ . | jsonify | safeJS }}{{ else }}{}{{- end }};
 
     // Site params are stored with lowercase keys; lookup correct casing
     // from Mermaid default config.
--- a/userguide/content/en/docs/adding-content/diagrams-and-formulae/index.md
+++ b/userguide/content/en/docs/adding-content/diagrams-and-formulae/index.md
@@ -388,18 +388,18 @@ By default, docsy pulls in the latest officially released version of Mermaid at 
 {{< tab header="Configuration file:" disabled=true />}}
 {{< tab header="hugo.toml" lang="toml" >}}
 [params.mermaid]
-version = "10.8.0"
+version = "10.9.0"
 {{< /tab >}}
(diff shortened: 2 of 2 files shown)


layouts/partials/scripts/mermaid.html
```diff
@@ -10,15 +10,19 @@
 {{ end -}}
 
 <script type="module" async>
-import mermaid from "{{ $cdnurl }}";
+import mermaid from "{{- $cdnurl -}}";
```
chalin, line 13: I don't believe that this edit has any effect since there is no whitespace to trim outside of the double braces since they are immediately (each) next to a double quote.

layouts/partials/scripts/mermaid.html
```diff
@@ -10,15 +10,19 @@
...
         mermaid.initialize({startOnLoad: false});
         return;
     }
 
-    var params = {{ .Site.Params.mermaid | jsonify | safeJS }};
+    {{ with .Site.Params.mermaid -}}
+      var params = {{ . | jsonify | safeJS -}};
+    {{ else -}}
+      var params = {};
+    {{- end -}}
```
chalin, line 25: How about:

```suggestion
    var params = {{ with .Site.Params.mermaid }}{{ . | jsonify | safeJS }}{{ else }}{}{{ end }};
```

chalin (commented):
Thanks @deining. See inline for suggested edits.

layouts/partials/scripts/mermaid.html
deining: You are right. Reverted to original version.

layouts/partials/scripts/mermaid.html
deining: Changed as proposed.

chalin (approved):
LGTM, and thanks for the fix ✨

PR #1890, opened 2024-03-11
Fix section index partial NPE

### LINKED ISSUE #1874: NPE for `$parent.File.UniqueID` using Hugo 0.123.1.
_Originally posted by @chulcher in https://github.com/google/docsy/issues/1863#issuecomment-1958211996_

> This appears to be further broken with Hugo 0.123.1.

> ERROR render of "taxonomy" failed: 
"C:\Users\Carson\AppData\Local\hugo_cache\modules\filecache\modules\pkg\mod\github.com\google\docsy@v0.9.1\layouts\docs\list.html:12:5": execute of template failed: template: docs/list.html:12:5: executing "main" at <partial "section-index.html" .>: error calling partial: "C:\Users\Carson\AppData\Local\hugo_cache\modules\filecache\modules\pkg\mod\github.com\google\docsy@v0.9.1\layouts\partials\section-index.html:8:69": execute of template failed: template: partials/section-index.html:8:69: executing "partials/section-index.html" at <$parent.File.UniqueID>: error calling UniqueID: runtime error: invalid memory address or nil pointer dereference
> Built in 936 ms
> Error: error building site: render: failed to render pages: render of "taxonomy" failed: "C:\Users\Carson\AppData\Local\hugo_cache\modules\filecache\modules\pkg\mod\github.com\google\docsy@v0.9.1\layouts\docs\list.html:12:5": execute of template failed: template: docs/list.html:12:5: executing "main" – File is nil; wrap it in if or with: {{ with partial "section-index.html" .>: error calling partial: "C:\Users\Carson\AppData\Local\hugo_cache\modules\filecache\modules\pkg\mod\github.com\google\docsy@v0.9.1\layouts\partials\section-index.html:8:69": execute of template failed: template: partials/section-index.html:8:69: executing "partials/section-index.html" at <.File }}{{ .UniqueID }}{{ end }}

https://github.com/gohugoio/hugo/issues/12068, specifically https://github.com/gohugoio/hugo/issues/12068#issuecomment-1953287052 relevant?

tobiaskohlbau:
During rendering the section index of a docs list it could happen that a file has no parent for e.g. the homepage. If such scenario is existing the .Parent.File of a file is nil. Therefore this case should be handled within the rendering itself.

Fixes #1874

### DIFF
--- a/layouts/partials/section-index.html
+++ b/layouts/partials/section-index.html
@@ -4,7 +4,8 @@
     {{ $pages = (where $pages "Type" "!=" "search") }}
     {{ $pages = (where $pages ".Params.hide_summary" "!=" true) -}}
     {{ $pages = (where $pages ".Parent" "!=" nil) -}}
-    {{ if and .Parent .Parent.File -}}
+    {{ $pages = (where $pages ".Parent.File" "!=" nil) -}}
+    {{ if and $parent $parent.File -}}
         {{ $pages = (where $pages "Parent.File.UniqueID" "==" $parent.File.UniqueID) -}}
     {{ end -}}
     {{ if or $parent.Params.no_list (eq (len $pages) 0) -}}


deining (changes requested):
Just ran into the same issue, and the proposed fix solved my problem.
I realized that this [contribution](https://github.com/gohugoio/hugo/issues/12263#issuecomment-2000230210) proposes a slight more elegant solution to the problem .
@tobiaskohlbau: can you have a look, please? Can we take over this solution here?

tobiaskohlbau:
> Just ran into the same issue, and the fix proposed with this PR solved my problem. I realized that this [contribution](https://github.com/gohugoio/hugo/issues/12263#issuecomment-2000230210) proposes a slight more elegant solution to the problem . @tobiaskohlbau: can you have a look, please? Can we take over this solution here?

AFAIK the proposed solution would not work. As the _and_ within the if only checks for the currently scoped Parent but not for every pages `Parent.File.UniqueID`. Therefore if pages do not have a `Parent.File` the access to _UniqueID_ would throw a NPE. It's required to filter out all pages where _Parent.File=nil_. (IIRC I had such solution before and it still thread said NPE.)

yann-soubeyrand:
Hello,

I confirm what @tobiaskohlbau said.

Using Hugo 0.124.1, which contains the fix referred to above, on a docs only site, that is with the following front matter in `content/_index.adoc`

```
title: 'Documentation'
layout: 'section'

cascade:
 - type: 'docs'
```

I still get a warning:

```
WARN  .File.UniqueID on zero object. Wrap it in if or with: {{ with .File }}{{ .UniqueID }}{{ end }}
```

But the warning is solved when adding the modified partial from this PR.

yann-soubeyrand:
I re-tested today and, in fact, this is not a warning, but an error:
```
ERROR render of "taxonomy" failed: "/home/yann/.cache/hugo/modules/filecache/modules/pkg/mod/github.com/camptocamp/docsy@v0.9.2-0.20240325160920-2e674ae180f3/layouts/docs/list.html:12:5": execute of template failed: template: docs/list.html:12:5: executing "main" at <partial "section-index.html" .>: error calling partial: "/home/yann/.cache/hugo/modules/filecache/modules/pkg/mod/github.com/camptocamp/docsy@v0.9.2-0.20240325160920-2e674ae180f3/layouts/partials/section-index.html:8:69": execute of template failed: template: partials/section-index.html:8:69: executing "partials/section-index.html" at <$parent.File.UniqueID>: error calling UniqueID: runtime error: invalid memory address or nil pointer dereference
Built in 3662 ms
Error: error building site: render: failed to render pages: render of "taxonomy" failed: "/home/yann/.cache/hugo/modules/filecache/modules/pkg/mod/github.com/camptocamp/docsy@v0.9.2-0.20240325160920-2e674ae180f3/layouts/docs/list.html:12:5": execute of template failed: template: docs/list.html:12:5: executing "main" – File is nil; wrap it in if or with: {{ with partial "section-index.html" .>: error calling partial: "/home/yann/.cache/hugo/modules/filecache/modules/pkg/mod/github.com/camptocamp/docsy@v0.9.2-0.20240325160920-2e674ae180f3/layouts/partials/section-index.html:8:69": execute of template failed: template: partials/section-index.html:8:69: executing "partials/section-index.html" at <.File }}{{ .UniqueID }}{{ end }}
```

I still confirm this PR fixes the issue.

deining:
> AFAIK the proposed solution would not work. As the _and_ within the if only checks for the currently scoped Parent but not for every pages `Parent.File.UniqueID`. Therefore if pages do not have a `Parent.File` the access to _UniqueID_ would throw a NPE. It's required to filter out all pages where _Parent.File=nil_. (IIRC I had such solution before and it still got said NPE.)

You are right. Your solution is more general and should be merged unchanged.

deining (approved):

layouts/partials/section-index.html
```diff
@@ -4,7 +4,8 @@
     {{ $pages = (where $pages "Type" "!=" "search") }}
     {{ $pages = (where $pages ".Params.hide_summary" "!=" true) -}}
     {{ $pages = (where $pages ".Parent" "!=" nil) -}}
-    {{ if and .Parent .Parent.File -}}
+    {{ $pages = (where $pages ".Parent.File" "!=" nil) -}}
+    {{ if and $parent $parent.File -}}
```
chalin, line 8: After looking at this more closely, it seems to make sense indeed.

Btw, the `$parent` variable is misnamed, because`$parent` is set to `.Page`.

Can we inline the variable instead?

Btw, IMHO `.Page` cannot be falsy, but `.Page.File` could be falsy. So the if condition can be simplified.

chalin (approved):
I'm willing to give this a try.
RSLGTM based on appraisals by @deining and @yann-soubeyrand.
Thanks all! ✨

layouts/partials/section-index.html
chalin: Will track separately:

- #1946

Issue #1528, opened 2023-05-31
Upgrade to Bootstrap 5.3


chalin:
- BS 5.3 changes are significant enough that, IMHO, we'll want to careful walkthrough of the [5.3 migration](https://getbootstrap.com/docs/5.3/migration/) details before upgrading Docsy to BS 5.3.
- [Older comment] After we've released and sufficiently vetted #470, we should upgrade to 5.3. **Color modes** have changed, and might help address some Docsy feature requests.

Tasks that we might be able to address at the same time:

- [x] #1845

## Base & prep

- [x] #1563

### [Color modes](https://getbootstrap.com/docs/5.3/migration/#color-modes-1)

- [x] Color modes replace dark variants for components, so `.btn-close-white`, `.carousel-dark`, `.dropdown-menu-dark`, and `.navbar-dark` are deprecated - #1896

### [Utilities](https://getbootstrap.com/docs/5.3/migration/#utilities-1)

- [x] Deprecated the `.text-muted` utility and `$text-muted` Sass variable. It’s been replaced by `.text-body-secondary` and `$body-secondary-color` - #1892

## Related

- Last PR preview using BS 5.2: https://deploy-preview-1886--docsydocs.netlify.app/
- https://github.com/twbs/bootstrap/releases/tag/v5.3.3
- #470
- #1510

chalin:
Done AFAICT. Closing. Will reopen if anything comes up.

PR #1896, opened 2024-03-14
[BS5.3] Replace `.navbar-dark` by data attribute and other style adjustments

chalin:
- Contributes to #1528
- Replaces `.navbar-dark` by `data-bs-theme="dark"` data attribute
- Adjusts search element styles to work with the new navbar's dark mode _and_ in the left-nav too (not in dark mode)

**Preview**: https://deploy-preview-1896--docsydocs.netlify.app/docs/

### Screenshots

Light mode (default):

>  image

> image

Dark mode:

> image

> image

### DIFF
--- a/assets/scss/_search.scss
+++ b/assets/scss/_search.scss
@@ -27,7 +27,7 @@
     // Styling adjustments for the navbar
     @at-root {
       .td-navbar & {
-        color: $navbar-dark-color;
+        color: inherit;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


fekete-robert:
When using a dark system theme under firefox the text in the navbar search is dark (tested on mac and linux):

Screenshot from 2024-03-16 09-36-39

Using Brave (a chrome derivative) it's fine on both systems.

But this is probably an edge case, since firefox has a marginal market share, so we can consider it a known issue and go forward.

fekete-robert (approved):

chalin:
Thanks for that careful cross-browser check. Let's track that issue via:

- #1906

PR #1900, opened 2024-03-16
Address .Site.Author deprecation warning

### LINKED ISSUE #1899: Site param .Site.Author was deprecated in Hugo 0.124.0
We should address the following deprecation warning:

```console
$ npx hugo --logLevel info | grep deprecate
INFO  deprecated: .Site.Author was deprecated in Hugo v0.124.0 and will be removed in a future release.
```

According to this [comment](https://github.com/gohugoio/hugo/pull/11569#issue-1946296332), it should be replaced with `.Site.Params.Author`.

deining:
- Switches to use of `.Site.Params.Author`
- Fixes #1899
- **Preview**: https://deploy-preview-1900--docsydocs.netlify.app/docs/adding-content/feedback/#setup


chalin (commented):
Thanks for the update.

Why don't we just use the default Hugo template? /cc @LisaFC @emckean in case you might recall why.

chalin:
> Why don't we just use the default Hugo template? /cc @LisaFC @emckean in case you might recall why.

I've created an issue to track this, see:

- #1926

Issue #1558, opened 2023-06-09
Search in multilingual sites should preserve the language


chalin:
Upstreamed from:

- https://github.com/cncf/glossary/issues/2087

### Steps to reproduce the problem

- Visit https://example.docsy.dev/no/docs/
- Enter some text (e.g. `java`) in the search box and type <kbd>Enter</kbd>
- The page is redirected to:
  https://example.docsy.dev/search/?q=java
- **Expected behavior?** It should instead have redirected to:
  https://example.docsy.dev/no/search/?q=java

I'm opening the issue here since I'm quite sure the problem is with Docsy and not the example site.

/cc @cjyabraham

PR #1904, opened 2024-03-24
Fixes issue with the search input box always taking the user to default language page

cjyabraham:
Fixes issue with the search input box always taking the user to the default language search results (https://github.com/google/docsy/issues/1558)


chalin:
Thanks for this PR @cjyabraham. FYI, this is still on the radar. I'll add it as a candidate to release 0.11.0 (#1944). Docsy 0.10.0 is just hot off the press! 🔥

PR #1909, opened 2024-03-28
Color-mode / dark-mode support

chalin:
- Contributes to #331
- The Bootstrap docs invite us to copy their color-theme picker code. Well, I did just that :)
- Adds a color theme picker partial
- Illustrates the use of `` in the UG, which causes the color-theme picker to be added to the end of the navbar, just before the search box
- I'll create a changelog entry in a separate PR. My intent is to make clear that this early support is ~**experimental**~ in alpha, and still subject to change.

**Preview**: https://deploy-preview-1909--docsydocs.netlify.app/docs/

### Screenshots

> image

> image

Issue #331, opened 2020-08-25
Support dark mode


eXpire163:
Providing a dark version of https://github.com/google/docsy/blob/master/assets/scss/_variables.scss would be great

---

**Edits** (@chalin)

### Prep

- [x] #1908

### Tasks

- [x] #1909
- [x] #1913
- [x] #1916
  #1917
- [x] #1919
- [x] #1914
- [x] #1958
- [x] #1959
  - [x] #1974
- [x] #1976
- [x] #1957

Adjustments that will be required to Docsy element styling in support of Bootstrap's dark mode:

- [x] Search input to for for navbar; and left-nav in both light and dark modes
  #1896
  - [x] #1906 
- [x] Left nav entry colors -- the text style isn't appropriate for dark mode
  #1908 
- [x] Alert backgrounds -- the bg style isn't appropriate for dark mode
  #1908 
- [x] Card background -- e.g., see [Shortcode card: programming code](https://deploy-preview-1909--docsydocs.netlify.app/docs/adding-content/shortcodes/#shortcode-card-programming-code) - #1922
- [x] [Pageinfo](https://deploy-preview-1909--docsydocs.netlify.app/docs/adding-content/shortcodes/#pageinfo) backgrounds - #1915
- [x] [Tabbed pane](https://deploy-preview-1909--docsydocs.netlify.app/docs/adding-content/shortcodes/#tabbed-panes) backgrounds - #1920
- [x] #1952
  - [x] #1927
  - [x] #1943
- [x] #1961
- [x] `.td-box-*`:
  - [x] #1967
  - [x] #1970
- [x] #1981

huehnerlady:
Any update on this? Would love that feature to be available 😍

frodriguezsmartclip:
Someone could add that feature to Docsy, It seems not to hard to do it: https://github.com/GoogleContainerTools/skaffold/pull/5364/commits/ccdffd1e27785b203f953dd9b8b8735972ca5d99

If I have time, I'll try to code it ⚡

bwplotka:
Help wanted, that would be pretty epic to see! 🤗

narrenfrei:
I have tried to implement the above mentioned - I would call it - "darkmode hack" ;-) and made a PR with the WIP flag #563 
You can have a look on the result at: https://deploy-preview-563--docsydocs.netlify.app/

[...]

PR #1913, opened 2024-04-05
[Dark-mode] Tweak of param, file and resource names

chalin:
- Contributes to #331
- Followup to #1909
- Renames BS color-modes.js to dark-mode.js
- Renames local-storage key so that it's Docsy specific
- Relocates the light/dark-mode drop-down menu styles into the SCSS assets folder
- Renames the config param to `showLightDarkModeMenu`

/cc @geriom


chalin:
Going to merge now; more changes will follow.
@geriom, et al.: feedback is still welcome.

PR #1923, opened 2024-04-07
Drop deprecated shortcode 'card-code'

### LINKED ISSUE #1921: Drop `card-code` in 0.10.0
`card-code` was deprecated in 0.7.0 -- [CHANGELOG#070](https://github.com/google/docsy/blob/main/CHANGELOG.md#070), I think that we should consider dropping it in 0.10.0.

(I believe that this matches Hugo's new deprecation policy.)

WDYT @deining?

I don't know if the UG would also need to be udpated if we drop the `card-code` shortcode?

deining:
This PR closes #1921.


CHANGELOG.md
```diff
@@ -26,7 +26,8 @@ For the full list of changes, see the [0.x.y] release notes.
 
 **Breaking changes**:
 
-- ...
+- Removed deprecated shortcode `card-code`; use shortcode `card` with named
+  parameter `code=true` instead.
```
chalin, line 30:
```suggestion
- Removed shortcode `card-code` that was deprecated in [0.7.0](#070); use
  shortcode `card` with named parameter `code=true` instead.
```

chalin (approved):
I resolved the conflict and tweaks the CL entry. Thanks!

chalin:
Hmm, it builds fine locally. I'm going to merge and see.

Issue #1894, opened 2024-03-14
Site param .Site.DisqusShortname was deprecated in Hugo 0.120.0


deining:
We should address the following deprecation warning:

```console
$ npx hugo --logLevel info | grep deprecate
INFO  deprecated: .Site.DisqusShortname was deprecated in Hugo v0.120.0 and will be removed in a future release. Use .Site.Config.Services.Disqus.Shortname instead.
```

PR #1934, opened 2024-04-17
Fix use of deprecated param .Site.Params.DisqusShortname

deining:
This is a follow-up to my merged PR #1895.
It switches to the recommended use of `.Site.Config.Services.Disqus.Shortname` in two more places.
For background info, see: #1894


chalin (approved):
👍🏻

PR #1935, opened 2024-04-17
Introduce parameter to exclude site title from the Title attribute

samiahmedsiddiqui:
In one of my website, I would like to remove the site title from the HTML title element without overriding `head.html` file. This PR allows to do that by just adding `excludeSiteTitle = true` in the params settings.


fekete-robert:
I think there was a discussion about this (or something very similar) somewhere in the past, and the conclusion was that this can be achieved from the project CSS. (By setting `navbar-brand__name` to `display: none`)

samiahmedsiddiqui:
@fekete-robert I am talking about HTML `<title>` tag which is displayed at the top bar of the browser and I don't think that there is anyway to use CSS on that. Secondly, I don't want to hide the title, I just want to remove the site title from the `<title>` tag which is added by Docsy.

fekete-robert:
Sorry, I misunderstood, my bad.

samiahmedsiddiqui:
@chalin Please take a look of this PR.

fekete-robert (approved):
LGTM

samiahmedsiddiqui:
@fekete-robert @chalin Just wanted to nudge about the PR. Any chance we can get it merged soon?

huanlin:
Hello,
The `<title>` element in the HTML seems reasonable and no harm. So I'm wondering in what situation we'll need to remove it from a page's HTML?

samiahmedsiddiqui:
@huanlin I am not saying to remove/hide `title` element in HTML. I am just saying to introduce an option to remove site title/name from the `title` element in HTML.

### Example of `title` element with site title/name:

```html
<title>Use Docsy as a Hugo Module | Docsy</title>
```

### Example of `title` element without site title/name:

```html
<title>Use Docsy as a Hugo Module</title>
```

huanlin:
> ### Example of `title` element with site title/name:
> ```
> <title>Use Docsy as a Hugo Module | Docsy</title>
> ```
> 
> ### Example of `title` element without site title/name:
> ```
> <title>Use Docsy as a Hugo Module</title>
> ```
 
Ah, I see.  Thank you for your explanation. It makes sense to me :)

huanlin:
Just to add one little idea: the name `excludeSiteTitle` may lead to questions like "What does this parameter do?"
I think it would be better if it has a more intuitive name. For example:  `noSiteTitleInHtmlTitleElement` 
I don't know, maybe you can come out a better name.
Just an idea :)

samiahmedsiddiqui:
I have no issues in changing the parameter name but I think it is sufficient to understand the purpose of this parameter for the developers.

But I am open to change. Let's wait for others feedback.

samiahmedsiddiqui:
@chalin @fekete-robert When can I expect this PR to be merged?

chalin:
- The utility was questioned in review and no consensus formed. Thanks.

PR #1933, opened 2024-04-16
Bump algolia docsearch to latest version

deining:
This PR bumps algolia docsearch to latest released version 3.6.0.

PR #1940, opened 2024-04-23
Bump KaTeX to latest released version

deining:
This PR bumps KaTeX to latest released version 0.16.10.
It also finalized #1933 by upgrading the javascript component of algolia docsearch.


chalin (approved):
🙏👍🏻

PR #1947, opened 2024-04-25
section-index.html code cleanup

### LINKED ISSUE #1946: section-index.html: code cleanup
_Originally posted by @chalin in https://github.com/google/docsy/pull/1890#discussion_r1555101226_

> After looking at this more closely, it seems to make sense indeed.
> 
> Btw, the `$parent` variable is misnamed, because`$parent` is set to `.Page`.
> 
> Can we inline the variable instead?
> 
> Btw, IMHO `.Page` cannot be falsy, but `.Page.File` could be falsy. So the if condition can be simplified.

chalin:
- Followup to #1890
- Renames `$parent` to `$page`
- Fixes #1946
- Wraps up #1874

/cc @yann-soubeyrand @tobiaskohlbau @chulcher @marshalc

### DIFF
--- a/layouts/partials/section-index.html
+++ b/layouts/partials/section-index.html
@@ -1,16 +1,16 @@
 <div class="section-index">
-    {{ $parent := .Page -}}
+    {{ $page := .Page -}}
     {{ $pages := (where .Site.Pages "Section" .Section).ByWeight -}}
     {{ $pages = (where $pages "Type" "!=" "search") }}
     {{ $pages = (where $pages ".Params.hide_summary" "!=" true) -}}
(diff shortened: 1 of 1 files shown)


deining (approved):
Just checked this simplification on an existing site that caused problems in the past and it worked well, so LGTM.

tobiaskohlbau (approved):
Sorry for the late response in my PR was quite busy the last couple of days. Checked out this PR and tested it. Works like expected for the site having issuers earlier.

fekete-robert (approved):
LGTM

chalin:
Thanks all for your reviews and feedback.

PR #1931, opened 2024-04-16
Google analytics: drop deprecated async template

deining:
This PR fixes #1930.

PR #1949, opened 2024-04-26
remove reference to non-existent Google Analytics partial

dmah42:
it looks like the non-gtag Google Analytics partial was removed (makes sense, given the GA migration to gtag) but the head partial is still trying to reference it.

this causes build errors (https://discourse.gohugo.io/t/error-error-building-site/49340).


chamiz:
I hope Docsy maintainers team will make a release soon. With Hugo version .0125.0+ we are facing issue with Google Analytics page

chalin:
Thanks @dmah42 but this has already be resolved:

- #1931

chalin:
> I hope Docsy maintainers team will make a release soon. With Hugo version .0125.0+ we are facing issue with Google Analytics page

@chamiz - yes, a release should be out within the next week or so. In the meantime there are several interim solutions you can try, not the lease of

chalin:
@chamiz, see:

- #1973

chamiz:
Thanks @chalin. I did changes in the local. However, this is for Selenium documentation where we use automated build. 
For now, I downgraded Hugo to v0.124.1.

PR #1991, opened 2024-05-07
Update index.md

AdamFermier:
Just wanted to add an example to show it quite clearly for new folks like me.


chalin:
Hi @AdamFermier 👋🏻. Thanks but we decided to drop the `iframe` example from the docs a while back. For details, see https://github.com/google/docsy/pull/1475#issuecomment-1534879936.

PR #2006, opened 2024-05-16
Cleanup of Google analytics templates

### LINKED ISSUE #1932: head.html: remove outdated internal template (google analytics v3)
The template `head.html` still references the internal template for google analytics v3:

https://github.com/google/docsy/blob/b95a4948ee42e601d2685a334e0b2362e31820ef/layouts/partials/head.html#L50-L51

Google Analytics v4 is the current version, and the referenced template is not part of the hugo repo any, as seen [here](https://github.com/gohugoio/hugo/tree/master/tpl/tplimpl/embedded/templates).
Therefore this outdated template should be removed from the partial.

chalin:
- Fixes #1932
- Drops now obsolete sections of the UG related to custom processing of GA site IDs
- Temporarily adds a new internal template, which will be dropped if/when is merged and then released


chalin:
@LisaFC @deining @fekete-robert - going to merge this now, ahead of the RTL work so that I can test it in production. Feedback, if any will be addressed in a followup PR.

Issue #2011, opened 2024-05-19
The menu does not switch sides when the site changes to RTL


mboukhalfa:
### Problem
When switching to Persian (RTL lang) the menu does not change direction as shown in the pictures when we switch to RTL the logo and the search input switch the side but the menu stick to its place instead of moving to left side.

ref: https://github.com/twbs/bootstrap/issues/22780#issuecomment-307592587
![Screenshot from 2024-05-19 04-04-38](https://github.com/google/docsy/assets/19722896/0e8d2d5f-14d5-4b20-9a17-c910ec278900)
![Screenshot from 2024-05-19 04-04-54](https://github.com/google/docsy/assets/19722896/3c5e0016-ff89-41e6-ab4a-e791693699ff)

### The root of the issue
`ms-md-auto` class to push the menu always to the right does not support direction change

chalin:
For the record, this was reported earlier too:

- #829

chalin:
For a confirmation of the fix, see:

- https://github.com/google/docsy-example/pull/305

/cc @theletterf

PR #2012, opened 2024-05-19
Adapt navbar to support RTL

mboukhalfa:
Replace bootstrap classes `flex-column flex-md-row` and `ms-md-auto` by `w-100` and ` justify-content-end` which have better support for RTL

Fix: https://github.com/google/docsy/issues/2011

TODO:

- [ ] Improve small-device responsivity


mboukhalfa:
/hold
 when switching to small device the behavior of the logo changed

PR #2027, opened 2024-05-29
[RTL] Warn if RTL CSS not found + Hugo mod fix + more

### LINKED ISSUE #2025: Ensure RTL works when Docsy is used via Hugo modules

chalin:
- Fixes #2025
- `ci:prepare` now copies over a cached copy of the Bootstrap RTL CSS
- Adjusts the head partial:
  - Warn if Bootstrap RTL CSS is needed for some pages but not found; and embeds the warning as a `<meta name="docsy-info"` element of affected HTML pages
  - Report an error if the main CSS file isn't found
- Adds a mount point for Hugo-module-based projects so that the BS RTL file
- `CONTRIBUTING.md` typo fix


chalin:
Merging now, but feedback is still welcome @LisaFC @geriom @emckean @fekete-robert @deining. I'll be testing this soon on the docsy-example.

Issue #2017, opened 2024-05-27
Unable to use swagger UI short code


alexandre-touret:
### Environment

- [X] Version of Docsy you are using: 0.10.0
- [x] How are you using Docsy? As a: 
  - [X] Hugo module
  - [ ] NPM module
  - [ ] Git submodule
  - [ ] Other:
- [x] Version of Hugo you are using (output of `hugo version`): ``hugo v0.125.6-69ede10edcd539380914bbee58d4d32953dd8b43+extended linux/amd64 BuildDate=2024-05-05T10:52:52Z VendorInfo=gohugoio``
- [x] OS:
  - [X] Linux (Ubuntu 24.04)
  - [ ] macOS
  - [ ] Windows
  - [ ] Other:

### Problem

The SwaggerUI shortcode doesn't work anymore.

On my browser's console (I also tested it on several browsers), I have this error:

```
Uncaught ReferenceError: SwaggerUIBundle is not defined
    onload http://localhost:1313/docs/test-swagger/:227
```

Then, when I check the network console, I can see the following issues:

![image](https://github.com/google/docsy/assets/943841/2543397f-1803-4b1b-a2cd-5a4daff7ff61)

It seems the URL used (e.g., https://unpkg.com/swagger-ui-dist@5.1.0/swagger-ui.css) to get the JS artifacts are now blocked.

I tested it on a fresh setup of the docsy example.
The JS and CSS are indeed not unpackaged in the public folder.

Here is the way to reproduce this issue:
* Checkout the docsy example as explained on the documentation (w/ hugo modules)
* Create the folder ``my-new-site/content/en/docs/test-swagger ``
* Add the following page: ``_index.md`` with the following content:

```markdown
---
title: API Reference
description: TEST
weight: 9
type: swagger
---

{{< swaggerui  src="test.json" >}}

```

* Add also a OPENAPI/SWAGGER file besides names test.json

Thanks for your help

chalin:
@theletterf can you take a look?

alexandre-touret:
Hi,
* I tested it on Edge & Firefox 
* I tested both configurations : an update and I reproduced it on a freshly cloned docsy example (BTW, I used to explain how to reproduce this issue)
* I just thought about my corporate computer embeds a local proxy/firewall ZScaler
* I can't access the unpkg URLs from the same browsers

theletterf:
[...]

PR #2030, opened 2024-06-03
Add note on SwaggerUI and networking

theletterf:
Following findings in:

- #2017


chalin (approved):
Thanks for the followup PR!

Issue #1537, opened 2023-06-05
Support `<link rel="canonical" ...>` in `<head>`


chalin:
- https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-link-method
- Possibly something like https://github.com/open-telemetry/opentelemetry.io/pull/1649

iArchitSharma:
Hey @chalin, I've raised a [PR](https://github.com/google/docsy/pull/2032) which add support for `<link rel="canonical" ...>` in `<head>` for all pages

PR #2032, opened 2024-06-03
Added Support for Canonical page link in <head>

### LINKED ISSUE #2010: [i18n] Canonical page link of a fallback page should refer to default language page

iArchitSharma:
This PR fixes #2010 #1537 

this code generates a canonical link tag for a webpage based on the value of the `canonical_url` parameter in the content file's front matter. If the parameter is not defined, it falls back to the permalink of the page, ensuring that search engines understand the preferred URL for the page.

### DIFF
--- a/layouts/partials/hooks/head-end.html
+++ b/layouts/partials/hooks/head-end.html
@@ -1,0 +1,26 @@
+{{ with .Params.canonical_url -}}
+
+<link rel="canonical" href="{{ . }}">
+
+{{- else -}}
+
+  {{ $canonicalURL := .Permalink -}}
+
+  {{ $defaultLang := "en" -}}
+  {{ if and (ne .Language.Lang $defaultLang) .File -}}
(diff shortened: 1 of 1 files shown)


fekete-robert:
Hi, thanks for the PR!
I think this functionality should go directly into layouts/partials/head.html, or into a separate partial that's included in head.html.
The head-end hook is not a good place for it, because that's a placeholder file and many projects override it with their own code.

chalin:
Hi @iArchitSharma - it looks like you copied the file that I put together for the OTel.io site. Is that the case? If so, then that file is a partial experimental design (I still have work to do on it), _and_ it isn't suitable, in it's current form, for inclusion in Docsy yet.

iArchitSharma:
Yes, @chalin, I copied this from the OTel.io site. Although when I tested it, it seemed to work as intended. What improvements are needed here? I would love to help

PR #2044, opened 2024-07-02
Persist sidebar scroll position

huanlin:
Firstly, thank you all for the hard work!
Secondly, I'd like say that I'm no expert in Hugo, Docsy, and JavaScript. Therefore, this is more of a proposition for fixing issue  #257. 

After [discussing with prdsoftware](https://github.com/google/docsy/issues/257#issuecomment-2181744401), I realized that the fix must also take into account the `sidebar_cache_limit` option in the Hugo config file. So I implemented the fix in `/layouts/partials/sidebar.html`, and I've tested it on [my website](https://huanlin.cc/blog/) with side bar cache both enabled and disabled.

If there is anything that I've missed. Kindly let me know how to improve. Thanks!


fekete-robert:
LGTM, thanks a lot!

fekete-robert (approved):

Issue #2045, opened 2024-07-02
Taxonomy supports LinkedTitle


TonyWu1995:

TonyWu1995:
Additional Content:
Docsy does not support LinkedTitle in Taxonomy, which can cause the Taxonomy to become messy when there is a lot of lengthy content configured.

PR #2046, opened 2024-07-02
Added taxonomies LinkedTitle

TonyWu1995:
issue  #2045

Before:

image

After:

image

image

### DIFF
--- a/layouts/partials/taxonomy_terms_cloud.html
+++ b/layouts/partials/taxonomy_terms_cloud.html
@@ -10,7 +10,7 @@
       {{ end -}}
       <ul class="taxonomy-terms">
         {{ range $taxonomy -}}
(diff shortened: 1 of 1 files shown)


fekete-robert (approved):
LGTM, thanks for the fix!

huanlin (commented):
Looks good. Thank you!

deining (approved):
LGTM

chalin (commented):
@fekete-robert @deining @TonyWu1995 - won't this break if there is no `.LinkTitle`? Have you tested this scenario?

TonyWu1995:
> @fekete-robert @deining @TonyWu1995 - won't this break if there is no `.LinkTitle`? Have you tested this scenario?

> The LinkTitle method returns the linkTitle field as defined in front matter, falling back to the value returned by the [Title](https://gohugo.io/methods/page/title/) method. --from: [LinkTitle](https://gohugo.io/methods/page/linktitle/)


If there is no `linktitle`, the tag will be like this:

image

because the frontmatter is defined like this:
```
---
title: Artificial Intelligence
--- 
```

PR #2087, opened 2024-09-22
remove rss_sections from docs

### LINKED ISSUE #2055: Drop `rss_sections` from the user guide
While is see `rss_sections` in the docs, I don't see it referenced in the layouts. Am I missing something here?

I've done a global search through GitHub and only found one project's layout that uses this parameter: https://github.com/apache/beam/blob/50a3403a4742e1c9e264f57f4411969daeff4642/website/www/site/layouts/index.feed.xml#L23

> image

LDReid:
- Fixes #2055


chalin (approved):
LGTM, thanks @LDReid ✨

PR #2091, opened 2024-10-08
Show only one year when copyright year range is a single year

### LINKED ISSUE #2047: Show only one year when copyright year range is a signle year
- Followup to #1821
- Don't show year range if `from_year` is the same as the current year; or, said an almost equivalent way, show single year when range is one year.

tcrouch:
Updates the copyright template so a single year or a range is shown as appropriate.

For example, given `from_year` and `to_year` are set to 2024.

Previous output:

> © 2024–2024

New output:

> © 2024

Closes #2047

### DIFF
--- a/theme/layouts/_partials/footer/copyright.html
+++ b/theme/layouts/_partials/footer/copyright.html
@@ -1,21 +1,22 @@
 {{ $page := .Page -}}
 {{ with .Site.Params.copyright -}}
-  {{ $fromYear := "" -}}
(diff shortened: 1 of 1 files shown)


chalin:
- Superseded by #2720, which lands this same-year collapse together with the `from_year`/`to_year` contract: docs, changelog entry, and red-proven contract tests. Thanks @tcrouch for seeding the fix!

PR #2097, opened 2024-10-15
[RTL] Drop unused BS asset and scripts to fetch it

chalin:


chalin:
I thought that I'd develop more over the `rtl` branch, but we seem to be pretty much good to go with the RTL support, so I'm merging into `main` now.

PR #2104, opened 2024-10-22
User guide: correct instructions on how to disable RSS feed

### LINKED ISSUE #2103: disableKinds should not be located under params in the hugo.yaml file
I tried to use disableKinds to disable the generation of the index.xml file. Following the documentation I added

```
params:
  disableKinds: [RSS]
```

to my hugo.yaml file. Unfortunately it did not have any effect, the file was still being generated. I then noticed the hugo.yaml file already contained a disableKinds entry for the taxonomies, not under params. Adding "disableKinds: [RSS]" in the same place disabled the generation of the index.xml file.

Please update the documentation to reflect this.

deining:
This PR resolves #2103.

### DIFF
--- a/userguide/content/en/docs/adding-content/content.md
+++ b/userguide/content/en/docs/adding-content/content.md
@@ -619,27 +619,23 @@ You can also use this directory for other files used by your project, including 
 
 ## RSS feeds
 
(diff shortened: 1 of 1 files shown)


chalin (approved):
Thank you for the update.

PR #2112, opened 2024-11-06
Update pkgs, with Hugo to 0.137.0

chalin:


chalin:
I was hoping that this might address the `markdown-link-check` failures under GH Ubuntu, but it doesn't. Will still merge, and address the link-checking issue separately.

PR #2117, opened 2024-11-06
[UG] Give guidance on what to do on "failed to load modules" error

### LINKED ISSUE #2116: [CI] Docsy 0.8.0 NPM-package use may fail
This was uncovered by @sftim through his work on

- https://github.com/kubernetes/website/issues/41171

To reproduce the problem, run the following:

```sh
hugo new site --format yaml myproject
cd myproject
echo 'theme: docsy' >> hugo.yaml
echo 'themesDir: node_modules' >> hugo.yaml
npm init -y
npm install --save-dev google/docsy#semver:0.8.0 autoprefixer postcss-cli
hugo # Ensure that it is v0.120.4
```

Hugo reports: "Error: failed to load modules: module "github.com/FortAwesome/Font-Awesome" not found in "/Users/chalin/tmp/myproject/node_modules/github.com/FortAwesome/Font-Awesome"; either add it as a Hugo Module or store it in "/Users/chalin/tmp/myproject/node_modules".: module does not exist"

The problem is that Docsy's `postinstall` NPM script isn't being run automatically, but it should.

chalin:
- Fixes #2116
- Gives guidance on what to do if Hugo reports a "failed to load modules" error when using Docsy via NPM
- Ran Prettifier over the updated files
- Makes setting of `themesDir` in the sample config the "happy path"
- **Preview**: https://deploy-preview-2117--docsydocs.netlify.app/docs/get-started/other-options/#option-3-docsy-as-an-npm-package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -155,10 +155,11 @@ For the full list of changes, see the [0.8.0] release notes.
 
 **Breaking changes**:
 
-- Docsy is packaged as a **single Hugo module** ([#1120]).
-  - For details, see [Use Docsy as a Hugo Module].
-- **Important**: non-Hugo-module projects should read the [Docsy NPM install
-  side-effect] note.
+- Docsy is packaged as a **single Hugo module** ([#1120]). For details, see [Use
+  Docsy as a Hugo Module].
+- **Important**: non-Hugo-module projects should be aware of the [Docsy
+  NPM install side-effect]. Also, for guidance on Hugo-reported "failed to load modules"
+  error, see [Docsy as an NPM package].
 - **Page feedback**, or [User feedback]:
   - In support of projects configuring analytics outside of Docsy, feedback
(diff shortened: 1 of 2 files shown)


LisaFC (approved):

sftim:
A related concern; the postinstall I think runs `cp -R node_modules/bootstrap/scss/vendor/* assets/_vendor/bootstrap/scss/` and this is making assumptions about both the current working directory and the name of the install target.

Seems like NPM might not guarantee those are always the right values to make that command work.

chalin:
For the `cp` command, we've switched to use of the NPM `cpy-cli` command. As for the working directory, I'll leave that as is for now. We can address any specific issues if/when they arise.

chalin:
@sftim - btw, unless you run `ci:prepare` you won't hit the copy commands. The `ci:prepare` is only for those maintaining and releasing Docsy.

PR #2140, opened 2024-12-08
[0.5.x] Create v0.5.1+2 "patch" without devDep

chalin:
When using `docsy` as an NPM module either directly from GitHub or via a path, devDeps get installed too, and this seems to cause problems for projects using Docsy in some containerized environments (like what we're trying to do in https://github.com/kubernetes/website/pull/48812). This patch effectively hides the devDep.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,9 @@
-tmp/
 node_modules/
 package-lock.json
 
 # Hugo
 .hugo_build.lock
-public/
+/public
 resources/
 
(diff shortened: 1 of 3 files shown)


chalin:
Note that this PR contains an intermediate commit that experiments with the NPM package name `@docsy/theme` for Docsy. I'm keeping the commit but have overridden the changes. We'll consider implementing this, or not, at `HEAD` if when the need arises.

PR #2149, opened 2024-12-22
i18n: Add feedback_* to i18n/ja.toml

rinsuki:
Context: while reading OpenTelemetry's document, I noticed that their feedback form's i18n is completely broken in Japanese:

image

(Screenshot of https://opentelemetry.io/ja/docs/ )

and after some research, I think it was from docsy, not opentelemetry.io's fault, then I just made this patch.

### DIFF
--- a/i18n/ja.toml
+++ b/i18n/ja.toml
@@ -77,3 +77,13 @@ other = "If you want to get more involved by contributing to {{ .Site.Title }}, 
 other = "You can find out how to contribute to {{ .Site.Title }} in our"
 [community_guideline]
 other = "Contribution Guidelines"
+
+# Feedback
+[feedback_title]
+other = "フィードバック"
+[feedback_question]
+other = "このページは役に立ちましたか?"
+[feedback_positive]
+other = "役に立った"
+[feedback_negative]
+other = "役に立たなかった"


chalin:
Thank you @rinsuki, I've asked https://github.com/orgs/open-telemetry/teams/docs-ja-maintainers to review and comment.

chalin (approved):
LGTM. Will give OTel `docs-ja-maintainers` some time to give feedback.

i18n/ja.toml
```diff
@@ -77,3 +77,13 @@ other = "If you want to get more involved by contributing to {{ .Site.Title }},
 other = "You can find out how to contribute to {{ .Site.Title }} in our"
```
deining, line 77: Just realized that there are some more untranslated lines/strings in this file:

https://github.com/google/docsy/blob/fa81a86be8bf130d94a43b8898c6716819fdae87/i18n/ja.toml#L63C1-L79C34

@rinsuki: are you able to translate these lines as well? Your help is appreciated!

rinsuki: technically I can directly translate word-by-word, but I think it wouldn't be a good translate (If I do that, japanese speakers (like me) feels it is translated from other language rather than written by native), but I don't have a idea to good paraphrase for its text right now.

by the way, `community_how_to` assume text ends with `community_guideline`:

https://github.com/google/docsy/blob/fa81a86be8bf130d94a43b8898c6716819fdae87/layouts/partials/community_links.html#L18-L21

but some languages wants to have additional text *after* link, it would be good to have something like `community_how_to_suffix`.

ymotongpoo:
@chalin Do you need Japanese speakers review? (it seems https://github.com/orgs/open-telemetry/teams/docs-ja-maintainers consists of non-Japanese speakers. Asking just in case.)

If yes, the content looks good to me. (btw, https://github.com/orgs/open-telemetry/teams/docs-ja-approvers is `docs-ja-maintainers` + me and @katzchang)

chalin:
> @chalin Do you need Japanese speakers review? (it seems https://github.com/orgs/open-telemetry/teams/docs-ja-maintainers consists of non-Japanese speakers. Asking just in case.)

Oh! Yes, that was my intention.
 
> If yes, the content looks good to me.

Thank you for validating.

> (btw, https://github.com/orgs/open-telemetry/teams/docs-ja-approvers is `docs-ja-maintainers` + me and @katzchang)

Got it, thanks.

i18n/ja.toml
chalin: Ok, thanks @rinsuki. I'll merge this as an increment, and we can address the other fields as necessary. Thanks for pointing that out @deining.

PR #2154, opened 2025-01-07
[CI] Upgrade NPM pkgs, Hugo to 0.140.2

chalin:
- Upgrades NPM packages, including Hugo to 0.140.2
- Adjusts `build` scripts so that we can pass in extra arguments like `npm run build -- --logLevel info`


chalin:
All checks are passing. Merging.

PR #2227, opened 2025-04-25
User guide: remove deprecated kind 'taxonomyterm' from disableKinds

deining:
This PR helps to avoid a deprecation warning about deprecated kind `taxonomyterm` in disableKinds.

**Reproduction:**:

Add this line to your `hugo.yaml`:

```
disableKinds = ["taxonomy", "taxonomyTerm"]
```

Now when previewing the site, a warning appears:

```
WARN  DEPRECATED: Kind "taxonomyterm" used in disableKinds is deprecated, use "taxonomy" instead.
```

This PR corrects the user guide so that this warning will **not** be printed out.


chalin (approved):
LGTM, thanks. I'll make a followup pass for copyedits.

PR #2234, opened 2025-05-06
Fix css variable font-weight-bold

jorado:
Use the css variable `font-weight-bold` instead of hardcoding the font weight to 700.

### DIFF
--- a/assets/scss/_taxonomy.scss
+++ b/assets/scss/_taxonomy.scss
@@ -24,7 +24,7 @@
     display: inline-block;
     width: 100%;
     font-size: 1rem;
-    font-weight: 700;
+    font-weight: var(--font-weight-bold);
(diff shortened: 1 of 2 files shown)


chalin (commented):
Thanks. Instead use `$font-weight-bold` (as defined in https://getbootstrap.com/docs/5.3/utilities/text/#sass-variables).

assets/scss/_taxonomy.scss
```diff
@@ -24,7 +24,7 @@
     display: inline-block;
     width: 100%;
     font-size: 1rem;
-    font-weight: 700;
+    font-weight: var(--font-weight-bold);
```
chalin, line 27:
```suggestion
    font-weight: $font-weight-bold;
```

chalin (approved):

PR #2237, opened 2025-05-09
Cleanup: drop unneeded template

deining:
With https://github.com/gohugoio/hugo/pull/12505 merged, we can drop this file.


chalin (approved):
Thanks, LGTM. (Hugo's template seems to generate the same GA code in the quick test that I did.)

PR #2238, opened 2025-05-09
Adapt to new template system in Hugo v0.146.0

deining:
This PR brings all adaptions to Hugo's new template system, as described in the [official docs](https://gohugo.io/templates/new-templatesystem-overview/).

Please note that with these changes, preview of user guide fails with hugo version 0.145.0 or lower.


chalin:
Thanks for exploring this. I've created an issue to track the required changes and hold any discussions about it:

- #2243

I'll dig in deeper as soon as I can, hopefully today.

chalin:
@deining - I think that instead I'm going to step through these changes incrementally and update the UG as I go along -- e.g., #2246. /cc @LisaFC @emckean

chalin:
Thanks for initiating this. Closing since the work is being covered via #2243.

I'm glad to have stepped through this piecemeal. Among other things, the taxonomy-terms layout filename swap might have been more difficult to pinpoint otherwise (#2257).

PR #2252, opened 2025-05-18
Update Swagger UI to 5.21.0 (latest)

Gr3q:
I've found that the examples don't work with the current version included in some cases, see https://github.com/opensubsonic/open-subsonic-api/pull/137. In that case updating to the latest swagger version seems to solve the problem.

### DIFF
--- a/layouts/swagger/baseof.html
+++ b/layouts/swagger/baseof.html
@@ -6,7 +6,7 @@
   <head>
     {{ partial "head.html" . }}
     <title>{{ if .IsHome }}{{ .Site.Title }}{{ else }}{{ with .Title }}{{ . }} | {{ end }}{{ .Site.Title }}{{ end }}</title>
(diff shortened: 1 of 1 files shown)


chalin:
@theletterf - WDYT?

theletterf:
@Gr3q  Thanks! An update was due anyway, though I'd like to see what OAS snippet wasn't rendering properly. Could you share the OAS and show a screenshot of the issue?

Gr3q:
I checked again, we were using docsy 0.5.1, changing to unpkg in [0.7.2](https://github.com/google/docsy/releases/tag/v0.7.2) fixed that already. This is how the examples looked before. This is the OAS file: https://opensubsonic.netlify.app/docs/openapi/openapi.json

Screenshot_2025-05-27_22-25-29
Screenshot_2025-05-27_22-25-16

Issue #2243, opened 2025-05-16
Adapt to new template system in Hugo v0.146.0


chalin:
For details, see:

- https://gohugo.io/templates/new-templatesystem-overview/
- https://discourse.gohugo.io/t/documentation-of-the-new-template-system-in-hugo-v0-146-0/54686

### Tasks

- [x] #2246
- [x] #2247
- [x] #2250
- [x] #2251
- [x] #2255
- [x] #2257
    #2260
- [ ] ...

### Related refactorings

- [x] #2256
  - Seems like I'll have to revert this in light of https://discourse.gohugo.io/t/can-partials-be-page-path-specific-too/54836
- [x] #2259
- [x] #2260
- [x] #2261
- [ ] Consider using `.Render` for the render-heading hook inclusion?
- [ ] ...

deining:
WIth the change to the new template system, this is not vlaid any more:

https://github.com/google/docsy/blob/8e5229130c4bda0a1238822cd0af0fd7c076a141/theme.toml#L9

We should also scan the user guide and the exampel site for any statements regarding the require minimum version of Hugo. I guess we have to raise it to v0.146.0 now.

Issue #1654, opened 2023-08-10
Namespace Docsy SCSS files


chalin:
Consider the following list of SCSS files from a Docsy-based projects:

```
_about.scss
_archive.scss
_buttons.scss
_cookie-banner.scss
_footer.scss
_hamburger.scss
_header.scss
_main.scss
_projects.scss
_reset.scss
_search.scss
_sidebar-left.scss
_sidebar-toc.scss
_styles_project.scss
_taxonomy.scss
_typography.scss
_variables_project.scss
```

Which of these override Docsy files? I can't tell without doing a directory diff. It shouldn't be that hard. Also, as projects and Docsy SCSS files are added and renamed, we currently run the risk of unintended shadowing/overrides.

We should namespace Docsy most SCSS files to avoid such problems. 

I see two options for namespacing:

- Put files in a directory, such as `docsy`
- Use a file prefix or suffix, such as `td`

PR #2256, opened 2025-05-22
Move new content-after-header hook into partials folder, add UG entry

chalin:
- Cleanup done as a part of #2243 
- Moves `layouts/td-content-after-header.html` into the `_partials` folder since the hook is actually a partial just like the other hooks.
- Developer note: why place the file under `_partials/td` and not `_partials/td/hooks`? Because everything under `td` is a hook atm. Note that use of `td` is in support of #1654
- Adds an entry to the UG for this new feature, marking it as experimental. /cc @LisaFC 
  **Preview**: https://deploy-preview-2256--docsydocs.netlify.app/docs/adding-content/lookandfeel/#before-page-content

### DIFF
--- a/userguide/content/en/docs/adding-content/lookandfeel.md
+++ b/userguide/content/en/docs/adding-content/lookandfeel.md
@@ -507,6 +507,16 @@ Both [head.html] and [scripts.html] are included from [baseof.html], Docsy's
 [scripts.html]:
   https://github.com/google/docsy/blob/main/layouts/_partials/head.html
 
+### Adding a banner before page content (EXPERIMENTAL) {#before-page-content}
+
+To have a banner, or other similar content, appear at the top of the pages in a
+section, add the content to the [td/content-after-header.html] partial. It will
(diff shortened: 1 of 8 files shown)


userguide/content/en/docs/adding-content/lookandfeel.md
```diff
@@ -507,6 +507,16 @@ Both [head.html] and [scripts.html] are included from [baseof.html], Docsy's
 [scripts.html]:
   https://github.com/google/docsy/blob/main/layouts/_partials/head.html
 
+### Adding a banner before page content (EXPERIMENTAL) {#before-page-content}
+
+To have a banner, or other similar content, appear at the top of the pages in a
+section, add the content to the [td/content-after-header.html] partial. It will
+appear inside the `div.td-content`, after `</header>`, just before `.Content` is
```
LisaFC, line 514: My only suggested change here is to take out one of the commas as it's a bit hard to read, so:

It will appear inside the `div.td-content` after `</header>`, just before `.Content` is
rendered.

PR #2297, opened 2025-07-21
Add dark theme for Swagger

### LINKED ISSUE #2293: Swagger UI shortcode doesn't render well under dark mode
### Environment

- [x] Version of Docsy you are using: v0.12
- [ ] How are you using Docsy? As a:
  - [x] Hugo module
  - [ ] NPM module
  - [ ] Git submodule
  - [ ] Other:
- [x] Version of Hugo you are using (output of `hugo version`): v0.147.8
- [ ] OS:
  - [x] Linux
  - [ ] macOS
  - [ ] Windows
  - [ ] Other:

### Problem

under dark mode, the swagger UI renders like the following:

Image

Image

theletterf:
This adds a dark theme SCSS file for Swagger UI, adapted from https://github.com/Amoenus/SwaggerDark

Demo:

Screenshot 2025-07-21 at 09 31 45

Fixes #2293

### DIFF
--- a/assets/scss/_swagger_dark.scss
+++ b/assets/scss/_swagger_dark.scss
@@ -1,0 +1,764 @@
+// Dark-mode overrides for Swagger UI, converted from https://github.com/Amoenus/SwaggerDark
+// These styles are only applied when the Docsy/Bootstrap data attribute indicates dark theme.
+
+[data-bs-theme="dark"] {
+   .swagger-ui, .swagger-ui section h3 { color: #b5bac9; }
+
+   .swagger-ui a { background-color: transparent; }
+
+   .swagger-ui mark {
+       background-color: #664b00;
+       color: #bfbfbf;
+   }
+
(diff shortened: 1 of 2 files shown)


xuchen:
thank you! this works for me perfectly!

theletterf:
@chalin @LisaFC PTAL

chalin (commented):
Thanks for this @theletterf! A few comments:

- Some styling / formatting seems broken in your sample, or is that normal? image
- We need a test of this either from the Docsy Example or in this repo. 🤔 
- I'd prefer a deeper integration with Bootstrap and it's CSS variables, but this is better than no dark mode support.

chalin:
@theletterf - a few more comments:

- Please address the build failures (maybe by running `fix:format`).
- Might there be a way that we can use the style(s) from https://github.com/Amoenus/SwaggerDark as is, and provide a delta over that. It might be easier to keep up with fixes to SwaggerDark.

PR #2303, opened 2025-08-02
Always show lang menu in navbar

### LINKED ISSUE #2035: Language dropdown hidden for narrow displays
Originally reported by @svrnm via https://github.com/open-telemetry/opentelemetry.io/issues/4597:

> When I go to opentelemetry.io with a mobile browser the language drop down is not visible
>
> image

### LINKED ISSUE #2132: [UX] Leave language dropdown menu in the top nav, use icon for title
Both the language menu and the search box in the top nav move to the side nav when the display gets narrow. This has been a source of confusion for many users and site designers. I'd like to suggest the following (IMHO) improved design for the language menu

- **Leave the language menu in the top nav** for all display widths
- Consider the option of having a shorter menu title, such as a [globe](https://fontawesome.com/icons/globe): 

---

Other thoughts about the menu title:

- By default, we'd show only the icon, which could be set via CSS, via `:after` for easier customization. And the current language could be hidden by default (again via CSS).
- An alternative icon could be the [language](https://fontawesome.com/icons/language?f=classic&s=solid) icon, but visually it is difficult to make out, even if set to a `larger` font size: 

@svrnm @theletterf, this is proposed as a solution to:

- https://github.com/open-telemetry/opentelemetry.io/issues/4597

chalin:
- Fixes #2035: the language menu will always be shown in the navbar for multilingual sites.
- Fixes #2132: added the globe icon as suggested.
  - On mobile, uses language code as menu title
- Adds a placeholder language so that multilingual features can be demonstrated. Ensure that those pages are marked as noindex.
- Fixes placement issues with light-dark dropdown
- Removes overflow-hidden for the navbar on mobile. That'll not look super pretty, but it'll make obvious a feature that most users don't know exist: that the navbar can scroll! Feedback is welcome. 

**Preview**: TBC

### Screenshots

TBC

/cc @vitorvasc @theletterf @svrnm @tiffany76 - PTAL

### DIFF
--- a/assets/scss/_nav.scss
+++ b/assets/scss/_nav.scss
@@ -46,7 +46,7 @@
   }
 
   .navbar-nav {
-    padding-top: $spacer * 0.5;
+    // padding-top: $spacer * 0.5;
     white-space: nowrap;
   }
 
@@ -69,10 +69,6 @@
     .nav-item {
       padding-inline-end: $spacer * 0.5;
     }
-
-    .navbar-nav {
-      padding-top: 0 !important;
-    }
   }
 
   @include media-breakpoint-down(lg) {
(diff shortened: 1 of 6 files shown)


LisaFC:
Oh that's excellent. I think the scrolly navbar on mobile works well.

LisaFC (approved):

PR #2339, opened 2025-10-16
Fix left-padding of language menu in a foldable sidebar

### LINKED ISSUE #2001: Languages menu in side nav design issues: mispositioned and more
Consider the image below. Here are the issues:

- [x] The menu appears far to the right, rather than below the menu title. **Fixed** a while a go this seems.
- [x] The language dropdown needs left margin or padding
  - [x] Fix for regular sidebar - #2183
  - [ ] Fix for foldable sidebar - ...
- [x] Text color:
  - [x] Menu title text color seems off - #2183
  - [x] Menu entry text doesn't appear to be of the right color - #2183
- [x] #2035

---

> Screen Shot 2024-05-16 at 03 59 53

---

Related:

- #2131

chalin:
- Fixes #2001 for foldable sidebars
- Followup to 2183
- The problem was that `.dropdown` was applied to two elements. We need it on only one.
- Ran Prettier over changed style file
- Enables the foldable config parameter for the UG sidenav so that we can see this fix.
- **Preview**: https://deploy-preview-2339--docsydocs.netlify.app/docs/ - you'll need a tablet-sized viewport to see the language menu in the sidenav

### Screenshot

> image

### DIFF
--- a/layouts/_partials/sidebar-tree.html
+++ b/layouts/_partials/sidebar-tree.html
@@ -42,7 +42,7 @@
       {{- if .Site.Params.ui.sidebar_root_enabled }} data-sidebar-root="{{ $sidebarRootID }}"{{ end -}}
     >
     {{ if and .Site.Params.ui.sidebar_lang_menu (gt (len .Site.Home.Translations) 0) -}}
-    <div class="td-sidebar-nav__section nav-item dropdown d-block d-lg-none">
+    <div class="td-sidebar-nav__section nav-item d-block d-lg-none">
       {{ partial "navbar-lang-selector.html" . }}
     </div>
     {{ end -}}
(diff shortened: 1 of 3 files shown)


layouts/_partials/sidebar-tree.html
```diff
@@ -42,7 +42,7 @@
       {{- if .Site.Params.ui.sidebar_root_enabled }} data-sidebar-root="{{ $sidebarRootID }}"{{ end -}}
     >
     {{ if and .Site.Params.ui.sidebar_lang_menu (gt (len .Site.Home.Translations) 0) -}}
-    <div class="td-sidebar-nav__section nav-item dropdown d-block d-lg-none">
+    <div class="td-sidebar-nav__section nav-item d-block d-lg-none">
       {{ partial "navbar-lang-selector.html" . }}
```
chalin: The `.dropdown` style is applied to a div inside "navbar-lang-selector.html": https://github.com/google/docsy/blob/1b961c9b3a1999b8a294b9d0d20f13ffafb3bb4e/layouts/_partials/navbar-lang-selector.html#L3

So we don't need it on this element too.

Issue #2345, opened 2025-10-21
With Hugo v0.152.0, getting error: invalid module config for "github.com/google/docsy": "../../node_modules/bootstrap"


tekezo:
### Environment

- [x] Version of Docsy you are using: v0.12.0
- [x] How are you using Docsy? As a:
  - [x] Hugo module
- [x] Version of Hugo you are using (output of `hugo version`): hugo v0.152.0+extended+withdeploy darwin/arm64 BuildDate=2025-10-21T16:29:49Z VendorInfo=brew
- [x] OS:
  - [x] macOS

### Problem

With Hugo v0.152.0, I'm getting the following error at running "hugo server".

```
Error: command error: failed to load modules: invalid module config for "github.com/google/docsy": "../../node_modules/bootstrap": mount source must be a local path for modules/themes
```

It seems this is happening because that behavior was disallowed by this PR in Hugo.

- https://github.com/gohugoio/hugo/pull/14071
- https://github.com/gohugoio/hugo/issues/14069

On my end, it started working after I removed the configuration for `Docsy as an NPM package` from hugo.yaml.
https://github.com/google/docsy/blob/main/hugo.yaml#L39-L44

fabiendelpierre:
This is happening on Linux as well FWIW (predictably). Working around it for now by capping the version of Hugo to v0.151.2 in my builds.

tekezo:
In Hugo v0.152.0 the minifier was also updated, and with tdewolff/minify v2.24.4, Popper's positioning code breaks. As a result, production builds with Hugo v0.152.0 cause popovers to render in the wrong place. (I haven't investigated this deeply)
So, I've also reverted to Hugo v0.151.2 for now.

marshalc:
For the purposes of linking up, my posting about this build issue on the Hugo forum led to this Issue being logged there: https://github.com/gohugoio/hugo/issues/14083

jmooring:
@tekezo ~~What does "Popper's positioning code breaks" mean?~~

See:
- <https://github.com/gohugoio/hugo/issues/14086>.
- <https://github.com/tdewolff/minify/issues/863>

tekezo:
@jmooring  Thank you. The Popper issue I encountered is exactly the same as the one you reported. The popover always appears at the origin of the browser window.

jmooring:
Both of the problems described above have been addressed with [v0.152.2](https://github.com/gohugoio/hugo/releases/tag/v0.152.2).

[...]

PR #2347, opened 2025-10-24
Fix paths for Bootstrap and FontAwesome sources

hwidjaya-telus:
To conform with the latest Hugo changes in 0.152.2 in module mount points.

---

For context, see:

- #2345


hwidjaya-telus:
This is a very simple change to fix the issue with relative paths in module's hugo.yaml pointing to non portable locations for dependent packages or hugo modules.

chalin:
Hi @hwidjaya-telus - please provide some context. What is your setup. What do you expect, and what are you getting instead?

@deining - are you confirming that you've face this problem too?

deining:
> @deining - are you confirming that you've face this problem too?

I don't consider this a problem, this is an improvement for me.
I think this us a useful change, I would like to see that merged.

chalin:
> I don't consider this a problem, this is an improvement for me.
> I think this us a useful change, I would like to see that merged.

Thanks @deining but if it breaks the builds for others, that isn't acceptable.

@hwidjaya-telus - If my understanding is correct, according to https://github.com/gohugoio/hugo/issues/14089, those relative paths are not a problem. I'm closing this for now. If there is still a problem, then consider opening an issue that describes the nature of the problem and your setup.

chalin:
This also seems to confirm that the change in this PR isn't required:

- #2345

PR #2389, opened 2025-11-19
chore: part 2 of reverting ScrollSpy safeId PR

chalin:
- Reverts the prefixing of changelog version-number headings do via #2370
- Followup to #2383


chalin:
Closing. After some thought: let's converge on using the git tag name, which starts with a v.

Issue #2390, opened 2025-11-19
Add support for Hugo (GFM, Obsidian) Markdown alert syntax


chalin:
References:

- https://gohugo.io/render-hooks/blockquotes/#alerts
- https://docs.github.com/en/contributing/style-guide-and-content-model/style-guide#alerts
- https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/

Consider driving them from the same map as BS uses to generate its alerts. This way, if a project extends that map, we will take it into account.

chalin:
Done! ✨

PR #2462, opened 2026-01-18
Add alert label translations for zh-cn and zh-tw

chalin:
- Contributes to #2390

### DIFF
--- a/i18n/zh-cn.yaml
+++ b/i18n/zh-cn.yaml
@@ -1,7 +1,7 @@
 # Alert labels
 caution: 注意
 important: 重要
-note: 备注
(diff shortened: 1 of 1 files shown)


chalin:
@reyang @windsonsea @my-git9 - might one of you be able to review this PR?

i18n/zh-cn.yaml
```diff
@@ -1,3 +1,10 @@
+# Alert labels
+caution: 注意
+important: 重要
+note: 备注
```
my-git9:
```suggestion
note: 说明
```
It's better to keep it consistent with zh-tw.

PR #2464, opened 2026-01-18
Add alert labels and all-rights-reserved in Japanese

chalin:
- Contributes to #2390

### DIFF
--- a/i18n/ja.yaml
+++ b/i18n/ja.yaml
@@ -15,7 +15,7 @@ ui_search: サイトを検索...
 ui_in: in
 
 # Footer text
-footer_all_rights_reserved: 著作権所有
+footer_all_rights_reserved: All Rights Reserved
 
 footer_privacy_policy: プライバシーポリシー
 


chalin:
@Msksgm @kohbis @ymotongpoo - might you be available to review this PR?

i18n/ja.yaml
```diff
@@ -8,7 +15,7 @@ ui_search: サイトを検索...
 ui_in: in
 
 # Footer text
-footer_all_rights_reserved: All Rights Reserved
+footer_all_rights_reserved: 著作権所有
```
ymotongpoo:
```suggestion
footer_all_rights_reserved: All Rights Reserved
```

Even in Japanese web sites, historically the English notion has been used as-is, so keeping English version is better than self translated Japanese one (Interestingly we don't have fixed translation for the phrase.)

ymotongpoo (changes requested):
@chalin I made a small request, but everything looks good otherwise.

chalin:
Thank you!

Issue #2420, opened 2025-12-04
GetRemote: failed to resolve media file `woff`, `woff2` when upgrading to Docsy v0.13 → v0.12.


jrjohnson91:
### Environment

- [x] Version of Docsy you are using: `v0.13`
- [x] How are you using Docsy? As a:
  - [x] Hugo module
  - [ ] NPM module
  - [ ] Git submodule
  - [ ] Other:
- [x] Version of Hugo you are using (output of `hugo version`): `hugo v0.152.2-6abdacad3f3fe944ea42177844469139e81feda6+extended linux/amd64 BuildDate=2025-10-24T15:31:49Z VendorInfo=gohugoio`
- [x] OS:
  - [x] Linux
  - [ ] macOS
  - [ ] Windows
  - [ ] Other:

### Problem:

I am receiving the error `GetRemote: failed to resolve media type for remote resoure `when trying to get `“https://unpkg.com/katex@latest/dist/fonts/KaTeX_***.woff”` files after upgrading to Docsy `v0.13` from `v0.12`.

### Steps:

1. Create a container with the following packages:

    ```sh
    sh-5.2# node --version
    v24.6.0
    sh-5.2# hugo version
    hugo v0.152.2-6abdacad3f3fe944ea42177844469139e81feda6+extended linux/amd64 BuildDate=2025-10-24T15:31:49Z VendorInfo=gohugoio
    sh-5.2# go version
    go version go1.25.3 (Red Hat 1.25.3-1.el10_1) linux/amd64
    ```
2. Copy `src` files
   

    ```toml
    # hugo.toml` 
    baseURL = 'https://compositepy.pages.boeing.com/site-hugo'
    languageCode = 'en-us'
    title = 'CompositePy Docs'
    
    enableRobotsTXT = true
    
    # Will give values to .Lastmod etc.
    enableGitInfo = true
    
    # Language settings
    contentDir = "content/en"
    defaultContentLanguage = "en"
    defaultContentLanguageInSubdir = false
    # Useful when translating.
    enableMissingTranslationPlaceholders = true
    
    [markup]
      [markup.highlight]
        style = "monokailight"
        noClasses = false
      [markup.goldmark]
        [markup.goldmark.renderer]
          unsafe = true
        [markup.goldmark.extensions] # https://www.docsy.dev/docs/adding-content/diagrams-and-formulae/
          [markup.goldmark.extensions.passthrough]
            enable = true
            [markup.goldmark.extensions.passthrough.delimiters]
[...]

PR #2465, opened 2026-01-19
KaTeX use: move media type declaration to theme's config file

deining:
- This PR brings a better user experience when using formulae with KaTeX, see #2420
- Motivation for this PR comes from this https://github.com/google/docsy/issues/2420#issuecomment-3757893854

**Preview**: https://deploy-preview-2465--docsydocs.netlify.app/docs/content/diagrams-and-formulae/

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -17,6 +17,14 @@ outputFormats:
     path: _print
     permalinkable: false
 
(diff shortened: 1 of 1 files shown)


chalin (approved):
👍🏻

PR #2470, opened 2026-01-21
bugfix: nested sidebar_root_for

### LINKED ISSUE #2475: Problem with nested `sidebar_root_for` directives
_Originally posted @reidmv in https://github.com/google/docsy/issues/2470#issuecomment-3795098237_

> when attempting to use nested `sidebar_root_for` directives, the sidebar root displayed would be inconsistent for nested "self" directives (the self page would have the correct sidebar, but not its children), and ineffective for nested "children" directives.
            
---

### From the repro PR

**Preview**: https://deploy-preview-2476--docsydocs.netlify.app/docs/get-started/docsy-as-module/installation-prerequisites/

| Expected sidebar | Actual sidebar |
|--------|--------|
| image | image | 

### Tasks

- [x] #2476 
- [x] #2470

reidmv:
Previously, when attempting to use nested `sidebar_root_for` directives, the sidebar root displayed would be inconsistent for nested "self" directives (the self page would have the correct sidebar, but not its children), and ineffective for nested "children" directives.

Investigation pointed to searching the ancestors collection in reverse order.

Hugo documentation for the Ancestors method states:

> Hugo orders the ancestors from closest to furthest.

See https://gohugo.io/methods/page/ancestors/.

The list returned by the Ancestors method is already in "reverse" order, with the deepest (closest) ancestor first. Calling Reverse on it gives a list with the root first, meaning that we will find the highest-level `sidebar_root_for` page, rather than the lowest level.

This commit removes the Reverse directive. `sidebar_root_for` directives can now be nested correctly.

---

- Fixes #2475
- **Preview**: https://deploy-preview-2470--docsydocs.netlify.app/docs/get-started/docsy-as-module/installation-prerequisites/

### DIFF
--- a/docsy.dev/content/en/docs/get-started/docsy-as-module/_index.md
+++ b/docsy.dev/content/en/docs/get-started/docsy-as-module/_index.md
@@ -5,6 +5,7 @@ weight: 1
 date: 2021-12-08T10:33:16+01:00
 description: >
   Learn how to get started with Docsy by using the theme as a Hugo Module.
+sidebar_root_for: children
 ---
 
 [Hugo modules](https://gohugo.io/hugo-modules/) are the simplest and latest way to use Hugo themes like Docsy when building a website. Hugo uses the modules mechanism to pull in the theme files from the main Docsy repo at your chosen revision, and it’s easy to keep the theme up to date in your site. Our example site uses Docsy as a Hugo module.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "docsy",
-  "version": "0.14.0-dev+59-g7bcae61",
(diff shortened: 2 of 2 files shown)


chalin:
Hi @reidmv. Thanks for reporting the issue, the analysis, and proposed fix. Well done 🙌🏻 

> when attempting to use nested `sidebar_root_for` directives, the sidebar root displayed would be inconsistent for nested "self" directives (the self page would have the correct sidebar, but not its children), and ineffective for nested "children" directives.

Let me get a repro first. Let's track this via:

- #2475

chalin:
I was sure I had tested nested scenarios, but maybe not 🤷🏼‍♂️. Thanks again @reidmv!

PR #2484, opened 2026-01-28
Simplify i18n files

### LINKED ISSUE #2401: Simplify i18n files since none of the entries can be pluralized
Entries all use the `other` form, but none can be pluralized, so they can all be of the form: ID = "value".

tcrouch:
Simplifies i18n files since none of the entries can be pluralised.

Closes #2401


chalin:
Hi. Actually, #2401 is done. The case you address is valid.

PR #2516, opened 2026-02-05
Fix links in repository CHANGELOG.md & CONTRIBUTING.md

rammba:
Hello @chalin, I've noticed that site links from repository markdown files are currently leading to the non-existing pages. This was introduced in #2446.

image


chalin:
@rammba - Actually, come to think of it, you're right: these shouldn't be 404s. I just wasn't sure which target to use. Can you update the links to use this host: https://main--docsydocs.netlify.app/

If not, that's ok, I'll update it later.

chalin:
Thanks. You'll need to run `npm run test` and commit the resulting changes.

PR #2553, opened 2026-02-13
layouts: use 'note' role instead of 'alert'

orangecms:
The 'alert' role is meant for feedback deserving immediate attention, such as an incoming message.
The 'note' role, on the other hand, is meant for content that has no suitable semantic element otherwise and is supposed to hint the user on ancillary information.

See:
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/note_role

same as https://github.com/kubernetes/website/pull/54363

### DIFF
--- a/layouts/_markup/render-blockquote-alert.html
+++ b/layouts/_markup/render-blockquote-alert.html
@@ -19,7 +19,7 @@
   {{ $classes = printf "%s %s" $classes . -}}
 {{ end -}}
 
-<div class="{{ $classes }}" role="alert"
+<div class="{{ $classes }}" role="note"
   {{- range $key, $value := .Attributes -}}
     {{ if eq $key "class" }}{{ continue -}}{{ end -}}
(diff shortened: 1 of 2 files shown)


chalin:
> The 'alert' role is meant for feedback deserving immediate attention, such as an incoming message. The 'note' role, on the other hand, is meant for content that has no suitable semantic element otherwise and is supposed to hint the user on ancillary information.
> 
> See:
> 
> * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role
> * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/note_role
> 
> same as [kubernetes/website#54363](https://github.com/kubernetes/website/pull/54363)

Hi @orangecms 👋🏻 Thanks for point this out! Let me create an issue to track this. I feel that we need a more nuanced solution that what is proposed here.

PR #2566, opened 2026-02-17
Doc-rooted mounts and more

chalin:
- Followup to #2565. Drops our disableKinds:home experiment, because it made `.Site.Home` map to section landing pages (e.g. to the about page when view it).
- Using a proper solution, which is to exclude the homepages (en and fr) via a mount files config
- ...


chalin:
Superseded

Issue #2504, opened 2026-02-01
Reintroduce support for doc-rooted sites


chalin:
I've suspected that docsy-only support wasn't working, but I've run out of time to retest the feature. Well, @adlerzei has been confirming that it's broken (thanks!). This is a tracker/meta issue for reintroducing support for it.

### Tasks

- [x] #2563
- [x] #2586
- [x] #2587

## Docs

For example see: https://doc-rooted--docsydocs.netlify.app/content/adding-content/#doc-rooted-sites

---

Related:

- [x] #2492 - #2496
- [x] #2499 
- [x] Copied from #2333, itself copied from #2328:
  - References for `toc_root` and docsy-only features:
    - #368 and https://mostlydocs.netlify.app
    - #173
    - Resources:
      - Repo: https://github.com/LisaFC/justdocs
      - Preview: https://docsyjustdocs.netlify.com/

PR #2579, opened 2026-02-18
chore(UG): use canonical URL to contributing page and simplify aliases

chalin:
- Cleanup in support of #2504

### DIFF
--- a/docsy.dev/config/_default/params.yaml
+++ b/docsy.dev/config/_default/params.yaml
@@ -6,8 +6,8 @@
 # version:
 #   status: archived # Then fetch include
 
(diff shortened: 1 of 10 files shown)


docsy.dev/config/_default/params.yaml
```diff
@@ -6,8 +6,8 @@
 # version:
 #   status: archived # Then fetch include
 
-version: &docsyVersion 0.14.3-dev
-tdBuildId: 017-over-main-b4d03ec0
+version: &docsyVersion 0.14.4-dev
```
chalin: Version bump is only because our version-bump script wasn't working, so fixing it now.

PR #2591, opened 2026-03-11
i18n, de: add missing translation for alert labels

deining:
This PR adds missing German translation for alert labels to `de.yaml`.


chalin (approved):
Thanks

PR #2635, opened 2026-05-16
Update start-from-scratch.md

dixonge:
In the "TL;DR" section the line

`hugo mod get github.com/google/docsy@vv0.15.0`

failed. There was an extra v. Removing it succeeds.

PR #2650, opened 2026-06-12
fix double v on version documentation

Marzal:
- Supersedes https://github.com/google/docsy/pull/2635
- Right now docs shows: github.com/google/docsy@**vv**0.15.0.
- So drooping de `v` on `v{{% param "version" %}}` should fix the issue.


chalin (approved):
Thanks for catching the rest 🙌🏻

Issue #2357, opened 2025-11-11
Improve favicon support, in particular for light/dark mode


chalin:
Possible improvements:

- [x] Support favicons at the site root
  - [x] For docsy.dev - #2653
  - [x] For all sites, via the theme's default favicons partial - #2654
- [x] Also support icon files being at the root
  - [x] For docsy.dev - #2653
  - [x] For all sites, via the theme's default favicons partial - #2654
- [x] Use `with` to avoid generating links to files that DNE - #2654
- [x] ~Report a (suppressible) warning if no favicon is found~ Favicons are not mandatory.
- [x] Support light/dark mode ico files maybe via `media` annotations (for ico file pairs with a `-dark` file variant)?
  - [x] For docsy.dev - #2653 - adds light/dark support for docsy.dev via an SVG favicon with a `prefers-color-scheme` rule
- [x] #2656

PR #2652, opened 2026-06-13
Stop shipping default favicons; generate from a single source image

### LINKED ISSUE #2595: Theme should not have default favicons
The Docsy theme should not publish Docsy's icons as defaults. This makes it messy to override icons, since _all_ of the ~25 Docsy-default icon files must be overridden. Most projects only generate a subset of these files, which means that for odd sizes or formats, browsers and other environments will be picking up the Docsy icons instead.

Related:

- #2357

chalin:
- Fixes #2595
- Contributes to #2357

The theme no longer ships its own favicon files. Instead, Docsy generates the favicons from a **single site-provided source image** and publishes them at the site **root**, where browsers and other clients probe for them by  analytics across Docsy sites show the bulk of favicon requests (and 404s) are root-path requests (`/favicon.ico`, `/apple-touch-icon.png`, `/favicon-16x16.png`, `/favicon-32x32.png`) that a `/favicons/`-based layout never satisfies.convention 

### What changed

- Removes the 18 bundled favicon files from `theme/static/favicons`, so the theme no longer imposes its own branding on downstream sites.
- Rewrites `_partials/favicons.html` to:
  - Generate PNG favicons (and a 180180 `apple-touch-icon`) from one source image via `Resize ... CatmullRom`.
  - **Publish them at the site root** with conventional names (`/favicon.ico`, `/favicon-16x16., `/apple-touch-icon.png`) via `resources.Copy`, so blind root probes resolve.png`, 
  - Publish `/apple-touch-icon-precomposed.png` by default (legacy iOS blind-probes that name; analytics show it as a top uncovered 404); sites can opt out via `params.td.favicons.no_apple_precomposed`.
  - Publish optional `favicon.svg` / `favicon.ico` variants as-is.
  - Warn once (suppressible via `ignoreLogs: ['docsy-no-favicons']`) when a site provides no favicons.

### Configuration

Sites supply a square source image (default `favicon.png`) under `assets/`, at the root or in `favicons/`. Optional overrides under `params.td.favicons` (defaults shown):

```yaml
params:
  td:
    favicons:
    path: favicon.* # base path of your icon files; the extension is ignored
    sizes: [16, 32, 48] # PNG sizes generated from the source image
    no_apple_precomposed: false # set true to skip apple-touch-icon-precomposed.png
```

### Breaking change

Favicon files now live under **`assets/`** rather than `static/`, because Hugo's resource pipeline (`resources.Get`) only sees `assets/`. Sites that previously dropped favicons in `static/favicons/` will need to move them and supply a square source image. This is noted for the 0.16.0 release notes.


chalin:
Superseded by #2654, which takes a simpler smart pass-through approach (discover and link conventionally named `static/` favicons) instead of in-Hugo generation. Closing in favor of that PR.

Issue #2683, opened 2026-07-19
Publish theme to the npm registry as @docsy/theme


chalin:
Publish Docsy to the npm registry as [`@docsy/theme`](https://www.npmjs.com/package/@docsy/theme), so that sites can install the theme (and its bundled npm dependencies) directly from the registry rather than from GitHub.

Planned scope, roughly in PR order:

- [x] Make `theme/` publishable: package metadata, `files` allowlist, bundled `gen-favicons` helper (#2684)
- [x] Docs: registry install as the lead install option (#2688)
- [x] Publish `@docsy/theme` (RC rehearsal done — `0.16.0-rc.1` under dist-tag `next`; stable publish at 0.16.0 tag time)
- [x] Announce in the 0.16.0 release post (#2690)

chalin:
Done: [`@docsy/theme@0.16.0`](https://www.npmjs.com/package/@docsy/theme) is published from the `v0.16.0` tag, with dist-tags `latest` = `next` = `0.16.0`, and the [release post](https://www.docsy.dev/blog/2026/0.16.0/) announcing the registry install is live.

PR #2684, opened 2026-07-19
Move gen-favicons into theme/, make it publishable as @docsy/theme

chalin:
- Contributes to #2683.
- **Preview(s)**: [Generate favicons](https://deploy-preview-2684--docsydocs.netlify.app/docs/content/iconsimages/#generate-favicons)
- Takes two steps toward publishing the theme to the npm registry; no publish happens in this PR (that's a tag-time step).
- **Moves `gen-favicons` into `theme/scripts/`** so the helper ships with the theme subtree. Bonus fix: the helper was unavailable to Hugo-module installs (the module root is `theme/`), and the docs pointed at the wrong module path — docs now use `DOCSY_THEME_DIR` with per-install-type definitions.
- **Makes `theme/` publishable as `@docsy/theme`**: package metadata, `files` allowlist (the git channel's file set minus npm-only exclusions), `bin` for `gen-favicons`, and a lean npm-facing `README.md`. The tarball `LICENSE` is materialized from the root LICENSE at pack time (`prepack`/`postpack`); the root tarball excludes any stranded copy, so both tarballs are disk-state-independent.
- **Guards the contracts in tests**: tarball contract test parameterized over both packages; manifest versions kept in sync by `set-package-version`; `test:tooling` lists test files explicitly so a missing file is a hard error.
- **Covers the consumer path** (`test:smoke`): packs `theme/`, installs the tarball into a scratch site outside the repo, builds, and runs the bin via `npx --no-install`; a registry-install variant stays skipped until a stable version is live under `latest`.
- **Verification**: `test:tooling` 53 tests, 52 pass, 1 pre-existing skip; `npm run check` green; `test:fixture-site` 37/37.


chalin:
Post-review refinement in 0cbec59c: the committed `theme/LICENSE` (and its byte-equality guard) is replaced by pack-time materialization — `prepack` copies the root LICENSE in, `postpack` removes it. Analysis showed the copy serves only the registry tarball: git-channel installs always carry the root LICENSE (npm's always-include list), and Hugo modules get it via the Go toolchain's nested-module injection. The pack contract's `required` LICENSE entry now doubles as the materialization guard, so drift is impossible by construction rather than merely detected (prior art: Angular/MUI/Nx). Tooling 51/52 green; the tarball consumer smoke test exercises the new lifecycle end to end.

Issue #2789, opened 2026-09-02
Add a JS plugin loading mechanism


chalin:
Tracking issue for a JS plugin mechanism (target: 0.18): projects add or replace page scripts through `assets/js/plugins/` plus a `params.docsy.plugins` entry, with no layout overrides.

- [x] Plugin substrate: scripts.html decomposition + plugin loop (#2788)
- [x] Convert built-in integrations onto the loop (existing config keeps working) (#2791)
- [x] Document the plugin mechanism (config, asset convention, companions, exact field types) (#2791)
- [ ] Opt-in interactive components built on the mechanism
- [ ] Decide head-side vs body-end emission for plugin stylesheets (body-end for now)
- [ ] 0.18 upgrade note: previously ignored override paths go live (`_partials/algolia/head.html`, and the new `_partials/scripts/*.html` names become overridable), and the internal named templates `algolia/head` and `algolia/scripts` no longer exist (replaced by the partials); sites calling or shadowing these will see the change

PR #2798, opened 2026-09-07
Plugins: gate tab persistence on the tabpane shortcode

chalin:
- Contributes to #2789
- **Scope**:
  - `tabpane-persist` ships only on pages that render the `tabpane` shortcode, from a `.HasShortcode` check in the plugin's shim; loop, schema, and registry untouched
  - fixture nets: where hook flags, shortcode flags, and shortcode names land per reuse path; the gate across direct, `.RenderShortcodes`, and `.Content` tabpanes
  - docs on the #2797 contract: plugins guide (shortcode predicate in § Adjust a plugin per page; § Page flags in included content restated), design § Gating decisions, quality nets, changelog, 0.18 post upgrade item
- **Out of scope**: section print (gated scripts never reach it; the print flag-merge partial is parked); the shim's name and contract wording (queued separately)
- **Why the shortcode name, not a shortcode-set flag**: a Store flag stays on the page whose file sets it, so a tabpane included through `.RenderShortcodes` never flags the includer (why 0.18-dev shipped it ungated); Hugo carries included pages' shortcode names to the includer since 0.123, pinned by a fixture test, so the documented include idiom gets the script with no site cooperation
- **Behavior change**: tab-less pages drop the script (scripts goldens: home page only); tabpanes reached through `.Content` or a printed section render without persistence

### DIFF
--- a/docsy.dev/content/en/blog/2026/0.18.0.md
+++ b/docsy.dev/content/en/blog/2026/0.18.0.md
@@ -100,6 +100,11 @@ three features onto it ([#2789][]).
   [`hooks/head-end.html`][head-end] partial; for the affected authoring paths
   and the one-line partial, see [When a MarkMap doesn't
   render][markmap-not-rendering].
+- **If tabpanes reach a page through `.Content` or a printed section**: tab
+  persistence now loads only on pages that render a `tabpane` shortcode,
+  includes through `.RenderShortcodes` included; other reuse paths render the
+  tabs without persistence. Include through `.RenderShortcodes`
+  ([why][ug-flags]).
 - **If you override a moved file**, [review and move your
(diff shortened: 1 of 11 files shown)


chalin:
Closing without merge: tab persistence stays ungated in 0.18, as in 0.17. The gate this draft added would trade a silent failure class (tabs without persistence under `{{< include >}}`, layout-level `.RenderShortcodes`, `.Content` includers, and section print) for a cached ~1 KB, which the 0.18 plugin design declines: a plugin gates itself only where the gate cannot miss, and a doubtful gate is ungated. The included-content facts this branch established are already stated in the plugins guide § Page flags in included content; its fixture test may return as a quality net separately. Tracked with the 0.18 plugin work on #2789.

PR #2806, opened 2026-09-17
Add Linux Foundation technical charter

krook:
* Adds the PDF
* Links from the README


krook (commented):
Thanks for linting @chalin!

